PR #8, opened 2017-01-05
update version and release notes for v00-10

gaede:


petricm:
I just bring it up again for consideration, once we discussed that maybe we should version lcgeo with the same number as DD4hep, so that we know easier what is matching. This would mean bumping lcgeo to 00-20, bu we never concluded on this.

gaede:
At some point (soon) DD4hep will be stable and there will be less frequent releases, whereas for lcgeo we will most likely have many releases with three concepts and calice using lcgeo. This will make it difficult to keep the version numbers in synch.

PR #19, opened 2017-01-27
DDSim: Change default tracker action to Geant4TrackerWeightedAction

andresailer:
This is the more sensible default and needed by the track reconstruction. Avoid problems when people don't use steering files

### DIFF
--- a/DDSim/DDSim/Helper/Action.py
+++ b/DDSim/DDSim/Helper/Action.py
@@ -8,7 +8,7 @@ class Action( ConfigHelper ):
   The default tracker and calorimeter actions can be set with
 
   >>> SIM = DD4hepSimulation()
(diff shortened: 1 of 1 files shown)


gaede:
In the ILD  default steering file we use:
`SIM.action.tracker = ("Geant4TrackerWeightedAction", {"HitPositionCombination" : 2 , "CollectSingleDeposits" :  False } )`
Should these options also be used as default ?

gaede:
@andresailer: any comment on the default configuration of the Geant4TrackerWeightedAction ?

andresailer:
HitPositionCombination 2 and CollectSingleDesposits False are the defaults. We could make it explicit if you prefer.

gaede:
I think it would be nice to make this explicit in DDSim if this can be done easily.

andresailer:
Now it is explicit

PR #24, opened 2017-02-01
New model CLIC_o3_v08

petricm:
Includes:
-redesign of tracker modules so that new updated values of module thicknesses are taken into account
-fixed the outer radius of solenoid
-removed all regions and only use global regions
-place ECal calo face at right possition
-add to plugins DD4hepVolumeManager
-remove deprecated drivers
-use global encodings strings for all trackes


petricm:
This fails, because the version of DD4hep in the CI does not include the moved drivers for Beampipe and Mask. I will merge and do a new CI build tomorrow.

PR #26, opened 2017-02-08
Added region definitions for G4 material scan.

simoniel:


andresailer:
Depends on AIDASoft/DD4hep#87 to work properly

PR #27, opened 2017-02-08
 Following the update of small ILD, the HCAL radius has been reduced …

shaojunlu:
…by ~34cm (HBU:36cm). Added one version 'SHcalSc04_Endcaps_sv01.xml' for the small AHCAL Endcaps, which has reduced one HUB at x and y, and 14 towers at x direction.(large ILD has 16 towers in endcaps from HBU.)


gaede:
Can you please also move ILD_common_v01/SHcalSc04_Endcaps.xml to ILD_l1_v01 for consistency.

PR #33, opened 2017-02-13
add vertex info in  Geant4EventReaderGuineaPig

gaede:
- adds a primary vertex for every pair particle to the event
 - requires current DD4hep HEAD with new interface:
   Geant4EventReader::readParticles(int,Vertices&, Particles& )

### DIFF
--- a/plugins/Geant4EventReaderGuineaPig.cpp
+++ b/plugins/Geant4EventReaderGuineaPig.cpp
@@ -24,13 +24,13 @@ namespace DD4hep {
   /// Namespace for the Geant4 based simulation part of the AIDA detector description toolkit
   namespace Simulation {
 
-    /// Class to populate Geant4 primaries from StdHep files.
     /**
-     * Class to populate Geant4 primary particles and vertices from a
-     * file in HEPEvt format (ASCII)
-     *
-     *  \author  P.Kostka (main author)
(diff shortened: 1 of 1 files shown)


andresailer:
Does this work if I need to simulate a full bunch crossing of background pairs, e.g., when looking at background in the BeamCal?

Could you add a test like for the other event readers in DD4hep
What about cleaning up the comments 
https://github.com/gaede/lcgeo/blob/0beb680424e65c2a8ab03ba8d0b9d2f84b1969ca/plugins/Geant4EventReaderGuineaPig.cpp#L27

gaede:
This should now work for simulating a full bunch crossing. You probably have to limit the number of particles per event using skipNEvents (!). Please tr y it.
Will add a test to DD4hep, once we move it there.

andresailer:
And what if I want to actually skip some events?

gaede:
There are no events in a pairs bg file - it is always one bunch crossing per file  - so nothing to skip.

andresailer:
But one might want to split simulation across multiple jobs, so then one wants to ignore the first N particles.

gaede:
This cannot be done then with the current implementation, there is only one parameter that we can 'mis-use' . The current design ofGeant4EventReader and Geant4InputAction is rather limited and I think we need change this eventually. There will be other cases where the readers need dedicated configuration parameters...

PR #35, opened 2017-02-15
Move cables more backwards / closer to measurements surface

petricm:
The changes to cables and air flow shell position as discussed in yesterdays meeting.


andresailer:
Ok, when ignoring whitespace changes the amount of changed lines looks reasonable
https://github.com/iLCSoft/lcgeo/pull/35/files?w=1

PR #40, opened 2017-02-24
add parameter ParticlesPerEvent to GuineaPig reader

gaede:
- parameter ParticlesPerEvent can be set in steering file with:
            SIM.readerParameters = {"ParticlesPerEvent":"3"}
- also set readerParameters for other readers for future use in ddsim

### DIFF
--- a/plugins/Geant4EventReaderGuineaPig.cpp
+++ b/plugins/Geant4EventReaderGuineaPig.cpp
@@ -48,9 +48,10 @@ namespace DD4hep {
       /// Read an event and fill a vector of MCParticles.
       virtual EventReaderStatus readParticles(int event_number,
                                               Vertices& vertices,
-                                              std::vector<Particle*>& particles);
-      virtual EventReaderStatus moveToEvent(int event_number);
-      virtual EventReaderStatus skipEvent() { return EVENT_READER_OK; }
+                                              std::vector<Particle*>& particles) override ;
+      virtual EventReaderStatus moveToEvent(int event_number) override ;
+      virtual EventReaderStatus skipEvent() override  { return EVENT_READER_OK; }
+      virtual EventReaderStatus setParameters( std::map< std::string, std::string > & parameters ) override ;
     };
   }     /* End namespace Simulation   */
 }       /* End namespace DD4hep       */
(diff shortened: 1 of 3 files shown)


gaede:
Build fails because DD4hep is too old - can this be easily updated ?
Or should we merge this PR as is ?

plugins/Geant4EventReaderGuineaPig.cpp
```diff
@@ -91,19 +92,32 @@ Geant4EventReaderGuineaPig::~Geant4EventReaderGuineaPig()    {
...
+Geant4EventReaderGuineaPig::moveToEvent(int event_number) {
+  
+  printout(DEBUG,"EventReaderGuineaPig::moveToEvent"," event_number: %d , m_currEvent %d",
+	   event_number,m_currEvent ) ;
   
-    printout(INFO,"EventReaderGuineaPig::moveToEvent"," --- Will read %d particles per event from pairs file ", m_part_num );
-    printout(INFO,"EventReaderGuineaPig::moveToEvent"," ***  Due to a workaround event numbers will start from %d  for now !!", m_part_num );
+  if( m_currEvent == 0 && event_number > 0 ){
 
+    printout(INFO,"EventReaderGuineaPig::moveToEvent"," --- Cannot skip to event %d in GuineaPig files - ignored ! ", event_number );
```
andresailer: We really need a correct implementation of skipEvent, why should this not be possible?

andresailer:
I will change the ddsim part to allow the steering of the ParticlesPerEvent for guineapig pairs files.

Maybe we could move the GuineaPigEventReader to DD4hep now...

gaede:
> I will change the ddsim part to allow the steering of the ParticlesPerEvent for guineapig pairs files.
I actually already did this - please check

andresailer:
I saw your changes.

gaede:
and ?

andresailer:
and I would do this: https://github.com/iLCSoft/lcgeo/commit/9c9abc115ef57a5092d7f6fe3993f6fc9cf98ec7

Let's you set the option also in the command line for example. Doesn't polute the main parameter space but keeps it in a separate location that can be expanded, and parameters can be set differently for different input files if they would ever be used together

plugins/Geant4EventReaderGuineaPig.cpp
gaede: Because GuineaPig files don't have "events" but only one BX, as you know...

What could be done is to skip N pair particles in the beginning, where:
 N = ( ParticlesPerEvent > 0 ?  event_number * ParticlesPerEvent  : 0 )
Not sure how useful this is, though.

andresailer: Yes, skipping over the N particles is what should be done, as was possible and very useful in Mokka.
It allows one to split the simulation of one pair background file into multiple jobs, or skip to the one particle that crashes things.

gaede: Ok, I implemented skipping for GuineaPig files.  
Splitting the simulation is fine - but one has to always look at a complete BX at a time, as the particles in the file are ordered.

gaede:
Ok, a bit heavy for handling one parameter but it is more convenient for the users like that.
We'll have to add more helper classes for the other readers eventually ...
I merge my PR and then you can rebase your stuff.

andresailer:
Should I create an issue to remind you to move this to dd4hep now that the development has finished?

PR #49, opened 2017-03-15
Update ecal

danieljeans:
by popular request, allow ECAL barrel with symmetry other than 8.

In the ECAL barrel, add possibility to include dead volume at the "inside" end of slabs (i.e. not at position of readout electronics) to simulate effect of slab retention system ("plug"), via optional parameter Ecal_Slab_Plug_length. (length of 0 used if not specified, so default behaviour should be unchanged).

### DIFF
--- a/detector/calorimeter/SEcal05_Helpers.h
+++ b/detector/calorimeter/SEcal05_Helpers.h
@@ -83,7 +83,7 @@ class SEcal05_Helpers {
   SEcal05_Helpers();
 
   ~SEcal05_Helpers() {
-    delete _layering;
+    if ( _layering ) delete _layering; _layering=NULL;
   }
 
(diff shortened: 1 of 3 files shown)


detector/calorimeter/SEcal05_Helpers.h
```diff
@@ -83,7 +88,7 @@ class SEcal05_Helpers {
   SEcal05_Helpers();
 
   ~SEcal05_Helpers() {
-    delete _layering;
+    if ( _layering ) delete _layering; _layering=NULL;
```
andresailer: This check was not necessary.
`delete NULL` is a no-op. Also setting the variable in the destructor is not necessary.
These changes introduce a warning about "misleading indentation as well.

PR #55, opened 2017-03-20
Encoder string tweaks

protopopescu:
ReleaseNotes:
 - updated encoder strings for consistency within Tracking susbsystem
 - added envelope, and type flags


protopopescu:
This was just a demo.

PR #66, opened 2017-03-29
Fill the cellEncoding string 'submodule' with 'slice_number'.

shaojunlu:
BEGINRELEASENOTES
- Fill the cellEncoding string "submodule" with "slice_number" for both GenericCalBarrel_o1_v01_geo.cpp and GenericCalEndcap_o1_v01_geo.cpp

ENDRELEASENOTES

### DIFF
--- a/detector/calorimeter/GenericCalBarrel_o1_v01_geo.cpp
+++ b/detector/calorimeter/GenericCalBarrel_o1_v01_geo.cpp
@@ -201,7 +201,8 @@ static Ref_t create_detector(LCDD& lcdd, xml_h e, SensitiveDetector sens) {
                 // Set region, limitset, and vis.
                 slice_vol.setAttributes(lcdd, x_slice.regionStr(), x_slice.limitsStr(), x_slice.visStr());
                 // slice PlacedVolume
-                layer_vol.placeVolume(slice_vol, Position(0, 0, slice_pos_z));
+                PlacedVolume slice_phv = layer_vol.placeVolume(slice_vol, Position(0, 0, slice_pos_z));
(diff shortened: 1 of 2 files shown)


shaojunlu:
@andresailer @webermat I would like to know whether we may fill the "submodule" with "slice_number" here like the PR.
If it is OK for your compact module, I would like to access the slice_number via submodule in the encoding string.

shaojunlu:
@andresailer I have generated a compact file to demonstrate how I want to try it out.
Here is the compact ILD_o4_v01:
https://github.com/shaojunlu/lcgeo_fork/tree/Usesubmodule/ILD/compact/ILD_o4_v01

submodule:2 has 48 layers and submodule:5 has 48 layers too.

Here is the original design ideas on page 6:
https://agenda.linearcollider.org/event/7512/contributions/38394/attachments/31207/46879/TC_report.pdf

If this PR will be accepted by CLIC, I will use them for testing.

shaojunlu:
@andresailer @webermat 

DDRec::LayeredCalorimeterData::Layer caloLayer

We have to fill the inner outer radiation length information, and so on, for reconstruction.
It may generate some problems for reconstruction.

It may open a window for the calorimeter collaboration to understand deeply the energy deposited in the whole calorimeter every part, which could be accessed with submodule here.

Directly to compare the visible energy and the true energy distribution in the calorimeter event by event level.

Anyway, we have to think about it further, it should be fine for the simulation level to access all the information.

Please do not merge this PR for the moment.

shaojunlu:
I have found one geometry driver which has ready filled the slice_number.
I will play with it first.
So I will close this PR.

PR #68, opened 2017-03-30
Initialise a combined RPC/Scintillator HcalBarrel simulation module. …

shaojunlu:
…Both RPC and Scintillator have 48 layers. MarlinProcessor may access RPC from slice:3 and access Scintillator from slice:6.

BEGINRELEASENOTES
- Initialise a module "ILD_o4_v01" for a combined RPC/Scintillator HcalBarrel simulation. 
   - Both RPC and Scintillator have 48 layers.
   - MarlinProcessor may access RPC from slice:3 and access Scintillator from slice:6
   - Assign different segmentations and output collections to the different sensitive type.

ENDRELEASENOTES

### DIFF
--- a/ILD/compact/ILD_o4_v01/HCalBarrel_o4_v01_01.xml
+++ b/ILD/compact/ILD_o4_v01/HCalBarrel_o4_v01_01.xml
@@ -1,0 +1,70 @@
+<lccdd>
+    <!--  Definition of global dictionary constants          -->
+    <define>
+        <constant name="HCalBarrel_layers" value="(int) 48"/>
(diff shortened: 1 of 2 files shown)


gaede:
@shaojunlu can you try and assign different segmentations and output collections to the different slices ? There is an example for doing this based on the key "layer"  which you should be able to modify to use "slice" as key:
https://github.com/AIDASoft/DD4hep/blob/master/examples/ClientTests/compact/MultiSegmentCollections.xml#L53-L65

shaojunlu:
@gaede I have updated readouts to assign different segmentations and output collections to the different sensitive type.
Please accept PR  #69 first, which update will allow to assign MultiSegmentaion and MultiCollection for DetaildShowerMode within a list.

PR #72, opened 2017-03-30
Fix overlaps

pawlikb:
BEGINRELEASENOTES
 - corrections for LHCal01.xml and LHCal_o1_v01_geo.cpp fixing overlaps.
 -  Inner cutout in LHCal envelope changed from octagon to tube, as for some reasons 
    PolyhedraRegular is not subtracted - causing lack of inner cutout in envelope and overlaps 
    with beam pipe.
-  Fix of bug in geo driver which caused  internal overlaps 
ENDRELEASENOTES


pawlikb:
I am lost .... What is the build error, to be fixed ? My local build does not have any errors

andresailer:
Might have been a problem with the build system. I restarted the builds.

PR #79, opened 2017-04-03
enforce definition of tracker volume in ddsim

gaede:
- add option to enable debug output actions for hits from Markus

BEGINRELEASENOTES
- enforce definition of tracker volume in ddsim - models need to define:
        - tracker_region_zmax
        - tracker_region_rmax
- this is needed to ensure that the MCTruth link for hits works correctly
- allow to enable additional DebugDumpActions with *enableDetailedHitsAndParticleInfo*
ENDRELEASENOTES

### DIFF
--- a/DDSim/DDSim/DD4hepSimulation.py
+++ b/DDSim/DDSim/DD4hepSimulation.py
@@ -76,7 +76,6 @@ class DD4hepSimulation(object):
     self._errorMessages = []
     self._dumpParameter = False
     self._dumpSteeringFile = False
-    self.enableDetailedHitsAndParticleInfo = False
 
     ## objects for extended configuration option
     self.output = Output()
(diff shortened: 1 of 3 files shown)


DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -76,6 +76,7 @@ def __init__(self):
     self._errorMessages = []
     self._dumpParameter = False
     self._dumpSteeringFile = False
+    self.enableDetailedHitsAndParticleInfo = False
```
andresailer: There is no argument parsing defined for this flag, so it is essentially a constant.
Should be better defined in the ParticleHandler helper
https://github.com/iLCSoft/lcgeo/blob/master/DDSim/DDSim/Helper/ParticleHandler.py
Then it will become a command line parameter automatically.
And add a docstring like 
```
self.enableDetailedHitsAndParticleInfo = False
self._enableDetailedHitsAndParticleInfo_HELP = "print out lots of info about MCHistory"
```

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -394,13 +395,48 @@ def run(self):
     part.MinDistToParentVertex= self.part.minDistToParentVertex
     part.OutputLevel = self.output.part
     part.enableUI()
+
+
+    if self.enableDetailedHitsAndParticleInfo:
```
andresailer: Move this block to the ParticleHandler helper as well.
just leave a call to the function here:
```
self.part.setDumpDetailedParticleInfo( kernel, DDG4 )
```
kernel and DDG4 should be passed as argument so we don't need to import anything in the helper, importing there would seriously slow down the response of ddsim when something goes wrong during argument parsing or when using tab completion

andresailer:
The change breaks the SiD model

gaede:
@andresailer I implemented the changes in ddsim, detailed hit info can now be enabled on the command line.  Concerning the SID models: they need to implement the tracker region - not having this is a bug, as then the MC-truth link is broken (and the MCParticle collection will be gigantic).

DDSim/DDSim/Helper/ParticleHandler.py
```diff
@@ -13,6 +13,18 @@ def __init__( self ):
     self._printEndTracking = False
     self._printStartTracking = False
     self._minDistToParentVertex = 2.2e-14*mm
+    self._enableDetailedHitsAndParticleInfo = False
+    self._enableDetailedHitsAndParticleInfo_HELP = "print out lots of info about MCHistory"
```
andresailer: if you use property, you don't need the "_HELP" variable, or vice versa., and in case of "_HELP" the original variable must not have a leading underscore, because variables with leading underscore are ignored and the property is used instead to add the argument parsing.
i.e.
```
self.enableDetailedHitsAndParticleInfo = False
self._enableDetailedHitsAndParticleInfo_HELP = "print out lots of info about MCHistory"
```

DDSim/DDSim/Helper/ParticleHandler.py
```diff
@@ -13,6 +13,18 @@ def __init__( self ):
...
+
+  @property
+  def enableDetailedHitsAndParticleInfo(self):
+    """Enamble lots of printout on simulated hits and MC-truth inforamation"""
```
andresailer: Enamble --> Enable, inforamation --> information
Unless of course you drop the property completely

Issue #56, opened 2017-03-20
Add runtime printout to ddsim


andresailer:
print the elapsed run time at the end of ddsim, similar to what Marlin does.

PR #86, opened 2017-04-07
DDSim: Add printout of execution time at the end of simulation

andresailer:
whitespace to emulate the output from DD4hep
Solves #56
```
...
Geant4Kernel     INFO  ++ Terminate Geant4 and delete associated actions.
DDSim            INFO  Execution Time: 10.28 s (User), 0.29 s (System)
```

BEGINRELEASENOTES
- DDSim: Added printout of execution time at the end of simulation

ENDRELEASENOTES


andresailer:
I think the behaviour is the same as marlin. It prints at INFO, DEBUG, VERBOSE.
It does not for WARNING,  ERROR, etc.

PR #89, opened 2017-04-13
Clean up the duplicate xml files in ILD_o1_v05

shaojunlu:
BEGINRELEASENOTES
- Reorganise the ILD_o1_v05 as optimisation models
     - clean up the duplicate xml files, which are already copied into 'ILD_common_v01' and 'extra_stuff'.
     - move the rest xml files into folder 'extra_stuff'.
     - use the commnon xml files from folder 'ILD_common_v01' as optimisation models, and SiECAL from folder 'extra_stuff'.
     - It is very clear to show that the different between ILD_o1_v05 and ILD_l1_v01 is the ECAL drivers. The rest are used the common compact files.

ENDRELEASENOTES


gaede:
@shaojunlu as just discussed, I would prefer to leave the ILD_o1_v05 model untouched as 'historic' reference and rather evolve the ILD_l/sx models. So I will close this PR.

PR #94, opened 2017-04-18
Add multi-readout drivers for ILD Hcal with Tesla geometry

gaede:
BEGINRELEASENOTES
- added versions of the ILD Hcal drivers (with Tesla geometry ) that allow to use multiple readout collections:
      - SHcalSc04_Barrel_v04
      - SHcalSc04_Endcaps_v01
      - SHcalSc04_EndcapRing_v01
- use these drivers in ILD_l4_v01
ENDRELEASENOTES


gaede:
all checks have passed - no existing drivers or models changed: will merge.

PR #96, opened 2017-04-22
Reduction of the outgoing beam tube radius to avoid overlap with BeamCal

StrahinjaLukic:
In addition:
- Correction of the QDEX1AFront link.

BEGINRELEASENOTES
- Outgoing beam tube radius was reduced by 1 mm from "BeamCal_min_z-5*mm" to 5999*mm because the BeamCal inner radius was reduced as part of the adjustments to L*=4.1m
- The protruding outer radius of QDEX1AFront link between tube sections was reduced to the outer radius of the larger tube.
- The changes affect all ILD models that use ILD_common.
ENDRELEASENOTES


gaede:
@StrahinjaLukic can you please edit the release notes and add that this affects all ILD models that use ILD_common.

StrahinjaLukic:
Ok.

PR #98, opened 2017-04-26
use correct LHCAL setup for ILD_l2_v01 model

danieljeans:
BEGINRELEASENOTES
- use new LHCal01 detector setup (was LHcal.xml)

ENDRELEASENOTES


andresailer:
I would propose to put the detector that this change affected also in the release notes.

gaede:
done.

PR #109, opened 2017-06-09
New location of Handle.inl in DD4hep

petricm:
The nightly is broken because Markus has changes the location of Handle.inl in the last [PR](https://github.com/AIDASoft/DD4hep/pull/159)

BEGINRELEASENOTES
- Update the location of DD4hep Handle.inl include file to adapt to DD4hep v00.24

ENDRELEASENOTES

CI will fail!


andresailer:
Use version of dd4hep to check which location should be used for include?

andresailer:
Added version check, should now work with CI and new nightly as well.

PR #116, opened 2017-06-22
Fix ild o4

gaede:
BEGINRELEASENOTES
- fix ILD_o4 models:
- ILD_l4_v01:  
   - reverse order of RPC and scintillator  ( RPC first)
- ILD_l/s4_v02:
   - reverse order of RPC and scintillator  ( RPC first)
   - add parameters for tracking volume and reconstruction geometry
   - add surface plugins for tracking

ENDRELEASENOTES

### DIFF
--- a/ILD/compact/ILD_common_v02/SHcalSc04_Barrel_v04.xml
+++ b/ILD/compact/ILD_common_v02/SHcalSc04_Barrel_v04.xml
@@ -25,10 +25,10 @@
       <layer repeat="Hcal_nlayers" vis="SeeThrough">
 
         <slice material="Air"      thickness="Hcal_chamber_thickness - ( HcalSD_glass_anode_thickness + HcalSD_sensitive_gas_gap + HcalSD_glass_cathode_thickness + Hcal_scintillator_thickness)" vis="Invisible" />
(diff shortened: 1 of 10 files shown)


danieljeans:
I had moved the definitions of the plugins to ILD_common_v2/plugins.xml to avoid duplication between different detector models. Do these not get picked up correctly?

For the same reason I also intended to move the definitions of     
tracker_region_rmax and tracker_region_zmax to ILD_common_v02/tracker_defs.xml
but this change did not get committed: apologies.

gaede:
Ok, I had overlooked the plugin definition in ILD_common_v2/plugins.xml, as they did not get picked up correctly from there.
 @MarkusFrankATcernch: do you know why global plugins like the SurfaceInstaller cannot be defined in an included xml file, like here:
https://github.com/iLCSoft/lcgeo/blob/master/ILD/compact/ILD_l4_v02/ILD_l4_v02.xml#L73
with:
https://github.com/iLCSoft/lcgeo/blob/master/ILD/compact/ILD_common_v02/plugins.xml
Is this a feature or a bug ?

Concerning the tracker volume defs: yes, I think it is better to have them in tracker.xml - will move them there...

PR #118, opened 2017-06-28
 Fix overlaps services 'SitDisk' with 'tube_LumiCalFront'.

shaojunlu:
BEGINRELEASENOTES
- Follow the TUBE update, and use the TUBE envelope radius to fix the services SitDisk overlaps. 
- Define the SitDisk Rmin with TUBE_IPOuterBulge_end_envradius instead of TUBE_IPOuterBulge_end_radius.
- Implement HcalBarrel_EndcapServices which also include the services coming from  Ecal and TPC.
- Implement HcalBarrel_EndcapServices parameters into both ILD_comon_v01 and ILD_common_v02.
- Add parameter 'TPC_cooling_nRings' in SServices00.xml to allow change wrt large and small ILD model for this moment.

ENDRELEASENOTES

### DIFF
--- a/ILD/compact/ILD_common_v01/SServices00.xml
+++ b/ILD/compact/ILD_common_v01/SServices00.xml
@@ -9,7 +9,7 @@
   <type_flags type=" DetType_SUPPORT " />
   
   <!-- SQL command: "select * from sit_layers;"  -->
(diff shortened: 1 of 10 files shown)


shaojunlu:
Added new implementation of HcalBarrel_EndcapServices

PR #125, opened 2017-07-04
ILD_*_v02 models: more explicit layering in SEcal*

danieljeans:
BEGINRELEASENOTES
- previously several ECAL sub-layers were combined into an averaged material to simplify simulation.
This requires to redefine this material every time thicknesses change (quite often at the moment), which was not done consistently. 
For the time being, separate the components to ensure correct material description. 
(Consider recombining when sub-layer thicknesses are fixed and stable.)
ENDRELEASENOTES

### DIFF
--- a/ILD/compact/ILD_common_v02/SEcal05_siw_Barrel.xml
+++ b/ILD/compact/ILD_common_v02/SEcal05_siw_Barrel.xml
@@ -21,7 +21,10 @@
 
       <layer repeat="Ecal_nlayers1/2" vis="SeeThrough" > <!-- 10 double SD Layers, and 5 W palte after each double SD Layer -->
         <slice material = "G4_AIR"         thickness = "Ecal_Alveolus_Air_Gap/2."                vis="Invisible" />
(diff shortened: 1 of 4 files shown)


gaede:
@danieljeans it seems that you have added an extra slice of Air with thickness `Ecal_Slab_ASIC_thickness` 
- this parameter does not seem to be defined anywhere !?
- does this not increase the total thickness  of the Ecal ?

danieljeans:
@gaede , thanks for checking.

I added the Ecal_Slab_ASIC_thickness parameter to ecal_defs.xml.

Previously 1.7mm was allocated to the "PCB", whereas the current proposal is actually 1mm PCB + 0.7mm space for ASICs, which is what is now implemented.
This ASIC space is mostly an air gap, partially covered by packaged ASICs, whose material I ignore (plastic case, small silicon chip, wire bonds).

So the total thickness is unchanged. (In fact, the driver checks that the thickness it calculates corresponds to what is defined in the "top-level" Ecal_barrel_thickness parameter.)

PR #134, opened 2017-07-12
SEcal05_ECRing.cpp : fix LayeredCalorimeterData ; correct layering (take from compact descr)

danieljeans:
BEGINRELEASENOTES
changes to SEcal05_ECRing.cpp:
- include first layer in LayeredCalorimeterData (if not preshower)
- take layering within slab from compact description to ensure correct treatment (removing some assumptions from driver): should now be identical layering to ecal endcaps from SEcal05_Endcaps
ENDRELEASENOTES

### DIFF
--- a/detector/calorimeter/SEcal05_ECRing.cpp
+++ b/detector/calorimeter/SEcal05_ECRing.cpp
@@ -146,16 +146,8 @@ static Ref_t create_detector(Detector& theDetector, xml_h element, SensitiveDete
 //====================================================================
 
   //  read parametere from compact.xml file
-  double Ecal_Alveolus_Air_Gap              = theDetector.constant<double>("Ecal_Alveolus_Air_Gap");
-  double Ecal_Slab_shielding                = theDetector.constant<double>("Ecal_Slab_shielding");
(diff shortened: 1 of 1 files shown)


shaojunlu:
Dear Daniel, 

Since you are talking about SEcal "LayeredCalorimeterData", have you checked SEcal Barrel and Endcaps? 

Do we lost the first layer information for reconstruction if this information are used by MarlinPandora? 

Do you have time to have a look at SEcal Barrel and Endcaps for "LayeredCalorimeterData" (if not preshower), too. 

Cheers 
Shaojun

danieljeans:
Hi Shaojun,
from looking at the code, I believe that the definition of LayeredCalorimeterData was and still is OK for barrel and endcap; only the EC ring was affected.
To be honest, I don't know how the LayeredCalorimeterData is used by marlinpandora. however, since only the plug was affected, I would be surprised if this has any effect on central JER. Maybe a rather (invisibly?) small degradation of the single photon energy resolution in the plug region.
Cheers,
Daniel.

shaojunlu:
Dear Daniel, 

Thank you very much for the confirmation on the SEcal Barrel and Endcaps 
Yes, I agree with you. Only ECRing may have invisibly effect. 

Cheers 
Shaojun

PR #137, opened 2017-07-13
DD4hepSim : parseOptions(argv) customize

rete:
BEGINRELEASENOTES
- Modified DD4hepSimulation.parseOptions() to receive argv argument, allowing user scripts to pass custom command line args 
ENDRELEASENOTES

### DIFF
--- a/DDSim/DDSim/DD4hepSimulation.py
+++ b/DDSim/DDSim/DD4hepSimulation.py
@@ -89,6 +89,8 @@ class DD4hepSimulation(object):
     self.filter = Filter()
     self.physics = Physics()
 
+    self._argv = None
+
     ### use TCSH geant UI instead of QT
     os.environ['G4UI_USE_TCSH'] = "1"
 
@@ -108,8 +110,12 @@ class DD4hepSimulation(object):
         self.__dict__ = obj.__dict__
     self.steeringFile = os.path.abspath(sFileTemp)
 
-  def parseOptions(self):
+  def parseOptions(self, argv=None):
     """parse the command line options"""
+
+    if argv is None :
+        self._argv = list(sys.argv)
+
     parser = argparse.ArgumentParser("Running DD4hep Simulations:",
                                      formatter_class=argparse.RawTextHelpFormatter)
 
@@ -117,7 +123,7 @@ class DD4hepSimulation(object):
                         help="Steering file to change default behaviour")
 
     #first we parse just the steering file, but only if we don't want to see the help message
-    if not any( opt in sys.argv for opt in ('-h','--help')):
+    if not any( opt in self._argv for opt in ('-h','--help')):
       parsed, _unknown = parser.parse_known_args()
       self.steeringFile = parsed.steeringFile
       self.readSteeringFile()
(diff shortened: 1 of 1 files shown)


rete:
Pull-requested to early... I need modifications. Sorry ... do not merge

andresailer:
I don't think any of this necessary, you can directly modify `sys.argv` in your script before calling `parseOptions`.

gaede:
Andre, we had discussed this option but came to the conclusion that it is much cleaner to do it like this, as the calling program might also have it's own set of argv that than will have to be parsed before (and not accessed later in the code anywhere )....

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -108,16 +110,18 @@ def readSteeringFile(self):
         self.__dict__ = obj.__dict__
     self.steeringFile = os.path.abspath(sFileTemp)
 
-  def parseOptions(self):
+  def parseOptions(self, argv=sys.argv):
```
andresailer: This is bad due to the way python treats default arguments.
Use `None` as the default argument and then
```
if argv is None:
  self._argv = list(sys.argv)
```

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -108,16 +110,18 @@ def readSteeringFile(self):
...
     """parse the command line options"""
+    self.argv = argv
```
andresailer: This just assigns the "reference" to argv so `self.argv` and `argv` are referring to the same object.
Should do
`self._argv = list(argv)` to avoid surprises later
```
>>> l = []
>>> b = l
>>> b.append("arg")
>>> l
['arg']
>>> c = list(l)
>>> c
['arg']
>>> l.append("arg2")
>>> c
['arg']
>>> l
['arg', 'arg2']
```

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -635,7 +639,7 @@ def __setupSensitiveDetectors(self, detectors, setupFuction, defaultFilter=None)
       for pattern in self.action.mapActions:
         if pattern.lower() in det.lower():
           action = self.action.mapActions[pattern]
-          print  '       replace default action with : ' , action 
+          print  '       replace default action with : ' , action
```
andresailer: Unnecessary whitespace change, please exclude from your commit

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -89,6 +89,8 @@ def __init__(self):
     self.filter = Filter()
     self.physics = Physics()
 
+    self.argv = None
```
andresailer: use 
```
self._argv
```
because, for example, we don't need this to show up the steering file when dumped

andresailer:
Please don't forget to squash your commits and remove the "WIP" from the title when you are done.

andresailer:
Please squash your commits

rete:
I followed the guideline from the CONTRIBUTING.md file (NoSY) but I didn't know what squashing commits mean so I didn't do it. I merged my local branch into the master, pushed it to my fork and deleted my local branch.
Now I don't know how to squash them. Sorry...

andresailer:
Ahh, you brought your branch back.
```
git checkout ddsim-argv-parsing
git branch backup-parsing # just in case make a "copy" of the current branch
git reset --hard 0a5d736
git push -f downstream ddsim-argv-parsing ## depends what you call your remote
```

PR #138, opened 2017-07-14
new ILD_?5_v02 models with hybrid ecal structure. Cosmetic fix to ILD_?4 hcal description.

danieljeans:
ILD_?5_v02 models require update to MegatileLayerGridXY class in DD4hep/segmentations (for which I have just made a pull request)

BEGINRELEASENOTES
- first implementation of ILD_?5_v02 models, which have a hybrid readout ECAL structure with both silicon and scintillator layers. Otherwise, should be indentical to the ILD_?4_v02 models.
- removed a duplicated definition of Hcal_readout_segmentation_slice_* (was in both ILD_*xml and hcal_defs.xml; now only in the latter). It gave correct results before, but was confusing.
ENDRELEASENOTES

### DIFF
--- a/ILD/compact/ILD_common_v02/SEcal05_hybrid_Barrel.xml
+++ b/ILD/compact/ILD_common_v02/SEcal05_hybrid_Barrel.xml
@@ -1,0 +1,95 @@
+<lccdd>
+  <detectors>
+
+    <detector name="EcalBarrel" type="SEcal05_Barrel" id="ILDDetID_ECAL" readout="EcalBarrelCollection" vis="BlueVis" >
+
+      <comment>EM Calorimeter Barrel</comment>
+
+      <envelope vis="ILD_ECALVis">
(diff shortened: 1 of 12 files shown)


danieljeans:
this fails because the required update to the segmentation is not yet in DD4hep

gaede:
@danieljeans can you please move the SEcal05 drivers to SEcal06 and remove any (even trivial) changes to any of the existing (ILD_*v02) models from this PR ?
This way, we can try and make a patch release for lcgeo and include it in a small patch to v01-19-03. For this to work, however, you would need to copy the new MegatileLayerGridXY
segmentation to the lcgeo package and namespace.
I cannot (easily) include a new DD4hep release in  a patch to v01-19-03 as there have been massive changes to DD4hep recently.
The important thing about preparing such a patch PR, would be that none of the existing models in lcgeo are modified.
If this is not possible, we'd have to wait for the next full iLCSoft release ( September ?).

danieljeans:
OK, I agree. 
I closing this pull request, and reformat it in a cleaner way to avoid touching presently used code.

PR #141, opened 2017-07-17
New model CLIC_o3_v12

petricm:
The starting position of outer tracker barrel was 1264mm but 30.1mm times 84 modules gives a starting position of 1264.2*mm. This bug happened sometime around when we truncated sub milometer precision from the tracker design, but this causes a problem since the delta(z_incr) in module positioning is

```cpp
double z_incr = nz > 1 ? (2.0 * z0) / (nz - 1) : 0.0;
``` 

BEGINRELEASENOTES
- New model CLIC_o3_v12 that fixes an overlap that was considered a false positive in outer tracker barrel due to shift in z0
- In file `OuterTracker_o2_v06_01.xml` L6 value of `OuterTracker_Barrel_half_length` change from 1264mm -> 1264.2mm

ENDRELEASENOTES

### DIFF
--- a/CLIC/compact/CLIC_o3_v12/BeamCal_o1_v01_01.xml
+++ b/CLIC/compact/CLIC_o3_v12/BeamCal_o1_v01_01.xml
@@ -1,0 +1,99 @@
+<lccdd>
+    
+    <define>
+        <constant name="BeamCal_spanning_angle" value="350*degree"/>
+    </define>
+    
+    <readouts>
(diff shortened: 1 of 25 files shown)


petricm:
`/geometry/test/run` on the whole geometry does not produce any overlap

gaede:
@petricm should this be merged ?
I need to  create a new release tag for ILD now (soon) and this could go in, if needed ...

PR #144, opened 2017-07-19
Fix VXD04 overlap by changing offset_phi.

shaojunlu:
BEGINRELEASENOTES
- Fix VXD04 overlap by changing offset_phi.

ENDRELEASENOTES


shaojunlu:
You may check the update by comparing the following two pictures.
Overlap: [Before changing offset_phi](http://www.desy.de/~shaojun/VXD04/VXD_All_v00.png)
Fixed: [After changing offset_phi](http://www.desy.de/~shaojun/VXD04/VXD_All_v01.png)

PR #150, opened 2017-07-28
Newendcap

TiborILD:
BEGINRELEASENOTES
- added inner box cut off for Hcal_Endcaps_SD_v01
- added SDHCAL Barrel module with TESLA geometry

ENDRELEASENOTES


TiborILD:
another trial to get it right ....

gaede:
Looks much better. But you don't seem to be using your new drivers anywhere now, i.e. 
 ILD/compact/ILD_l2_v02/ILD_l2_v02.xml has not been updated - is this on purpose ?

TiborILD:
well, in fact what is the purpose of those 2 versions  ILD_l2_v01 and ILD_l2_v02 ?
Seems that v02 is not the same as v01, just with different versions of the same drivers ?

gaede:
Tibor, why did you now remove again your additions ?
The only change that is left in this PR is now 35ce09838cb87aebe3562374dbdd4f56c881ceee, ie the one to  ILD/compact/ILD_l2_v02/ILD_l2_v02.xml....
The _v01 verisions are outdated ( they have the old thinner Ecal), so we only use _v02 models from now on.

TiborILD:
well, it was too hot here ....
I've thought that I commited wrong files.
What is the right way to include different versions of the same subdetector like Hcal_Barrel_SD_v01
  and its version v02 ? What naming convention for ILD_l2_  ...... ?

PR #159, opened 2017-09-05
Fix Hcal_readout_segmentation_slice_barrel for ILD_l4_v02.

shaojunlu:
BEGINRELEASENOTES
- Fix ILD_l4_v02 target readout slice number for AHCAL.
  - after re-order the Hybird HCAL layers, we should also update Hcal_readout_segmentation_slice_barrel to "3" as "Hcal_readout_segmentation_slice_endcap" now.

ENDRELEASENOTES


shaojunlu:
@TiborILD After the simulation done, when SDHcal want to run the reconstruction,  both "Hcal_readout_segmentation_slice_barrel" and "Hcal_readout_segmentation_slice_endcap" have to be changed to "1" to generate gear file and to run reco on simulation files.
Please validate your gear file and reconstruction.
__By the way, DO NOT need to re-run simulation! Simulation will run only once, and used by both!__

PR #171, opened 2017-10-06
Remove includes of removed files

andresailer:
BEGINRELEASENOTES
- Drop unused and no longer existing header includes AidaSoft/DD4hep#241

ENDRELEASENOTES


andresailer:
CI fails because of the change for the TPC zMin, will merge so the nightly build can create a new release

PR #173, opened 2017-10-11
assign SurfaceType::Unbounded to TPC surfaces

gaede:
BEGINRELEASENOTES
- use unbounded surfaces in the ILD TPC
    - assign surfaces only to the forward half of the TPC
    - as they are unbounded they will also be 'visible' in the backward half

ENDRELEASENOTES

requires https://github.com/AIDASoft/DD4hep/pull/244


gaede:
failed due to missing AIDASoft/DD4hep#244 - will merge...

PR #196, opened 2018-02-05
Cmd line extension

rete:
BEGINRELEASENOTES
- Added new command line parameters to ddsim
    - --eventParameters (list) to write event parameters to every output lcio events
    - --runNumberOffset (int) to offset the run number counter in lcio output file
    - --eventNumberOffset (int) to offset the event number counter in lcio output file

ENDRELEASENOTES

### DIFF
--- a/DDSim/DDSim/DD4hepSimulation.py
+++ b/DDSim/DDSim/DD4hepSimulation.py
@@ -44,6 +44,7 @@ from DDSim.Helper.Filter import Filter
 from DDSim.Helper.Physics import Physics
 from DDSim.Helper.GuineaPig import GuineaPig
 from DDSim.Helper.LCIO import LCIO
+from DDSim.Helper.Meta import Meta
 
 import os
 import sys
@@ -73,9 +74,6 @@ class DD4hepSimulation(object):
     self.vertexSigma = [0.0, 0.0, 0.0, 0.0]
     self.vertexOffset = [0.0, 0.0, 0.0, 0.0]
     self.enableDetailedShowerMode = False
-    self.eventParameters = []
-    self.runNumberOffset = 0
-    self.eventNumberOffset = 0
 
     self._errorMessages = []
     self._dumpParameter = False
@@ -90,6 +88,7 @@ class DD4hepSimulation(object):
     self.action = Action()
     self.guineapig = GuineaPig()
     self.lcio = LCIO()
+    self.meta = Meta()
 
     self.filter = Filter()
     self.physics = Physics()
@@ -197,15 +196,6 @@ class DD4hepSimulation(object):
 
     parser.add_argument("--dumpSteeringFile", action="store_true", dest="dumpSteeringFile", default=self._dumpSteeringFile,
                         help="print an example steering file to stdout")
-    
-    parser.add_argument("--eventParameters", nargs='+', dest="eventParameters", action="store", default=[],
-                        help="Event parameters to write in every event. Use C/F/I ids to specify parameter type. E.g parameterName/F=0.42 to set a float parameter" )
-    
-    parser.add_argument("--runNumberOffset", action="store", dest="runNumberOffset", default=self.runNumberOffset, type=int,
-                        help="The run number offset to write in slcio output file. E.g setting it to 42 will start counting runs from 42 instead of 0")
-
-    parser.add_argument("--eventNumberOffset", action="store", dest="eventNumberOffset", default=self.eventNumberOffset, type=int,
-                        help="The event number offset to write in slcio output file. E.g setting it to 42 will start counting events from 42 instead of 0")
 
     #output, or do something smarter with fullHelp only for example
     ConfigHelper.addAllHelper(self, parser)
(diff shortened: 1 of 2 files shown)


rete:
Requires [AIDASoft/DD4hep#301](https://github.com/AIDASoft/DD4hep/pull/301)

rete:
Unfortunately, Andre is in holiday until 12th February... Let's wait

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -636,6 +654,52 @@ def __addParametersToRunHeader( self ):
...
     return runHeader
 
+
+  def __addParametersToEvent(self, parameters):
+    """ extract event parameters from the command line, to write in the header
+    of each event.
+    
+    :param parameters: list of command line parameters to parse
+    """
+    print parameters
```
andresailer: remove the `print`

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -636,6 +654,52 @@ def __addParametersToRunHeader( self ):
...
+    stringParameters, intParameters, floatParameters, allParameters = {}, {}, {}, []
+    for p in parameters:
+      parameterAndValue = p.split( "=", 1 )
+      if len(parameterAndValue) is not 2:
```
andresailer: `is` should be used to compare identify of objects, not for numbers or strings, better use `!=`

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -636,6 +654,52 @@ def __addParametersToRunHeader( self ):
...
+        self._errorMessages.append("ERROR: Couldn't decode event parameter '%s'" %(p))
+        raise SyntaxError("ERROR: Couldn't decode event parameter '%s'" %(p))
```
andresailer: This whole function comes after the errorMessages were used up, so it is not necessary to store things in the error messages. Or add another check for error messages after this function was called.
So two possibilities:
1) remove the appending to errorMessages and remove the `continue` (because the code after `raise` is never executed)
2) remove the `raise` and add a check for error messages after the function was called

Same goes for the other places where errorMessages/exceptions are used.

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -636,6 +654,52 @@ def __addParametersToRunHeader( self ):
...
+      pvalue = parameterAndValue[1]
+      if ptype.lower() not in ["c", "f", "i"]:
+        self._errorMessages.append("ERROR: Event parameter '%s' with invalid type '%s'" %(pname, ptype))
+        raise ValueError("ERROR: Event parameter '%s' with invalid type '%s'" %(pname, ptype))
+        continue
+      if pname in allParameters:
+        self._errorMessages.append("ERROR: Event parameter '%s' specified twice" %(pname))
+        raise RuntimeError("ERROR: Event parameter '%s' specified twice" %(pname))
+        continue
+      if not len(pvalue):
```
andresailer: `if not pvalue`

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -636,6 +654,52 @@ def __addParametersToRunHeader( self ):
...
+        self._errorMessages.append("ERROR: Event parameter '%s' has empty value" %(pname))
+        raise RuntimeError("ERROR: Event parameter '%s' has empty value" %(pname))
+        continue
+      allParameters.append(pname)
+      print "Event parameter '%s', type '%s', value='%s'"%(pname, ptype, pvalue)
+      if ptype.lower() is "c":
```
andresailer: use `==` not `is`

rete:
Sorry for the delay.

It seems that I can move the arguments `eventNumberOffset` and `runNumberOffset` to a Meta helper class but not `eventParameters` argument because this argument expects a list of values. The current scheme of adding cmd line arguments from helper classes doesn't allow for arguments with the option `nargs='+'` ([here](https://github.com/rete/lcgeo/blob/1c63d8229c250d7626fd4d99e2517a4142ac76e0/DDSim/DDSim/DD4hepSimulation.py#L560)).

Does it makes sense to move two arguments in a meta helper class but not the third one ?
Should I keep all of them in the main section ?

andresailer:
I will rework the way helper arguments are passed to argparse to allow arbitrary parameters like nargs.

rete:
OK. Should I wait for your update to continue this implementation or merge it like this (with a few modification to cope with your requests also) and provide a future PR to reshuffle this ?

andresailer:
give me a few more minutes

andresailer:
have a look at #203

rete:
Make the trick yes !

rete:
Can be merged on my side, if approved.

DDSim/DDSim/DD4hepSimulation.py
```diff
@@ -618,6 +623,41 @@ def __addParametersToRunHeader( self ):
 
     return runHeader
 
+
+  def __addParametersToEvent(self, parameters):
```
andresailer: Please move this function to `Meta`, and maybe call it `getEventParameters` or `parseEventParameters`

PR #199, opened 2018-02-15
ILD: add material to represent FF magnets to beampipe description

danieljeans:
BEGINRELEASENOTES
- implement downstream magnets in ILD models:
    - thickened beampipe in regions of final focus magnets
    - should reproduce the Mokka models
    - currently verifying positions with accel. experts, so details may still change
ENDRELEASENOTES


danieljeans:
please wait before merging: I want to crosscheck the magnet positions with the latest final focus design.

danieljeans:
- updated positions and length of magnets according to Okugi-san's diagram of forward region
- moved definition of magnet positions and dimensions to tube_defs.xml

I'm happy for this to be merged now.

PR #209, opened 2018-03-12
ILD_?5_v02 models: update forward region

danieljeans:
BEGINRELEASENOTES
- new driver for tube of square cross-section (BoxSupport_o1_v01_geo.cpp)
- new driver for yoke endcaps with square central hole (Yoke06_Endcaps.cpp)
- adjust description of QD0 and first extraction quad, and add its cryostat
- remove more upstream magnets
- include forward region support tubes: around QD0, outer tube froum outside -> LHCAL, and between ecal endcap and ring
- update yoke design to fit with this support structure
- implement the above in the ILD_?5_v02 models
ENDRELEASENOTES


gaede:
@danieljeans the surfaces are only needed for the tracking - so no need to add it to this support structure.

danieljeans:
@shaojunlu thanks for reporting the volume overlap. I have now defined the portion of the support tube around the ECAL ring with respect to the ECAL parameters to avoid any overlaps.

gaede:
@danieljeans should this be merged for testing ?

ILD/compact/ILD_l5_v02/ILD_l5_o1_v02.xml
```diff
@@ -83,26 +83,7 @@
     <plugin name="InstallSurfaceManager"/>
   </plugins>
 
-  <fields>
-    <field type="solenoid" name="GlobalSolenoid" inner_field="Field_nominal_value"
-           outer_field="outerField_nominal_value" zmax="Coil_half_length"
-           inner_radius="Coil_inner_radius"
-           outer_radius="world_side" />
+  <include ref="../ILD_common_v02/Field_Solenoid_Ideal.xml"/>
+  <include ref="../ILD_common_v02/Field_FwdMagnets_Ideal_250GeV.xml"/>
```
rete: Shouldn't we put the 500 GeV magnet here for the incoming production ?

danieljeans:
I think this is ready for testing. remaining open points from my point of view:

- decide whether to keep fwd magnetic fields in the main "ideal solenoid" simulation models.
If we do, we should strictly speaking define different models for 250/500 GeV.
Since these fields will have (I suspect) have negligible impact on the general physics samples, I would suggest not including them for simplicity.

- define new ILD models with detailed realisitic fieldmap, antiDID, fwd fields, for L/S @ 250/500 GeV
For this we need the realistic  solenoid and antiDID field maps for the small model, which as far as I can see are not available.

ILD/compact/ILD_l5_v02/ILD_l5_o1_v02.xml
danieljeans: A good point. For the main production with ideal fields, I would even suggest not including the fwd magnet fields at all to avoid different models for different energies.

PR #215, opened 2018-04-26
new FCCee_o1_v03 detector model; 

oviazlo:
BEGINRELEASENOTES
- new FCCee_o1_v03 detector model:
  - extend ECAL endcap
  - shrink HCAL ring
  - reduce magnetic field in barrel yoke from 1.5 T to 1.0 T
  - fix in HCAL layer layout (swap steel absorber with steel wall of cassette)

ENDRELEASENOTES


oviazlo:
file name is fixed. test should be successful now

PR #224, opened 2018-06-26
revert https://github.com/iLCSoft/lcgeo/pull/219

gaede:
BEGINRELEASENOTES
 - make lcgeo (v00-16-03 and higher) compatible w/ DD4hep v01-17-0X releases
 -  revert https://github.com/iLCSoft/lcgeo/pull/219
      - use `DD4hep` rather than `dd4hep` in ddsim python

ENDRELEASENOTES


petricm:
This PR will fails since ti tests against the HEAD of DD4hep which does not have `DD4hep.py`

gaede:
Right. Will merge...

PR #226, opened 2018-07-12
new FCCee_o1_v04 detector model with updated MDI and shielding from Anna

oviazlo:
BEGINRELEASENOTES
- new FCCee_o1_v04 detector model
- update on MDI and shielding for CLD (FCCee) detector model (from Anna Kolano)
- new driver: detector/other/Mask_o1_v01_noRot_geo.cpp

ENDRELEASENOTES


andresailer:
What is different in the new mask driver compared to the existing one?
Also, the mask drivers in lcgeo are deprecated and were moved to DD4hep proper, so this new driver should also go to DD4hep.

petricm:
What is the qualitative difference of this noRot driver?

AMKolano:
This driver  has only the mirroring option for the masks (without the additional 180° rotation around z-axis - noRot). I needed it for asymmetric shield placement.

petricm:
I will close this, reopen a new PR once the changes implemented as agreed.

PR #234, opened 2019-02-18
SortingPolicyPlugin

andresailer:
BEGINRELEASENOTES
- Plugin: lcgeo_LinearSortingPolicy : new plugin to set the sorting policy for surfaces identified by placement path based on their z position and a linear function, requires AIDASoft/DD4hep#486
- CLIC_o3_v14: add use of lcgeo_LinearSortingPolicy for TrackerEndcaps
- TrackerEndcapSupport_o1_v01: attach support volumes to different DetElements so we can attach different extensions to them

ENDRELEASENOTES

### DIFF
--- a/detector/other/TrackerEndcapSupport_o1_v01.cpp
+++ b/detector/other/TrackerEndcapSupport_o1_v01.cpp
@@ -70,6 +70,9 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector)
     Tube    l_tub(rmin, rmax, layerWidth, 2 * M_PI);
     Volume  l_vol(l_nam, l_tub, air);
     l_vol.setVisAttributes(theDetector, x_layer.visStr());
+
+    DetElement layer_pos(sdet, l_nam + "_pos", l_num);
+
     for (xml_coll_t j(x_layer, _U(slice)); j; ++j, ++s_num)  {
       xml_comp_t x_slice = j;
       double thick = x_slice.thickness();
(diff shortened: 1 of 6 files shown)


petricm:
Shouldn't this rather be CLIC_o3_v15?

andresailer:
Except for the split in the support material it does not change any volumes. Anything simulated with previous CLIC_o3_v14 is compatible with the changed xml files.

detector/other/TrackerEndcapSupport_o1_v01.cpp
```diff
@@ -70,6 +70,13 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector)
     Tube    l_tub(rmin, rmax, layerWidth, 2 * M_PI);
     Volume  l_vol(l_nam, l_tub, air);
     l_vol.setVisAttributes(theDetector, x_layer.visStr());
+
+    DetElement layer_pos(sdet, l_nam + "_pos", l_num);
+    pv = envelope.placeVolume(l_vol, Position(0, 0, zmin + layerWidth / 2.));
+    pv.addPhysVolID("layer", l_num);
+    pv.addPhysVolID("barrel", 1);
+    layer_pos.setPlacement(pv);
+
```
eleogran: what changed here? did you just move this block before the 'for'?

andresailer: Yes, so that I could assign the surface to the `layer_pos` DetElement

PR #247, opened 2021-08-30
CI: replace travis with github actions

andresailer:
BEGINRELEASENOTES
- CI: Run against LCG_100, clang10, gcc10 and LCG_99python2 gcc8
- CMake: restructure main CMake file, more use of targets
- Remove requirement for Boost, wasn't actually used for some time (still needed by DD4hep)
- GenericEndcapCalo: move setting of sensitive type because of error in SID simulation with newer ddsim

ENDRELEASENOTES


CMakeLists.txt
```diff
@@ -83,24 +65,24 @@ file(GLOB G4sources
...
-
-target_link_libraries(${PackageName}G4 ${DD4hep_LIBRARIES} ${DD4hep_COMPONENT_LIBRARIES} ${ROOT_LIBRARIES} ${Geant4_LIBRARIES} ${LCIO_LIBRARIES})
+#make sure DD4hep comes before any other includes, fix problems with using DD4hep not from LCG view, doesn't matter
+#otherwise
+include_directories(BEFORE SYSTEM $<TARGET_PROPERTY:DD4hep::DDCore,INCLUDE_DIRECTORIES>)
 
+include_directories( SYSTEM
+  ${LCIO_INCLUDE_DIRS}
+  Boost::boost
+ )
```
tmadlener, line 85: What is the reason for not having these includes also attached to targets?

andresailer: Do you mean why not use target_include_directories?

tmadlener: Yes, exactly.

andresailer: just laziness. I moved to target_include_now.
Also, boost isn't needed at the moment.
I don't seem to need the BEFORE dd4hep anymore...

andresailer:
Running here: https://github.com/andresailer/lcgeo/runs/3482606283

tmadlener (approved):
Thanks for also going over the cmake config.

PR #253, opened 2022-06-13
Move SiD_o2_v04 beampipe constants to global list.

vvolkl:
This fixes an error in the key4hep builds:

```
XercesC FATAL +++ FATAL Error at file 
"/tmp/gitlab-runner/spack-stage/spack-stage-lcgeo-commit.032c4cc7bc8ed6be1071e931dccf904fb7976b8a-2javekkqt44e34d47dhiva3wh3srpqor/spack-src/SiD/compact/SiD_o2_v04/BeamPipe_o2_v04.xml", Line 4 Column: 1 
Message:comment or processing instruction expected
```

Most likely due to a newer version of dd4hep used there not available in the lcgeo CI.

EDIT: The error is tied to Xerces-C being used by DD4hep, which is not the case for the iLCSoft nightlies. 

BEGINRELEASENOTES
- Move SiD_o2_v04 beampipe constants to global list to fix an error in key4hep builds

ENDRELEASENOTES

### DIFF
--- a/.github/workflows/key4hep.yml
+++ b/.github/workflows/key4hep.yml
@@ -1,0 +1,43 @@
+name: key4hep-stack
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        SETUP: ['/cvmfs/sw.hsf.org/key4hep/setup.sh', '/cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh']
+    steps:
+    - uses: actions/checkout@v2
+    - uses: cvmfs-contrib/github-action-cvmfs@v2
+    - name: Start container
+      run: |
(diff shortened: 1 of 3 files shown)


andresailer:
Please add which detector this change belongs to to the Release Notes
Also a bit puzzling why the error is about Line 4.

vvolkl:
Done!

And yes, it's not the most helpful error message.

protopopescu (approved):
I agree with moving the SiD_o2_v04 beam pipe constants to the main XML

andresailer:
@vvolkl The geant4 data files are not available in the key4hep setup

vvolkl:
@andresailer The key4hep setup is ok, but the spack-installed `${G4INSTALL}/bin/geant4.sh` is not. It sets the directory for all datafiles to the (empty) install prefix of the geant4-data bundle package. I'll try to fix this.

vvolkl:
The geant data files on cvmfs should be fixed. Closing and re-opening to trigger CI.

PR #261, opened 2022-11-25
Update TPCSDAction

danieljeans:
BEGINRELEASENOTES
- bug fix: no longer carry over energy to next event (sometimes affected lowpt hits)
- address one FIXME: create highpt hits even if steps have not passed the padrow centre
- some slight code cleanup and reorganisation to hopefully make it clearer to read
ENDRELEASENOTES

PR #262, opened 2023-01-11
Update tpcsd

danieljeans:
BEGINRELEASENOTES
- TPCSD: Fix crash when "previous step" was undefined

ENDRELEASENOTES


tmadlener:
There is currently a merge conflict, because #261 was squashed before merging, and so git gets confused by the presence of the old commits. Could you resolve these conflicts? I think the easiest is probably to simply cherry-pick the two new commits from this branch.

danieljeans:
Thanks for explaining the issue. I will close this and open a new one with just the new commits.

danieljeans:
cleaned up pull request : https://github.com/iLCSoft/lcgeo/pull/263

PR #263, opened 2023-01-13
Update tpcsd2

danieljeans:
BEGINRELEASENOTES
- deal with case when previousStep is not yet defined (could cause crash)
- add timing information to the lowPt hits
ENDRELEASENOTES


danieljeans:
cleaned up version of https://github.com/iLCSoft/lcgeo/pull/262

tmadlener (approved):

PR #265, opened 2023-02-14
Rename TPCHit->RawTimeSeries

jmcarcell:
BEGINRELEASENOTES
- Rename TPCHit -> RawTimeSeries

ENDRELEASENOTES

Goes together with https://github.com/key4hep/EDM4hep/pull/179

There is only left but I think this is LCIO TPCHits: https://github.com/key4hep/k4geo/blob/a8bcaa529b6f5213e054b9ecf41367b6d3a49bd4/example/ced2go-template.xml#L323


andresailer:
That TPCHits is the name of a collection, and not a type, so can remain unchanged.

Issue #269, opened 2023-05-16
CLD DetectorModel with LAr Calorimeter: CLD_o4_v01 (5?)


andresailer:
* Create a CLD Detector model based on the detector model to be created in #268 where the ECal Barrel (initially) is replaced by a LAr calorimeter, using the driver from https://github.com/HEP-FCC/FCCDetectors/blob/main/Detector/DetFCChhECalInclined/src/ECalBarrelInclined_geo.cpp

SwathiSasikumar:
A new pull request opened in k4geo by adding ```FCCee_o3_v06``` with CLD detector model along with LAr as calorimeter: https://github.com/key4hep/k4geo/pull/270.  All the steering xml files and driver from  https://github.com/HEP-FCC/FCCDetectors/blob/main/Detector/DetFCChhECalInclined/src/ECalBarrelInclined_geo.cpp are included

Zehvogel:
I think this can be closed as the model was merged already?

jmcarcell:
> I think this can be closed as the model was merged already?

I think so too, closing.

PR #270, opened 2023-05-19
Adding the CLD LAr scripts with driver file

SwathiSasikumar:
A folder ```FCCee_o3_v06``` added to FCCee compact files folder with steering files for CLD with LAr as the ECAL. The driver file ```ECalBarrelInclined_geo.cpp``` added along with other cpp files under ```detector/calorimeter```.  

BEGINRELEASENOTES
- ...

ENDRELEASENOTES

### DIFF
--- a/FCCee/compact/FCCee_o3_v06/BeamInstrumentation_o3_v01_overlap.xml
+++ b/FCCee/compact/FCCee_o3_v06/BeamInstrumentation_o3_v01_overlap.xml
@@ -1,0 +1,26 @@
+<lccdd>
+
+    <detectors>
+        <comment>Beampipe Instrumentation</comment>
+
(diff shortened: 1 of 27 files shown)


jmcarcell:
The text between BEGINRELEASENOTES and ENDRELEASENOTES goes into the release notes when a tag is made, see https://github.com/key4hep/k4geo/releases, and what's left out doesn't go into the notes. Please edit the text of the PR, this is common to all the key4hep repos.
Less important, if in your text you mention the issue https://github.com/key4hep/k4geo/issues/269 by pasting the link to the issue (like I do here) it will appear in the issue that it has been mentioned here (with a link to this PR) so you don't need to copy what you say here also in the issue.

Issue #272, opened 2023-05-24
CMake tests fail when repo is compiled wth key4hep stack


atolosadelgado:
Hi,

there are 3 tests in k4geo that fail when using key4hep stack. The following code reproduce the error in lxplus

```
source /cvmfs/sw.hsf.org/key4hep/setup.sh 
cd /tmp/atolosad/
git clone -b arcdetector git@github.com:atolosadelgado/k4geo.git
cd k4geo/
cmake3 -B build -S . -DCMAKE_INSTALL_PREFIX=install
cmake --build build -- install
ctest -R t_test_steeringFile -V
ctest -R t_SensThickness_Clic -V
```

The test number 12, with name t_test_steeringFile, corresponding to the steering file example fails with the following error. The variable `eps_min` is defined in the steering file and seems to be out of range. 

```
12: -------- EEEE ------- G4Exception-START -------- EEEE -------
12: 
12: *** ExceptionHandler is not defined ***
12: *** G4Exception : Geometry001
12:       issued by : G4FieldManager::SetMinimumEpsilonStep
12: Incorrect proposed value of eps_min = 1
12:  Its value is outside the permitted range from 2.22045e-13  to 0.01
12:  Clarification: 
12:   b) It must be smaller than (or equal)     0.01 to ensure robustness of integration - (FAILED)
12: 
12: *** Fatal Exception ***
12: -------- EEEE ------- G4Exception-END -------- EEEE -------
```

The tests 13 and 14, with names t_SensThickness_*, fail with the following error:

```
14: /tmp/atolosad/k4geo/install/bin/TestSensThickness: error while loading shared libraries: libxerces-c-3.2.so: cannot open shared object file: No such file or directory
```

Is this an error caused by the current key4hep stack? 

The tests pass when I use my local installation of DD4hep.

Some of the tests of my PR are failing, and trying to figure out what is failing I found these issues. Please close the issue if it is considered non important

 - OS version: centos7 (lxplus)
 - Compiler version: gcc (Spack GCC) 11.2.0
 - Package version: master, HEAD
 - Reproduced by:  already explained
 - Input: this repository
 - Output: already given
 - Goal: make test pass

jmcarcell:
[...]

PR #274, opened 2023-05-26
Change eps_min to .001 in the example steering file

jmcarcell:
to make the test t_test_steeringFile pass, see https://github.com/key4hep/k4geo/issues/272

BEGINRELEASENOTES
- Change eps_min to .001 in the example steering file

ENDRELEASENOTES


jmcarcell:
Ah this is annoying and I saw it the other day, downstream-build won't work because the name of the repo (k4geo) is different from the name of the package in spack (lcgeo), to be fixed...

PR #275, opened 2023-06-12
Switch to dd4hep::CellID for compatibility with dd4hep

tmadlener:
BEGINRELEASENOTES
- Move to `dd4hep::CellID` for cellIDs, after [AIDASoft/DD4hep#1125](https://github.com/AIDASoft/DD4hep/pull/1125)

ENDRELEASENOTES

~@andresailer To make this slightly less brittle, I would need a new DD4hep tag, so that I can put in some pre-processor conditions.~ Actually scratch that, the `CellID` typedef exists in older versions as well, so this should be transparent.

### DIFF
--- a/detector/other/Beampipe_o1_v01_geo.cpp
+++ b/detector/other/Beampipe_o1_v01_geo.cpp
@@ -62,7 +62,7 @@ public:
   void setHalfLength( double half_length){
     _half_length = half_length ;
   }
-  void setID( dd4hep::long64 id ) { _id = id ; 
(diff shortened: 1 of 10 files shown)


tmadlener:
@jmcarcell It seems the key4hep based CI workflows are failing due to some missing libraries. On the other hand, the one based on the CLIC nightlies are OK, and they are building against the latest dd4hep where these changes are necessary to even build. Since the CLIC nightlies will stop working without these changes, I would be in favor of merging this quickly, unless you already have a fix in mind for the key4hep builds.

PR #276, opened 2023-06-13
[FCCDetectorsMigration] Noble liquid ecal barrel

BrieucF:
BEGINRELEASENOTES
- Migrate the noble liquid ECAL barrel C++ builder from [FCCDetectors](https://github.com/HEP-FCC/FCCDetectors/blob/main/Detector/DetFCChhECalInclined/src/ECalBarrelInclined_geo.cpp) together with its documentation
ENDRELEASENOTES


jmcarcell:
Looks good, some tests fail due to a missing xerces library but they are unrelated to this PR, merging so that people can use this. The cpp file has not beed added to any `CMakeLists.txt` but this is known by the authors (will be added later after more files are migrated)

PR #284, opened 2023-07-25
CMake: Install detector XML files in share/k4geo instead of additiona…

andresailer:
…l compact

BEGINRELEASENOTES
- CMake: Install detector XML files in share/k4geo instead of additional compact

ENDRELEASENOTES


andresailer:
And then let the K4GEO variable point to share/k4geo

andresailer:
We should really merge this...

PR #285, opened 2023-08-02
add CaloPrototype_v03: simple calo stack using MultiSegmentation

danieljeans:
BEGINRELEASENOTES
- Add a simple calorimeter stack, based on CaloPrototype_v02, which uses MultiSegmentation
   - e.g. can be used for a stack with alternating sensitive layer types
 
ENDRELEASENOTES

### DIFF
--- a/detector/CaloTB/CaloPrototype_v03.cpp
+++ b/detector/CaloTB/CaloPrototype_v03.cpp
@@ -1,0 +1,198 @@
+//====================================================================
+//  DD4hep Geometry driver for Sampling Calo BOX prototype
+//--------------------------------------------------------------------
+//  S.Lu, DESY
+//  $Id:  $
+//
+// Updated to use MultiSegmentation by D.Jeans (KEK) 08/2023
+//====================================================================
+#include "DD4hep/Printout.h"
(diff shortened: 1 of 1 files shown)


detector/CaloTB/CaloPrototype_v03.cpp
```diff
@@ -0,0 +1,198 @@
...
+            }
+            
+            // Set region, limitset, and vis.
+            layer_vol.setAttributes(theDetector, x_layer.regionStr(), x_layer.limitsStr(), x_layer.visStr());
+           
+            // Layer position in Z within the stave.
+            layer_pos_z += layer_thickness / 2;
+            // Layer physical volume.
+            PlacedVolume layer_phv = envelope.placeVolume(layer_vol, Position(0, 0, layer_pos_z));
+            layer_phv.addPhysVolID(identifierLayer, layer_num);
```
andresailer, line 182: Is this why the driver depends on the segmentation type? Would it be easier to just give that as a string parameter, instead of depending on the segmentation type?

andresailer: Thanks!

andresailer (approved):

PR #288, opened 2023-08-29
new ILD models: ILD_l5_v10 and ILD_l5_v11

danieljeans:
BEGINRELEASENOTES

- Update to SServices00 to allow individual components to be included/excluded
-- default behaviour unchanged

- add new model ILD_l5_v10 : ILD model for ILC
-- inner silicon (VTX, SIT, FTD) replaced by FCCee_o2_v02-inspired design

- add new model ILD_l5_v11 : ILD model for FCCee
-- inner silicon replaced by FCCee_o2_v02-inspired design
-- beampipe, MDI from FCCee_o2_v02

ENDRELEASENOTES


danieljeans:
sorry the SServices00 commit is a mess because of new if() statements enveloping much of the code, so everything is indented...
Is there a better way to do it?

jmcarcell:
You could not make indentation changes to keep the diff clean and then fix everything with `clang-format` at the end in the files you are changing

danieljeans:
@jmcarcell thanks for the hint. I recommitted without indenting to make the changes easy to see.
Now I'll investigate how to use clang-format : any hints welcome!

danieljeans:
I think this is now ready for review/merge

danieljeans:
Seems OK now.

detector/other/SServices00.cpp
```diff
@@ -76,237 +76,254 @@ static Ref_t create_element(Detector& theDetector, xml_h element, Ref_t)  {
 //====================================================================
   cout << "\nBuilding SServices00"<< endl;
 
-  //==================================================
-  //           BuildTPCEndplateServices
-  //==================================================
-  
-  // start to prepare the Material and geometry as Mokka
+  bool build_tpc_cooling  = dd4hep::getAttrOrDefault(x_det, _Unicode(build_tpc_cooling),   bool(true) );
```
tmadlener, line 79: This seems to be unused below. Is a condition missing, or is that intentional?

ILD/compact/README.md
```diff
@@ -40,6 +40,13 @@ The following ILD detector models are available in lcgeo ( current production mo
 
 ## Details
 
+### ILD_l5_v10
+same detector geometry as ILD_l5_v02 model, but with FCCee MDI and CLD-inspired silicon inner tracker & VTX in place of original ILD's VTX, SIT, FTD. 
+BeamCal, LHCAL removed.
+
+### ILD_l5_v10
```
tmadlener, line 47: One of these two should be `v11` right?

ILD/compact/ILD_l5_v11/TrackerDiskModuleIn.xml
```diff
@@ -0,0 +1,27 @@
+<lccdd>
```
tmadlener, line 1: This seems to be a copy(?) of `../ILD_common_v2/TrackerDiskModuleIn.xml`. Could we use that instead of creating a copy here?

ILD/compact/ILD_l5_v11/TrackerDiskModuleOut.xml
```diff
@@ -0,0 +1,27 @@
+<lccdd>
```
tmadlener, line 1: This seems to be a copy(?) of `../ILD_common_v02/TrackerDiskModuleOut.xml`. Could that be used instead of a dedicated copy here?

tmadlener (commented):
There seem to be a few copied modules that are present also in `ILD/compact/ILD_common_v02` (at least from a quick look, I might have missed some details that make a copy necessary):
- `TrackerDiskModuleOut.xml
- `TrackerDiskModuleIn.xml

these are there for both `v10` and `v11`, and I would prefer using the common definitions if possible, because it is less code to maintain and it is easier to spot differences between the models. We can always make a copy once it becomes necessary.

Additionally, the `InnerTrackerBarrelModule{Up,Down}` seem to be dupliated between `v10`, and `v11`. Can we also put them into the `ILD_common_v02` folder and reference them from there?

detector/other/SServices00.cpp
danieljeans: you missed it: it's used in line 87 to control whether to build the TPC-related services.

ILD/compact/README.md
danieljeans: yes, thanks for picking that up.

ILD/compact/ILD_l5_v11/TrackerDiskModuleOut.xml
danieljeans: fixed

danieljeans:
@tmadlener thanks for the useful comments. I've tidied up the code along the lines you suggested.

detector/other/SServices00.cpp
tmadlener: Ah yes. It was hiding very well ;) Sorry for the noise.

tmadlener (approved):
Thanks. Looks good to me now.

PR #290, opened 2023-09-17
Change `${LCIO_LIBRARIES}` to `LCIO::lcio`

jmcarcell:
BEGINRELEASENOTES
- Change ${LCIO_LIBRARIES} to `LCIO::lcio` in CMakeLists.txt for cases when `LCIO_LIBRARIES` is empty. Also prevents linking to `SIO` when it's not necessary but it's included in `LCIO_LIBRARIES`

ENDRELEASENOTES

This is the same as https://github.com/iLCSoft/LCCD/pull/7


tmadlener:
I am not entirely sure how `$LCIO_LIBRARIES` would be empty, and if it is whether we should not fix it upstream in LCIO instead of introducing workarounds in several downstream repositories.

jmcarcell:
PR updated, now works with https://github.com/iLCSoft/LCIO/pull/177
> I am not entirely sure how $LCIO_LIBRARIES would be empty, and if it is whether we should not fix it upstream in LCIO instead of introducing workarounds in several downstream repositories.
If you build together (at the same time) LCIO with another package then `LCIO_LIBRARIES` won't be set; one could set `LCIO_LIBRARIES` manually for this case to `lcio` and it will work. In this case where it's a single library that doesn't change over time we can use the target and have all build workflows working.

andresailer:
Superseded by #298

PR #289, opened 2023-09-06
CLD with ARC

atolosadelgado:
BEGINRELEASENOTES

* CLD_o3_v05 is created. This option evolves from option 2 version 05. The main characteristics with respect the mother version are:
	- ARC subdetectors included between the trackers and the ECal. 
	- Trackers are shrunk by 20 cm in radial and Z direction to accommodate the ARC. The space between layers is reduced proportionally, and the angular coverage is preserved.

ENDRELEASENOTES

Hi,

components from CLD (o2, v05) were taken to build the option 3, which contains the ARC detector, between the tracker and the ECAL. Modifications to the size of the outer tracker barrel, outer tracker endcap and inner tracker endcap were detailed in [this presentation](https://indico.cern.ch/event/1318491/contributions/5548508/attachments/2703694/4702470/ARC_IntegrationCLD_230823.pdf).

Overlap check with 10k points passed (it took 4h). 

I am doing now some efficiency plots, but I think it was a good moment to open the PR.

PR #295, opened 2023-11-02
Arc updates

atolosadelgado:
BEGINRELEASENOTES
- ARC updates. Optical properties of the radiators were updated. Aerogel layer is now placed parallel to the vessel walls. Vessel bulk material is now a realistic plastic-based foam. 

ENDRELEASENOTES

### DIFF
--- a/FCCee/CLD/compact/CLD_o3_v01/ARC_o1_v01.xml
+++ b/FCCee/CLD/compact/CLD_o3_v01/ARC_o1_v01.xml
@@ -65,7 +65,7 @@
   <vessel
       skinMaterial="CarbonFibStr"
       skin_vis="arc_vessel_vis"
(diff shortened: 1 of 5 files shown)


atolosadelgado:
The changes focus on ARC implementation. 
I have to modify the overlap macro file to have a tolerance of 1 fm, otherwise I found overlaps of 0 fm, caused by the lack of numerical precision. 
I have extended the time out of the overlap check of ARC. if the time is now not acceptable, I can remove the test.

atolosadelgado:
it seems ROOT system of units lead to an overlap, while using Geant4 there was none.

atolosadelgado:
I will move the changes of this PR into [PR289](https://github.com/key4hep/k4geo/pull/289)

PR #298, opened 2023-11-09
CMAKE: lcgeo to k4geo rename, k4geoConfig.cmake

andresailer:
BEGINRELEASENOTES
- CMake: The lcgeo project is renamed to k4geo, this also changes library names
- CMake: add the creation of a k4geoConfig.cmake file

ENDRELEASENOTES

Would be good to have https://github.com/iLCSoft/LCIO/pull/177/files for the LCIO::LCIO target


tmadlener:
Am I missing something and the non-nightlies based CI workflow should have failed with these changes? That one picks up LCIO v02-19-01 and that should not have the target, right?

andresailer:
I was puzzled by the result as well, until I remembered that DD4hep shims the LCIO target in if it doesn't exist

https://github.com/AIDASoft/DD4hep/blob/ff4c726c90084b378cd9dc413ae2114e829002a7/cmake/DD4hepBuild.cmake#L903-L911

tmadlener (approved):

PR #301, opened 2023-11-10
Make tests usable without install

Zehvogel:
BEGINRELEASENOTES
- Make tests usable without install

ENDRELEASENOTES

and hopefully do not break the ci by doing so?


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.20 FATAL_ERROR)
```
tmadlener, line 1: Is there a reason to bump the CMake version?

tmadlener (commented):
I am all for this. But this is not the only repo where we require installation for tests to run, so maybe we should discuss this once before and hopefully end up with a consistent situation in the end.

CMakeLists.txt
Zehvogel: required to be able to set the permissions with `configure_file`

PR #310, opened 2023-12-11
Change LCIO::LCIO to LCIO::lcio

jmcarcell:
BEGINRELEASENOTES
- Change LCIO::LCIO to LCIO::lcio. The target provided by LCIO is LCIO::lcio; DD4hep provides LCIO::LCIO so building without DD4hep and using LCIO::LCIO doesn't seem to work.

ENDRELEASENOTES

either this or we change the target in LCIO to be LCIO::LCIO; it seems in some repos like edm4hep it's all lowercase, in some others it's capitalized. In this case if anyone is using LCIO::LCIO they are using it through DD4hep and probably it's better to use the one from LCIO.

Using `LCIO::LCIO` was introduced in https://github.com/key4hep/k4geo/pull/298

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,8 +85,8 @@ add_library(lcgeo ALIAS k4geo)
 target_include_directories(${PackageName}   PRIVATE ${PROJECT_SOURCE_DIR}/detector/include )
 target_include_directories(${PackageName}G4 PRIVATE ${PROJECT_SOURCE_DIR}/detector/include )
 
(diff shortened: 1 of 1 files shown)


tmadlener (approved):

tmadlener:
Not sure how the lcg based releases work with this change, since I don't think they have the target yet.

jmcarcell:
From the logs I see some paths like: `/cvmfs/clicdp.cern.ch/iLCSoft/lcg/104/nightly/x86_64-centos7-gcc11-opt/lcio/HEAD/lib64` so it seems it does have it

tmadlener:
Ah right, those are clicdp nightlies and they come with the latest version of LCIO. Then it all makes sense. Thanks for checking.

PR #314, opened 2023-12-18
Updating IDEA vertex description and adding IDEA Silicon Wrapper

armin-ilg:
BEGINRELEASENOTES
- Fixing IDEA vertex issues: Use proper envelope, correcting positioning of disk staves in r and z, fixing overlaps, fixing surfaces so that DDPlanarDigitizer can be used with option to force hits onto surface
- Updating IDEA vertex to end of 2023 status (use different design for third inner barrel layer)
- First version of IDEA silicon wrapper using same constructor as IDEA vertex: Use tiles of 6, 12 or 24 modules with the size of ATLASPix3 but without insensitive region. The tiles are double-sided and two layers of such tiles are used. Most of the region should be covered by the Si Wrapper with at least one hit, most of the time >= two hits.
ENDRELEASENOTES

Hi @gganis @BrieucF @andresailer  (feel free to add others)
I made updates to the IDEA vertex (describing the [detector geometry presented in the MDI workshop in November 2023 by Fabrizio](https://agenda.infn.it/event/37720/contributions/212642/)).
Futhermore, I added a very first version of the Silicon Wrapper for IDEA, that I will present in January in Annecy (and maybe earlier at a detector concept or fullsim meeting).
Both detectors can be digitised with DDPlanarDigiProcessor or with [k4RecTracker](https://github.com/BrieucF/k4RecTracker/tree/master/VTXdigi).
Furthermore, the two subdetectors use the global detector dimensions from DectDimensions_IDEA_o1_v01.xml as far as possible. One thing I changed was to extend the Silicon Wrapper disk coverage in r (SiWrD_outer_radius) from 2020 mm to 2040 mm to minimise the gap in coverage between the disks and the Si Wrapper barrel. I think this is something that has to be looked at further in the future, once we make plots showing the hermeticity of the detector. Other than this, the dimensions are compatible with https://fcc-ee-detector-full-sim.docs.cern.ch/IDEA/, which I will update as well.

Let me know what you think!
Cheers,
Armin


armin-ilg:
I have no idea why the CMake test fails, it runs fine on lxplus for me. Is there a limit in the CI memory consuption maybe? IDEA now uses up to 2.2 GB of RAM, which means it just runs on lxplus.

armin-ilg:
Opening the MR for review, although I still try to fix the check

BrieucF:
Hi Armin, thanks. 

The test fail for me locally as well, I attach the log [here](https://github.com/key4hep/k4geo/files/13713295/log.txt). It seems to me that it is related to the xml parsing.

BrieucF:
Ok, now it seems that we have to increase the time out: https://github.com/key4hep/k4geo/blob/main/lcgeoTests/CMakeLists.txt#L87C108-L87C119

armin-ilg:
Ok, the test is working now :-)

atolosadelgado:
Hi @armin-ilg 

I have checked few things of this PR using the stable release of key4hep:
* pure navigation seems ok (0.02s/100 geantinos) 
* the hit pattern of geantinos in the detector seems ok, I do not see holes in the geometry, but a dedicated study would be needed to determine if there are cracks
* geometry takes 300MB when translated into Geant4, which is good
* Only 3 overlaps are reported, all of them related with a shape called `Body` (Tessellated solid) and another one called `Vertex` (subtraction solid). Does the `Body` shape correspond to the support structure from the CAD? Could you explain what is the shape `Vertex_0:0 (G4SubtractionSolid)`? 

I leave here the overlap report, do you see the same results? (the overlap points may differ because it is an statistical process)
```
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume Vertex_0:0 (G4SubtractionSolid) with AV_6374!VTXIBSupport_1#1!Body1_0#0:2 (G4TessellatedSolid)
          overlap at local point (-1.02369,-18.8181,182.322) by 2.27784 mm  (max of 9 cases)
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !

--
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume AV_6374!VTXIBSupport_1#1!Body1_0#0:2 (G4TessellatedSolid) with Vertex_0:0 (G4SubtractionSolid)
          overlap at local point (-29.0238,5.63001,176.553) by 1.12553 cm  (max of 9056 cases)
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !

--
*** G4Exception : GeomVol1002
      issued by : G4PVPlacement::CheckOverlaps()
Overlap with volume already placed !
          Overlap is detected for volume AV_6374!VTXIBSupport_1#1!Body1:1_1#1:3 (G4TessellatedSolid) with Vertex_0:0 (G4SubtractionSolid)
          overlap at local point (-28.7859,-7.89524,-179.351) by 1.12504 cm  (max of 9076 cases)
NOTE: Reached maximum fixed number -1- of overlaps reports for this volume !
```

PR #315, opened 2024-01-09
Updates for ALLEGRO_o1_v02

BrieucF:
BEGINRELEASENOTES
- [ALLEGRO_o1_v02] Replace the simplified drift chamber with the detailed one
- [TESTS] Add a test for ALLEGRO_o1_v02
- [ALLEGRO_o1_v02] File and variable renaming plus propagation of overlap fix to upstream and calibration xml's

ENDRELEASENOTES


BrieucF:
Can we merge this? The stable stack test fails but it seems to be due to https://github.com/key4hep/k4geo/pull/310

PR #317, opened 2024-02-06
ALLEGRO: EcalBarrel CaloData: caloDim.dZ is already the half length

andresailer:
BEGINRELEASENOTES
- ALLEGRO: EcalBarrel: fix the length of the Z extent for the calorimeter data extension for reconstruction

ENDRELEASENOTES


SwathiSasikumar (approved):
Thanks for pointing this out.

PR #318, opened 2024-02-14
Update some ALLEGRO xml config files

giovannimarchiori:
BEGINRELEASENOTES
* ALLEGRO: Update some xml config files

ENDRELEASENOTES

### DIFF
--- a/FCCee/ALLEGRO/compact/ALLEGRO_o1_v02/ECalBarrel_thetamodulemerged_upstream.xml
+++ b/FCCee/ALLEGRO/compact/ALLEGRO_o1_v02/ECalBarrel_thetamodulemerged_upstream.xml
@@ -25,7 +25,7 @@
     <constant name="CryoBarrelFrontCold" value="3.8*mm"/>  <!-- Al solid equivalent of 0.043 X0 sandwich CFRP -->
     <constant name="CryoBarrelFront" value="CryoBarrelFrontWarm+CryoBarrelFrontCold"/>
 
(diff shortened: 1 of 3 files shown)


BrieucF:
The tests (also with the nightlies) are failing but it has nothing to do with this PR, will open an issue

FCCee/ALLEGRO/compact/ALLEGRO_o1_v02/ECalBarrel_thetamodulemerged_upstream.xml
```diff
@@ -25,7 +25,7 @@
     <constant name="CryoBarrelFrontCold" value="3.8*mm"/>  <!-- Al solid equivalent of 0.043 X0 sandwich CFRP -->
     <constant name="CryoBarrelFront" value="CryoBarrelFrontWarm+CryoBarrelFrontCold"/>
 
-    <constant name="CryoBarrelBackCold" value="30*mm"/>   <!-- Al solid corresponding to 0.34 X0 -->
+    <constant name="CryoBarrelBackCold" value="1100*mm"/>   <!-- Al solid corresponding to 1.25 X0 -->
```
andresailer, line 28: I would expect the X0 to scale similarly to the increase in thickness, so from 30 to 1100 would be from 0.34 to 12.47. I guess one of the four numbers 0.34, 1.25, 30 or 1100 is wrong then?

giovannimarchiori: yep there was a factor 10 missing in 1.25 in the comment, should be 12.5 - just fixed

PR #327, opened 2024-03-01
[IDEA] Add a solenoid and magnetic fields

BrieucF:
BEGINRELEASENOTES
- Add a solenoid and magnetic fields for IDEA. The solenoid has the right material budget (0.75 X0) and spacial extent but its internals should be revised by ultra-thin solenoid designers.
- Add the endplate lead absorbers for IDEA

ENDRELEASENOTES


jmcarcell (approved):

andresailer (commented):
Nitpick / potential breakage of XML because of stray `-->` comment closer

PR #333, opened 2024-04-22
Fix event display for tracker endcap in CEDViewer

gaede:
BEGINRELEASENOTES
- fix CED event display for CLIC like detectors using TrackerEndcap_o2_v0x_geo 
       -  fix nPetals in ZDiskPetalsData (for CEDViewer) to use nmodules (e.g. 48 ) rather than nrings
       -  store the number of rings in  `ZDiskPetalsData::sensorsPerPetal`     

ENDRELEASENOTES

### DIFF
--- a/detector/tracker/SiTrackerEndcap_o2_v01ext_geo.cpp
+++ b/detector/tracker/SiTrackerEndcap_o2_v01ext_geo.cpp
@@ -136,8 +136,9 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
       sumZ += zstart;
       r = r + mod_shape->GetDY();
 
+      petal_num = nmodules;  // store the module number of this ring for nPetals in ZDiskPetalsData
+
       for(int k=0; k<nmodules; ++k) {
-        petal_num = nmodules;  // store the module number of this ring for nPetals in ZDiskPetalsData
         string m_base = _toString(l_id,"layer%d") + _toString(mod_num,"_module%d") + _toString(k,"_sensor%d");
         double x = -r*std::cos(phi);
         double y = -r*std::sin(phi);
--- a/detector/tracker/SiTrackerEndcap_o2_v01ext_geo.cpp
+++ b/detector/tracker/SiTrackerEndcap_o2_v01ext_geo.cpp
@@ -111,6 +111,7 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
     int l_id    = x_layer.id();
     int mod_num = 0;
     int ring_num = 0;
+    int petal_num = 0;
 
     double sumZ(0.), innerR(1e100), outerR(0.);
 
@@ -136,6 +137,7 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
       r = r + mod_shape->GetDY();
 
       for(int k=0; k<nmodules; ++k) {
+        petal_num = nmodules;  // store the module number of this ring for nPetals in ZDiskPetalsData
         string m_base = _toString(l_id,"layer%d") + _toString(mod_num,"_module%d") + _toString(k,"_sensor%d");
         double x = -r*std::cos(phi);
         double y = -r*std::sin(phi);
@@ -173,7 +175,8 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
     thisLayer.zPosition = sumZ/ring_num; // average z
     thisLayer.distanceSensitive = innerR;
     thisLayer.lengthSensitive = outerR - innerR;
-    thisLayer.petalNumber = ring_num; // number of rings in petalNumber, needed for tracking
+    thisLayer.petalNumber = petal_num;
+    thisLayer.sensorsPerPetal = ring_num; // number of rings in petalNumber, needed for tracking
     zDiskPetalsData->layers.push_back(thisLayer);
 
   }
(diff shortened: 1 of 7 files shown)


tmadlener:
As we have figured out ConformalTracking will most likely not be affected by these changes, but there are other geometries where a similar issue is probably present:

- `SiTrackerEndcap_o2_v02ext_geo.cpp:    thisLayer.petalNumber=ring_num; // number of rings in petalNumber, needed for tracking`
- `SiTrackerEndcap_o2_v01ext_geo.cpp:    thisLayer.petalNumber = ring_num; // number of rings in petalNumber, needed for tracking`
- `rackerEndcap_o2_v05_geo.cpp:            thisLayer.petalNumber   = ring_no ; //Store the number of rings in petalNumber needed for tracking`
- `TrackerEndcap_o1_v05_geo.cpp:            thisLayer.petalNumber   = ring_no ; //Store the number of rings in petalNumber needed for tracking`
- `TrackerEndcap_o1_v04_geo.cpp:            thisLayer.petalNumber   = ring_no ; //Store the number of rings in petalNumber needed for tracking`
- `TrackerEndcap_o2_v04_geo.cpp:            thisLayer.petalNumber   = ring_no ; //Store the number of rings in petalNumber needed for tracking`
- `TrackerEndcap_o2_v06_geo.cpp:            thisLayer.petalNumber   = ring_no ; //Store the number of rings in petalNumber needed for tracking`

tmadlener:
@andresailer any insights on where we should look for usages of this? We didn't find anything in Conformaltracking, but this has the potential to break in a few places and some of these models might not be in too much use.

tmadlener:
This is the list of hits I get with `grep -r petalNumber` on a (pretty much) complete checkout of all of iLCSoft (sans the ones from the comment above)

```
MarlinTrkProcessors/source/Digitisers/src/DDSpacePointBuilder.cc:    //int Nsensors = ft->layers.at(layer).petalNumber ; 
MarlinUtil/source/src/DDMarlinCED.cc:  int petalNumber = thisLayer->petalNumber;
MarlinUtil/source/src/DDMarlinCED.cc:  returnParams.Rmax = returnParams.Rmax/cos(M_PI/petalNumber);
MarlinUtil/source/src/DDMarlinCED.cc:  returnParams.Rmin = returnParams.Rmin/cos(M_PI/petalNumber);
MarlinUtil/source/src/DDMarlinCED.cc:  returnParams.inner_symmetry = petalNumber;
MarlinUtil/source/src/DDMarlinCED.cc:  returnParams.outer_symmetry = petalNumber;
MarlinUtil/source/src/DDMarlinCED.cc:  returnParams.phi0 = phi0 + 270. - 180./petalNumber - (360./petalNumber)*(int ((270.-180./petalNumber)/(360./petalNumber)));
ILDPerformance/trackerHitCtr/src/TrackerHitCounter.cc:                int nModules = layer.petalNumber;
ForwardTracking/src/ForwardTracking/ForwardTracking.cc:     if( l.petalNumber > nModules ) nModules = l.petalNumber ;
ForwardTracking/src/FTDNoise/FTDBackgroundProcessor.cc:      unsigned nModules = ftd->layers[layer].petalNumber ;
ForwardTracking/src/Analysis/TrueTrackCritAnalyser.cc:     if( l.petalNumber > nModules ) nModules = l.petalNumber ;
```

This looks like the main things that could break are forward tracking and potentially ILDPerformance. The ones in `DDMarlinCED` are the ones that we try to fix with this here. The others would need to be checked.

In the end we would probably need to fix all the drivers in the same way, I think.

tmadlener:
I have ported the changes to the drivers mentioned in the comment above.

We (as ILD) think this should be transparent and fix some visualization issues.

andresailer:
Do you have a before and after picture?

Otherwise, I agree that we are not using those parameters in ConformalTracking, so it shouldn't have an impact.

tmadlener:
The following screenshots are done via 
```bash
ced2go -d $K4GEO/ILD/compact/ILD_l5_v11/ILD_l5_v11.xml <some-slcio-input-that-doesnot-matter>
```

Before fix:
Screenshot from 2024-06-03 16-08-51

After fix:
Screenshot from 2024-06-03 16-10-50

tmadlener:
@andresailer anything speaking against merging this?

andresailer (approved):

andresailer:
Sorry, I missed the picture update! Thanks!

detector/tracker/SiTrackerEndcap_o2_v01ext_geo.cpp
```diff
@@ -173,7 +175,8 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
     thisLayer.zPosition = sumZ/ring_num; // average z
     thisLayer.distanceSensitive = innerR;
     thisLayer.lengthSensitive = outerR - innerR;
-    thisLayer.petalNumber = ring_num; // number of rings in petalNumber, needed for tracking
+    thisLayer.petalNumber = petal_num;
```
jmcarcell, line 178: Why not set this to `nmodules` that is fixed? `petal_num` is being set to `nmodules` `nmodules` times

tmadlener: `nmodules` is no longer in scope for using it when `thisLayer.petalNumber` is set. However, I can setting it `nmodules` times is unnecessary, and I can lift it out of the loop.

PR #334, opened 2024-05-06
Change the loop variable to size_t to remove the compilation warning

zwu0922:
BEGINRELEASENOTES

- xtalk_neighbors_moduleThetaMergedSegmentation.cpp: change the loop variable itField from int to size_t, in order to remove a compilation warning.

ENDRELEASENOTES


jmcarcell:
This is typically done by declaring the loop variable, `itField` in this case, to be `size_t` instead of `int`; it represents better that `itField` is going to represent a size or count and you don't have to do any casting. Can you change it to `size_t` and update the release notes above?

zwu0922:
> This is typically done by declaring the loop variable, `itField` in this case, to be `size_t` instead of `int`; it represents better that `itField` is going to represent a size or count and you don't have to do any casting. Can you change it to `size_t` and update the release notes above?

The type of itField has been changed to size_t. The title and comment of the PR are modified accordingly.

jmcarcell:
> Hi, and with this you do not need to cast `itField` to int inside the loop?

No, this is done for you. But with comparisons, signed and unsigned values can be trickier, for example  `cout << (1U < -2) << endl;` prints `1`. So that's why the warning appears.

jmcarcell (approved):

PR #335, opened 2024-05-17
[IDEA] Fix solenoid for o1_v03

BrieucF:
BEGINRELEASENOTES
- Fix for the IDEA_o1_v03 solenoid position

ENDRELEASENOTES


BrieucF:
@andresailer can we merge this?

andresailer:
This is changing XML parameters for a detector that I have no expertise in nor responsibility for. So unless you have some specific concern I should comment on, don't wait for me in these cases.

PR #336, opened 2024-05-17
PreShower updates

Ns921:
PreShower updates


jmcarcell:
Hi @Ns921, can you expand the text in your pull request so that people can know what this is about? It should say something about what is changing (in this case added) and what the purpose is. Con you also fill the release notes like the other pull requests do? That is the text between BEGINRELEASENOTES and ENDRELEASENOTES. Then when we make a release all these notes are collected like in https://github.com/key4hep/k4geo/releases/tag/v00-20-00

PR #339, opened 2024-06-06
ALLEGRO_o1_v03: VTX and drift chamber from IDEA, add magnetic fields

BrieucF:
BEGINRELEASENOTES
- [FCCee-ALLEGRO_o1_v03] Vertex detector and drift chamber updated to the last IDEA version
- [FCCee-ALLEGRO_o1_v03] Added solenoidal and MDI magnetic fields
- [FCCee-ALLEGRO_o1_v03]  Removed ALLEGRO_o1_v03_ecalonly.xml and ALLEGRO_o1_v03_trackeronly.xml to ease maintenance (they can be obtained by commenting out sub-detectors)

ENDRELEASENOTES


BrieucF:
The stable release is failing due to a missing driver (DriftChamber_o1_v02).

FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/ALLEGRO_o1_v03.xml
```diff
@@ -33,14 +33,79 @@
...
   <include ref="MuonTagger.xml"/>
 
+  <fields>
+    <!-- FIXME this is a place holder to get a reasonable magnetic field, it should be a real MAP obtained from FEM tools taking into account HCAL return yoke -->
+    <!-- The z extent should also accomodate space for the feedthroughs, need input from R&D and decisions on cold/warm electronics -->
+    <field name="GlobalSolenoid" type="solenoid"
+           inner_field="SolenoidField"
+           outer_field="-1.0*tesla"
+           zmax="EMBarrel_dz"
+           outer_radius="BarCryoECal_rmax-CryoBarrelBackWarm-CryoBarrelBackCold">
```
giovannimarchiori, line 52: should this also include a -NLiqBathThicknessBack ?
In any case, this is so small, and in both cases the outer_radius will be greater than that of the LAr volume, so it should have no practical consequences.

giovannimarchiori:
Hi @BrieucF , this looks very good to me. I only have a tiny doubt but I think that in practice it would have essentially no impact on the simulation

FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/ALLEGRO_o1_v03.xml
BrieucF: I would'nt say so, because `NLiqBathThicknessBack` is for services right after the sensitive calo, the solenoid is after that. Of course we are missing a detailed mechanical design here (and the software implementation of the cryostat + solenoid is just accounting for material budget, not for precise positions of the components) but indeed the impact for what we do at this point should be invisible. 

Screenshot 2024-06-06 at 4 00 05 PM

giovannimarchiori: ah, ok, thanks! I wasn't sure whether the NLiq was where you pointed at, or was the liquid nitrogen inside the cryostat, between the warm and cold parts.

giovannimarchiori (approved):
Looks good to merge to me!

PR #343, opened 2024-06-24
Make TPC  have detector ID 4 and use consistent tracker cell ID

tmadlener:
BEGINRELEASENOTES
- Make the TPC have detector ID 4 and use a consistent cellID for all tracking detectors in `ILD_l5_v11` in order to make tracking code run again

ENDRELEASENOTES

@Victor-Schwan @danieljeans these are some minor fixes that are necessary. Clupatra really wants the TPC to have detector ID 4 and some other tracking code also requires a consistent cellID encoding scheme throughout the detector. We switched to the ILD one here for the CLD bits as well.

### DIFF
--- a/ILD/compact/ILD_l5_v11/FCCdefs.xml
+++ b/ILD/compact/ILD_l5_v11/FCCdefs.xml
@@ -104,6 +104,6 @@
 
   <constant name="InnerTracker_outer_radius" value="top_TPC_inner_radius-5*mm" />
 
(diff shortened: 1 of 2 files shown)


tmadlener:
@andresailer, @danieljeans anything obvious that we could be missing here? Otherwise I would merge this.

danieljeans:
@tmadlener I can't think of any potential problem. I think I changed the TPC detectorID just to keep the CLD tracker det IDs unchanged, but I don't think that is relied on anywhere in the downstream chain.

tmadlener:
Thanks @danieljeans, we weren't entirely sure, but from running tracking we also haven't seen any obvious issues yet.

PR #345, opened 2024-06-27
Add support for ECal endcap turbine

varnes:
BEGINRELEASENOTES

Initial pass at code to support "turbine" style endcap ECal geometry

ENDRELEASENOTES


giovannimarchiori:
Hi @varnes ,

good to see that we start to have a G4 implementation of the ECAL endcap for ALLEGRO!

Some questions:
- why there are so many file changes that do not relate to your PR, did you rebase to main before creating it?
- why the endcap model is not included also in our current baseline ALLEGRO detector model which is o1_v03?
- what is the difference between ECalEndcaps_Turbine.xml and FCCee_ECalEndcaps_Turbine.xml, are they both needed? And the difference between ECalEndcaps_Turbine_Calib.xml and ECalEndcaps_Turbine_calibration.xml?
- Since the geometry is not straightforward, can you add some description about it in the README with possibly some pointers to documents/presentations with more details? Same suggestion also for the related segmentation class

Thanks,  Giovanni

varnes:
Hi Giovanni,

  Thanks for the comments!  Some responses:

- indeed I did not rebase before creating this PR.  I will mark this as draft, rebase, and re-open it when ready
- o1_v03 didn't exist when I created my fork.  I will add my code there after the rebase
- these are indeed duplicates, and only one of the two files is needed in both cases.  I will remove the duplicates before reopening the PR
- yes, I will add comments and pointers to documentation

Best regards,
Erich

PR #351, opened 2024-07-05
Include common review comments in the PR template

BrieucF:
I open this PR to discuss the content of `.github/PULL_REQUEST_TEMPLATE`

BEGINRELEASENOTES
- Include common review comments in the PR template

ENDRELEASENOTES

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -5,7 +5,7 @@ To ease the review process, please consider the following before to open the pul
 - [ ] the code is covered by tests. See https://github.com/key4hep/k4geo/blob/main/test/CMakeLists.txt
 - [ ] the PR source branch has been rebased (`--ff-only`) to `k4geo/main`
 - [ ] the PR does not contain any additions or modifications that do not belong to it
-- [ ] the changes in this PR have not introduced any overlaps. You can do so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
+- [ ] the changes in this PR have not introduced any overlaps. You can check so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
 - [ ] The release notes below contain a succinct and comprehensive description of the changes that are sufficiently self-explanatory
 
 BEGINRELEASENOTES
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,14 +1,12 @@
-<!--
 To ease the review process, please consider the following before to open the pull request:
-- is your code well commented?
-- did you update the relevant README(s)? See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md
-- if you added some code, is it covered by tests? See https://github.com/key4hep/k4geo/blob/main/lcgeoTests/CMakeLists.txt
-- is your PR branch up to date with origin 'main'?
-- are you not adding/modifying files that should not have been added/modified (or at least, not in this PR)?
-- if you added/modified a detector geometry, have you checked that was no overlap? You can do so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
-
-If your modifications are compliant with the above, populate the release note text below, and open the PR!
--->
+- [ ] the code is sufficiently well documented
+- [ ] the xml file free parameters that can not be modified without additional prescriptions are well indicated
+- [ ] the relevant README(s) were updated. See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md
+- [ ] the code is covered by tests. See https://github.com/key4hep/k4geo/blob/main/test/CMakeLists.txt
+- [ ] the PR source branch has been rebased (`--ff-only`) to `k4geo/main`
+- [ ] the PR does not contain any additions or modifications that do not belong to it
+- [ ] the changes in this PR have not introduced any overlaps. You can do so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
+- [ ] The release notes below contain a succinct and comprehensive description of the changes that are sufficiently self-explanatory
 
 BEGINRELEASENOTES
 - Thank you for writing the text to appear in the release notes. It will show up
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -5,6 +5,7 @@ To ease the review process, please consider the following before to open the pul
 - if you added some code, is it covered by tests? See https://github.com/key4hep/k4geo/blob/main/lcgeoTests/CMakeLists.txt
 - is your PR branch up to date with origin 'main'?
 - are you not adding/modifying files that should not have been added/modified (or at least, not in this PR)?
+- if you added/modified a detector geometry, have you checked that was no overlap? You can do so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
 
 If your modifications are compliant with the above, populate the release note text below, and open the PR!
 -->
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,16 +1,18 @@
-To ease the review process, please consider the following before to open the pull request:
-- [ ] the code is sufficiently well documented
-- [ ] the xml file free parameters that can not be modified without additional prescriptions are well indicated
+To ease the review process, please consider the following before opening a pull request:
+- [ ] the code is sufficiently well documented (e.g. inline comments)
 - [ ] the relevant README(s) were updated. See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md
 - [ ] the code is covered by tests. See https://github.com/key4hep/k4geo/blob/main/test/CMakeLists.txt
 - [ ] the PR source branch has been rebased (`--ff-only`) to `k4geo/main`
 - [ ] the PR does not contain any additions or modifications that do not belong to it
-- [ ] the changes in this PR have not introduced any overlaps. You can check so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
 - [ ] The release notes below contain a succinct and comprehensive description of the changes that are sufficiently self-explanatory
+
+If you are modifying detector dimensions or adding new xml parameters, also consider the following:
+- [ ] the xml file free parameters that can not be modified without additional prescriptions are well indicated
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,8 +1,15 @@
+To ease the review process, please consider the following before to open the pull request:
+- is your code well commented?
+- did you update the relevant README(s)? See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md
+- if you added some code, is is covered by tests? See https://github.com/key4hep/k4geo/blob/main/lcgeoTests/CMakeLists.txt
+- is your PR branch up to date with origin 'main'?
+- are you not adding/modifying files that should not have been added/modified (or at least, not in this PR)?
 
+If your modifications are compliant with the above: delete thoses lines from the PR description, populate the release note text below, and open the PR!
```
tmadlener, line 8: I think you can make this a comment block using `<!--` and `-->`. That way the text would still show up for people if they open a PR, but it would not clutter the PR once it has been submitted, because it won't be visible. This obviously depends a bit on whether you want to make it harder to ignore or not ;)

tmadlener: I am also not sure if PRs also have the fancy features that issues have (see, e.g. [DD4hep issues](https://github.com/AIDASoft/DD4hep/blob/master/.github/ISSUE_TEMPLATE/bug_report.yml))

jmcarcell:
Another question would be if we want the same templates for all of the Key4hep repositories or not; at least the questions in the template in this PR are quite generic and don't need to refer to the exact file in the current repository.

.github/PULL_REQUEST_TEMPLATE
BrieucF: Yes, I think we can make it a comment block (it is anyway easy to ignore :-P )

BrieucF:
> Another question would be if we want the same templates for all of the Key4hep repositories or not; at least the questions in the template in this PR are quite generic and don't need to refer to the exact file in the current repository.

That is a good point, what is written there is indeed very generic. The only thing is that I wanted to link the README(s) and the tests here because people might not know where they are... But ok, this is not a super strong argument, I let you guys decide

tmadlener:
I think (most of) the other Key4hep ones already pull theirs from the [central templates](https://github.com/key4hep/.github/tree/main/.github). `k4geo` is likely a special case due to its history and the fact that it was originally in the iLCSoft organisation.

I have no strong opinion either, but as Brieuc has already pointed out two things that are somewhat k4geo specific, I would lean slightly towards keeping the dedicated template here.

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,8 +1,19 @@
+<!--
+To ease the review process, please consider the following before to open the pull request:
+- is your code well commented?
+- did you update the relevant README(s)? See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md
+- if you added some code, is it covered by tests? See https://github.com/key4hep/k4geo/blob/main/lcgeoTests/CMakeLists.txt
+- is your PR branch up to date with origin 'main'?
+- are you not adding/modifying files that should not have been added/modified (or at least, not in this PR)?
+- if you added/modified a detector geometry, have you checked that was no overlap? You can do so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
```
andresailer, line 8:
```suggestion
- if you added/modified a detector geometry, have you checked that there is no overlap? You can do so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
```

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,8 +1,19 @@
+<!--
```
andresailer, line 1: Why the XML comment signs?

Change everything into checkboxes?

.github/PULL_REQUEST_TEMPLATE
andresailer, line 8:
```suggestion
To ease the review process, please consider the following before to open the pull request:
- [ ] My code is sufficiently well documented
- [ ] the relevant README(s) were updated. See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md
- [ ] The code is covered by tests? See https://github.com/key4hep/k4geo/blob/main/tests/CMakeLists.txt
- [ ] the PR source branch has been rebased to `k4geo/main`
- [ ] the PR does not contain any additions or modifications that do not belong to it
- [ ] the changes in this PR have not introduced any overlaps. You can do so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
```

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,8 +1,19 @@
...
+- is your PR branch up to date with origin 'main'?
+- are you not adding/modifying files that should not have been added/modified (or at least, not in this PR)?
+- if you added/modified a detector geometry, have you checked that was no overlap? You can do so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
 
+If your modifications are compliant with the above, populate the release note text below, and open the PR!
+-->
 
 BEGINRELEASENOTES
 - Thank you for writing the text to appear in the release notes. It will show up
   exactly as it appears between the two bold lines
```
andresailer, line 15:
```suggestion
  exactly as it appears between the two ALL CAPS lines
```

.github/PULL_REQUEST_TEMPLATE
andresailer: People are going to ignore it anyway, so if we make it checkboxes, we can see from the PR overview that we can also ignore the PR!

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,8 +1,19 @@
+<!--
+To ease the review process, please consider the following before to open the pull request:
+- is your code well commented?
+- did you update the relevant README(s)? See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md
+- if you added some code, is it covered by tests? See https://github.com/key4hep/k4geo/blob/main/lcgeoTests/CMakeLists.txt
+- is your PR branch up to date with origin 'main'?
+- are you not adding/modifying files that should not have been added/modified (or at least, not in this PR)?
+- if you added/modified a detector geometry, have you checked that was no overlap? You can do so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
 
+If your modifications are compliant with the above, populate the release note text below, and open the PR!
```
andresailer, line 10:
```suggestion
  - [ ] The release notes below contain a succinct and comprehensive description of the changes that are sufficiently self-explanatory
```

atolosadelgado:
There is a battery of tests I do for geometries, can we add these as check boxes, at least it would be useful for ourselves as Andre suggested :)

* overlap check (maybe increasing the number of points up to 100k or more). I think this one is already in the list
* simulation with geantinos, 300k or more. If error is print out during navigation, one has to go back to previous point. Verbosity can be increased to catch inconsistencies
* material scan. This one is more handmade, because one has to spot the error by oneself (that is, if you expect aluminum and the geantinos do not see aluminum, there is a problem)
* simulation with physical particles (particle gun), which can trigger errors during navigation (as it was the case of the twisted tube)

Even with default verbosity, the amount of text printed out is too much, so I usually redirect the output to a file and then look for some keywords such as `error` and `G4-Exception`.

The tests of the simulation with physical particles and geantinos are also useful to get an idea of the speed.

BrieucF:
Sorry for the long silence, here is a new version.

atolosadelgado:
Hi,

the transition from `k4simgeant4` to `DDSim` exposed the fact that many subsystems do not have a region and limits defined, which may lead to lack of accuracy. I would advice to attach a region to each subsystem, defining the production cut and optionally associating it with some limits.

Some information about the _Region_  [[link](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/TrackingAndPhysics/cutsPerRegion.html?highlight=region), [link](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/TrackingAndPhysics/thresholdVScut.html)] and the associated _user limits_ [[link](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/TrackingAndPhysics/userLimits.html?highlight=limit)] can be found in Geant4 documentation.

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,8 +1,17 @@
-
+To ease the review process, please consider the following before to open the pull request:
+- [ ] the code is sufficiently well documented
+- [ ] the xml file free parameters that can not be modified without additional prescriptions are well indicated
+- [ ] the relevant README(s) were updated. See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md
+- [ ] the code is covered by tests. See https://github.com/key4hep/k4geo/blob/main/test/CMakeLists.txt
+- [ ] the PR source branch has been rebased (`--ff-only`) to `k4geo/main`
+- [ ] the PR does not contain any additions or modifications that do not belong to it
+- [ ] the changes in this PR have not introduced any overlaps. You can check so with the following command: ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt
```
tmadlener, line 8:
```suggestion
- [ ] the changes in this PR have not introduced any overlaps. You can check so with the following command: `ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" >> overlapDump.txt`
```

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -1,8 +1,17 @@
-
+To ease the review process, please consider the following before to open the pull request:
```
tmadlener, line 1:
```suggestion
To ease the review process, please consider the following before opening a pull request:
```

tmadlener (commented):
Should we consider splitting this list into more than one? E.g. one that definitely almost always applies (e.g. test coverage, README updates, etc..) and one (or more) other ones that apply for entirely new developments or significant changes. 

How long does an overlap check take? Can we run that in CI automatically? Everything that doesn't have to be done manually is less work for everyone involved.

atolosadelgado:
> Should we consider splitting this list into more than one? E.g. one that definitely almost always applies (e.g. test coverage, README updates, etc..) and one (or more) other ones that apply for entirely new developments or significant changes.
> 
> How long does an overlap check take? Can we run that in CI automatically? Everything that doesn't have to be done manually is less work for everyone involved.

The overlap check can take several minutes already for 1 subsystem, but some tricks may be applied to speed it up. I added dedicated tests for ARC and the drift chamber [[link](https://github.com/key4hep/k4geo/blob/456aafcac6bb38732718ab753bbc5f6fd5ae0ad0/test/CMakeLists.txt#L164)], but they are heavily customized and standalone.

Recently, I noticed that simulation of Geantinos is also useful to spot problems, and it may be better suited for big detectors (10M geantinos should be fast and they would cover the 4pi with some granularity). Adding this check to the CI tests looks like a good idea to me, thanks for the question!

.github/PULL_REQUEST_TEMPLATE
andresailer:
```suggestion
- [ ] the changes in this PR have not introduced any overlaps. You can check so with the following command: `ddsim --compactFile PATH_TO_COMPACT_FILE --runType run --ui.commandsInitialize "/geometry/test/run" > overlapDump.txt`
```

Or why append to the file?

BrieucF:
Here a new version with comments implemented. I did not add a requirement on the detector region/limits, I feel like it is a bit too detailed, but feel free to propose a sentence if you think it is important and should be included here.

I propose to tackle the 'overlap check in CI' in another PR.

andresailer (approved):

PR #322, opened 2024-02-16
[IDEA] Adding the detailed version of the muon system.

mahmoudali2:
BEGINRELEASENOTES
- Defining the first draft of the detailed muon system, which depend on mosaics of 50 * 50 cm^2 mRWELL chambers.
- Defining a suitable XML for the new detailed version.
- Describing µRWELL materials.
- Adding the parameters of the muon system into the full IDEA implementation.
- ...

ENDRELEASENOTES

PR #353, opened 2024-07-09
Use + std::to_string to convert integers before appending to a string

jmcarcell:
BEGINRELEASENOTES
- muonSystemMuRWELL_o1_v01.cpp: Use + std::to_string to append to a string, instead of adding an integer to a string (introduced in https://github.com/key4hep/k4geo/pull/322). Adding a string and an integer cuts the string by as many characters as the value of the integer.

ENDRELEASENOTES

See https://gcc.godbolt.org/z/qfojcnMK9 for an example. Should we worry that this has not been caught when testing? @mahmoudali2 @BrieucF

### DIFF
--- a/detector/muonSystem/muonSystemMuRWELL_o1_v01.cpp
+++ b/detector/muonSystem/muonSystemMuRWELL_o1_v01.cpp
@@ -524,7 +524,7 @@ static dd4hep::Ref_t createmuonSystemMuRWELL_o1_v01(dd4hep::Detector& lcdd,
   dd4hep::Position detectorLayerTrans(0., 0., 0.);
   dd4hep::PlacedVolume detectorLayerPhys = BarrelVolume.placeVolume(BarrelDetectorLayerVolume, dd4hep::Transform3D(dd4hep::RotationZ(0.), detectorLayerTrans));
   detectorLayerPhys.addPhysVolID("layer", numBarrelLayer+1); 
-  dd4hep::DetElement BarrelDetectorLayerDE(BarrelDE, "MSBarrelDetectorLayerDE" + numBarrelLayer+1, numBarrelLayer+1);
+  dd4hep::DetElement BarrelDetectorLayerDE(BarrelDE, "MSBarrelDetectorLayerDE" + std::to_string(numBarrelLayer+1), numBarrelLayer+1);
   BarrelDetectorLayerDE.setPlacement(detectorLayerPhys);
   BarrelDetectorLayerVolume.setVisAttributes(lcdd.visAttributes("no_vis")); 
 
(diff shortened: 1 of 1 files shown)


detector/muonSystem/muonSystemMuRWELL_o1_v01.cpp
```diff
@@ -616,7 +616,7 @@ static dd4hep::Ref_t createmuonSystemMuRWELL_o1_v01(dd4hep::Detector& lcdd,
     endcapTrans = dd4hep::Position(0., 0., endcapType * endcapOffset);
     endcapPhys = detectorVolume.placeVolume(endcapVolume, dd4hep::Transform3D(dd4hep::RotationZ(0.), endcapTrans));
     endcapPhys.addPhysVolID("type", endcapType);
-    EndcapDE = dd4hep::DetElement(detElement, "MSEndcapDE" + endcapType , endcapType);
+    EndcapDE = dd4hep::DetElement(detElement, "MSEndcapDE" + std::to_string(endcapType) , endcapType);
```
jmcarcell, line 619: In this case `endcapType` will be `-1` (see https://github.com/key4hep/k4geo/blob/master/detector/muonSystem/muonSystemMuRWELL_o1_v01.cpp#L610)  so it will look like "MSEndcapDE-1" and you can only know it's -1 if you see it with other numbers like 0, 1.

BrieucF:
Thanks for spotting this Juan! I guess it would be difficult for the tests to see this, there is nothing wrong per-se and the names were still "unique" since the number of character being cut were different.

detector/muonSystem/muonSystemMuRWELL_o1_v01.cpp
```diff
@@ -526,7 +526,7 @@ static dd4hep::Ref_t createmuonSystemMuRWELL_o1_v01(dd4hep::Detector& lcdd,
   dd4hep::Position detectorLayerTrans(0., 0., 0.);
   dd4hep::PlacedVolume detectorLayerPhys = BarrelVolume.placeVolume(BarrelDetectorLayerVolume, dd4hep::Transform3D(dd4hep::RotationZ(0.), detectorLayerTrans));
   detectorLayerPhys.addPhysVolID("layer", numBarrelLayer+1); 
-  dd4hep::DetElement BarrelDetectorLayerDE(BarrelDE, "MSBarrelDetectorLayerDE" + numBarrelLayer+1, numBarrelLayer+1);
+  dd4hep::DetElement BarrelDetectorLayerDE(BarrelDE, "MSBarrelDetectorLayerDE" + std::to_string(numBarrelLayer+1), numBarrelLayer+1);
```
BrieucF, line 529:
```suggestion
  dd4hep::DetElement BarrelDetectorLayerDE(BarrelDE, "MSBarrel_DetectorLayerDE_" + std::to_string(numBarrelLayer+1), numBarrelLayer+1);
```

detector/muonSystem/muonSystemMuRWELL_o1_v01.cpp
```diff
@@ -591,7 +591,7 @@ static dd4hep::Ref_t createmuonSystemMuRWELL_o1_v01(dd4hep::Detector& lcdd,
 
     dd4hep::Position radiatorLayerTrans(0., 0., 0.);
     dd4hep::PlacedVolume radiatorLayerPhys = BarrelVolume.placeVolume(BarrelRadiatorLayerVolume, dd4hep::Transform3D(dd4hep::RotationZ(0.), radiatorLayerTrans));
-    dd4hep::DetElement BarrelRadiatorLayerDE(BarrelDE, "MSBarrel_RadiatorLayerDE" + numBarrelRadiatorLayer+1, numBarrelRadiatorLayer+1);
+    dd4hep::DetElement BarrelRadiatorLayerDE(BarrelDE, "MSBarrel_RadiatorLayerDE" + std::to_string(numBarrelRadiatorLayer+1), numBarrelRadiatorLayer+1);
```
BrieucF, line 594:
```suggestion
    dd4hep::DetElement BarrelRadiatorLayerDE(BarrelDE, "MSBarrel_RadiatorLayerDE_" + std::to_string(numBarrelRadiatorLayer+1), numBarrelRadiatorLayer+1);
```

detector/muonSystem/muonSystemMuRWELL_o1_v01.cpp
BrieucF: Indeed. It is not a big deal but this would slightly help
```suggestion
    EndcapDE = dd4hep::DetElement(detElement, "MSEndcapDE_" + std::to_string(endcapType) , endcapType);
```

the other namings could be homogenized as well, e.g. https://github.com/key4hep/k4geo/blob/a63787154f6c185b715c7bb998fb0e1ad08a2fc3/detector/muonSystem/muonSystemMuRWELL_o1_v01.cpp#L634 but we could let @mahmoudali2 choose a naming convention and homogenize. @mahmoudali2 , do you want to do it in a separated PR ?

jmcarcell:
I'm merging this since it has been up for 1 month

PR #356, opened 2024-07-18
Add dd4hep-drawer util

Zehvogel:
BEGINRELEASENOTES
- ...

ENDRELEASENOTES

Allows to draw parts of the detector tree while de-duplicating elements with counters in their names.
I did not quickly find something like this so it might also be useful for others :)

example output
[CLD_o2_v06.gv.pdf](https://github.com/user-attachments/files/16284897/CLD_o2_v06.gv.pdf)

### DIFF
--- a/utils/dd4hep-drawer.py
+++ b/utils/dd4hep-drawer.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+from dd4hep import Detector
+import graphviz
+import re
+from collections import Counter
+
+from argparse import ArgumentParser
+
+parser = ArgumentParser()
+
+parser.add_argument(
+    "--compactFile", help="DD4hep compact description xml", required=True
+)
+parser.add_argument(
+    "--maxDepth",
+    help="Maximum traversal depth of the detector tree",
+    default=10,
+    type=int,
+)
+parser.add_argument(
+    "--maxEdges", help="Maximum number of edges per connection", default=5, type=int
+)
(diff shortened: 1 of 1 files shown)


atolosadelgado:
Hi @Zehvogel 

I think it is useful indeed, thank you! 

Do you think it could be extended to display a similar diagram for the volume tree? I posted [here](https://github.com/AIDASoft/DD4hep/issues/1297#issuecomment-2230747626) a similar function that walks though the volume tree, but in C++... :/

I think visualizing the volume tree is useful to spot potential optimizations of geometry

tmadlener:
Is it possible to color the boxes of each layer in a different color? Or are they guaranteed to align horizontally for each layer?

Zehvogel:
> Is it possible to color the boxes of each layer in a different color? Or are they guaranteed to align horizontally for each layer?

It is possible to color the nodes. It is plain graphviz so [any node attribute](https://graphviz.org/docs/nodes/) should work.
Layers are not guaranteed to align. They do align when starting from the deepest common element, which is probably not what you are after... The default layout engine is [dot](https://graphviz.org/docs/layouts/dot/). It might be possible to tune that but I do not care enough at the moment :)

Here another example of a detector with a bit more consistent structure where it is better visible what aligns when.

[OpenDataDetector_OpenDataTracker_Pixels.gv.pdf](https://github.com/user-attachments/files/16348820/OpenDataDetector_OpenDataTracker_Pixels.gv.pdf)

PR #362, opened 2024-07-30
Fix CLD endcap radius calculation for CED

### LINKED ISSUE #355: CLD tracker endcaps too small in CED
They are the right size in geoDisplay. @gaede @tmadlener you recently fixed the shape, do you also have an idea what we can do about the size? :) 

The outer radius should roughly match the grey lines

image

Zehvogel:
The radius calculation was very broken, the radius supplied by the xml is already the inner radius and the Y half-length instead of the Z half-length needs to be used, Z is just the thickness of the modules... Thankfully this was only used for the event display, the placement of the modules seems to be correct :)

BEGINRELEASENOTES
- Fixed CLD endcap radius calculation for CED

ENDRELEASENOTES

fixes #355 

before: check issue above

after:
image

### DIFF
--- a/detector/tracker/TrackerEndcap_o2_v06_geo.cpp
+++ b/detector/tracker/TrackerEndcap_o2_v06_geo.cpp
@@ -184,11 +184,11 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
             
             Box mod_shape(m_vol.solid());
             
-             if(r - mod_shape->GetDZ()<innerR)
-                 innerR= r- mod_shape->GetDZ();
+             if(r < innerR)
+                 innerR = r;
              
-             if(r + mod_shape->GetDZ()>outerR)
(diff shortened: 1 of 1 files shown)


andresailer (approved):

detector/tracker/TrackerEndcap_o2_v06_geo.cpp
```diff
@@ -184,11 +184,11 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
             
             Box mod_shape(m_vol.solid());
             
-             if(r - mod_shape->GetDZ()<innerR)
-                 innerR= r- mod_shape->GetDZ();
+             if(r < innerR)
+                 innerR = r;
```
jmcarcell, line 188: A bit pedantic, these cases can be stated like `innerR = std::min(innerR, r)` which I think better shows the intention.

PR #367, opened 2024-08-07
New DRC sensitive detector using RegEx

SanghyunKo:
BEGINRELEASENOTES
- Skip the scintillation optical generation for the (monolithic) DRC and migrate to the `Geant4RegexSensitivesConstruction`.

ENDRELEASENOTES

(Rewrite the PR description 29 Jan, 2025)

The previous (monolithic) DRC consumed a large amount of computation power due to its feature to simulate the scintillation process. We had to make each fiber the sensitive detector to migrate to the simpler simulation on par with other calorimeters that record the energy deposits inside the active material (previously, this wasn't necessary since we collected photons at the readout end of the tower). Obviously, this created the issue with large memory consumption observed & discussed in #411. Hence, the SD is switched to the `Geant4RegexSensitivesConstruction`.

Some performance tests on the lxplus: The 10 GeV electron gun event simulation takes about ~ 1s/evt. The memory consumption of the full IDEA v1 geometry is about 5 GB. The number of Cherenkov photons per GeV is about 55, which agrees with the testbeam observation. The geometry information retrieved using `CellID` seems reasonable.

### DIFF
--- a/FCCee/IDEA/compact/IDEA_o1_v03/DectDimensions_IDEA_o1_v03.xml
+++ b/FCCee/IDEA/compact/IDEA_o1_v03/DectDimensions_IDEA_o1_v03.xml
@@ -16,15 +16,15 @@
 
   <define>
     <constant name="world_side" value="6100*mm"/>           <!-- Used in LumiCal and HOMAbsorber, != world_size -->
-    <constant name="CrossingAngle" value="0.030*rad"/>  
+    <constant name="CrossingAngle" value="0.030*rad"/>
 
     <constant name="GlobalTrackerReadoutID" type="string" value="system:5,side:-2,layer:3,module:16,sensor:6"/>
 
     <constant name="SolenoidField" value="2*tesla"/>
-    
+
     <!-- Detector IDs -->
     <constant name="DetID_NOTUSED"                value=" 0"/>
-    
+
     <constant name="DetID_VXD_Barrel"             value=" 1"/>
     <constant name="DetID_VXD_Disks"              value=" 2"/>
 
@@ -40,10 +40,10 @@
     <constant name="DetID_HCAL_Barrel"            value=" 10"/>
     <constant name="DetID_HCAL_Endcap"            value=" 11"/>
     <constant name="DetID_HCAL_Ring"              value=" 12"/>
-    
+
     <constant name="DetID_Yoke_Barrel"            value=" 13"/>
     <constant name="DetID_Yoke_Endcap"            value=" 14"/>
-    
+
     <constant name="DetID_LumiCal"                value=" 15"/>
     <constant name="DetID_LumiCalInstrumentation" value=" 16"/>
     <constant name="DetID_LumiCalCooling"         value=" 17"/>
@@ -68,7 +68,7 @@
     <constant name="SeparatedBeamPipe_z" value="1190.0*mm"/> <!-- was 1159.97*mm in FCCDetectors/> -->
     <constant name="CentralBeamPipe_rmax" value="10.0*mm"/>
     <constant name="ConeBeamPipe_Rmax" value="28.9*mm" />
-    <constant name="BeamPipeGoldWidth" value="0.005*mm" /> 
+    <constant name="BeamPipeGoldWidth" value="0.005*mm" />
     <constant name="BeamPipeGoldTolerance" value="0.001*mm" />  <!-- dummy tolerance, some small non zero value -->
     <constant name="BeamPipeConeHalfAngle" value="(ConeBeamPipe_Rmax + BeamPipeWidthFirstCone - CentralBeamPipe_rmax ) / (SeparatedBeamPipe_z - CentralBeamPipe_zmax)" />
 
@@ -82,7 +82,7 @@
     <constant name="MiddleOfSRMask_z" value="2.1*m" />
     <constant name="SynchRadMaskSize" value="5*mm" /> <!-- mask tip is at 10 mm from the beamline -->
     <constant name="mask_epsilon" value="0.001*mm" />
-        
+
     <!-- Vertex detector. Changing the values here it not enough to resize the detector, contact expert (Armin Ilg) -->
     <constant name="VertexClearanceTheta" value="0.1035*rad"/> <!-- Clearance of vertex detector in mrad from IP. !!! Too high currently, lumical acceptance is 107 mrad so consider 110 mrad as limit -> Need to adapt vertex disks!!! -->
     <constant name="VTXIB_r_min_clearance" value="1.0*mm"/>    <!-- Clearance of vertex detector in radius, used for definiton of vertex DD4hep_SubdetectorAssembly -->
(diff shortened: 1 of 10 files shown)


BrieucF:
> With these, the CPU consumption is now suppressed down to a few seconds per event.

:heart_eyes:

How long does it take to build the geometry now?

SanghyunKo:
@BrieucF I didn't measure it precisely, but maybe around 10mins? (when I tested with the `DRCfastSDsteer.py` in `lxplus9`, with the full IDEA detector) But as I mentioned in the description, one really needs to be careful when matching regex pattern, because with bad/complicated patterns this can explode easily (performing regex matching O(10^6) time is a expensive operation)

example/DRCfastSDsteer.py
```diff
@@ -0,0 +1,105 @@
...
+  # Configure G4 geometry setup
+  seq, act = geant4.addDetectorConstruction('Geant4DetectorGeometryConstruction/ConstructGeo')
+  #
+  # Assign sensitive detectors according to the declarations 'tracker' or 'calorimeter', etc
+  # seq, act = geant4.addDetectorConstruction('Geant4DetectorSensitivesConstruction/ConstructSD')
+  # Assign sensitive detectors in Geant4 by matching a regular expression in the detector sub-tree
+  seq, act = geant4.addDetectorConstruction('Geant4RegexSensitivesConstruction/ConstructSDRegEx')
+  act.Detector = 'DRcalo'
+  act.OutputLevel = Output.ALWAYS
+  act.Match = ['(.*)(core|clad)(.*)'] # regex is expensive, always try to avoid complex match
```
andresailer, line 49:
```suggestion
  act.Match = ['(core|clad)'] # regex is expensive, always try to avoid complex match
```
Is that not enough?

andresailer: So you need the full expression, because dd4hep uses `regex_match` and not [https://en.cppreference.com/w/cpp/regex/regex_search](regex_search)?
Would regex_search be faster?

lopezzot:
Hi @SanghyunKo thank you very much for this example.

We will need to check how to use the volume hierarchy with the tubes geometry but I am pretty sure it is not hard to do. The optical photons treatment you just described is identical to what we are doing for Hidra simulation since a few years. Indeed I can confirm a few seconds per event (depending on energy and slightly on particle type).

Apart from this, I will study with @s6anloes this solution and get back to you.

example/DRCfastSDsteer.py
SanghyunKo: Indeed I need the full expression, but in general I think it is already a fair trade to spend 300s to match ~ 600k volumes and save 9GB of memory (and I'm not sure how much we will gain by using `regex_search` over `regex_match`). For example, I got the following from the `lxplus9`:

```sh
ConstructSDRegEx       +++ Detector: DRcalo Handled 580482 nodes with 580482 sensitive volume type(s). Total of 237.361 seconds.
```

The most important thing from my experience is not using a list, e.g. one should avoid using
```python
act.Match = ['(.*)(core)(.*)','(.*)(clad)(.*)']
```
over
```python
act.Match = ['(.*)(core|clad)(.*)']
```
(and users can handle this).

andresailer: Using your branch and example
regex_match (with full regex)
```
ConstructSDRegEx       +++ Detector: DRcalo Handled 580482 nodes with 580482 sensitive volume type(s). Total of 108.905 seconds.
```
regex_search (with simpler regex)
```
ConstructSDRegEx       +++ Detector: DRcalo Handled 580482 nodes with 580482 sensitive volume type(s). Total of  68.064 seconds.
```

We should probably also change the verbosity level to avoid printing all matched volumes?

SanghyunKo: Nice! And yes, we should suppress printouts.

andresailer: I'll make the changes in DD4hep, and add the RegexSD construction to ddsim.

andresailer: https://github.com/AIDASoft/DD4hep/pull/1308

SanghyunKo:
since https://github.com/AIDASoft/DD4hep/pull/1308 is now merged, I'll update the steering file shortly once we have new nightlies

atolosadelgado:
Hi @SanghyunKo @lopezzot 

Could you remind me what are the next steps to be done in this PR?

Best,
Alvaro

lopezzot:
> Hi @SanghyunKo @lopezzot
> 
> Could you remind me what are the next steps to be done in this PR?
> 
> Best, Alvaro

Hello @atolosadelgado, this was opened as a draft to discuss about regexSD (thank you @SanghyunKo !). regexSD is working fine in IDEA_o2 and is now available in key4hep (not just the nightlies). Other than that I let Sanghyun comment about this PR.

SanghyunKo:
@atolosadelgado apart from the obvious checks that I have to make (e.g. fix conflicts, testing with new nightlies since it's been a while), I'll probably need to revisit the strategy on how to retrieve `volID`.

@lopezzot it seems that you eventually settled downed with using the G4 copy numbers, if I see the PR #411. Is it correct?

lopezzot:
> @lopezzot it seems that you eventually settled downed with using the G4 copy numbers, if I see the PR #411. Is it correct?

Yes eventually I managed. Happy to discuss about it if needed.

SanghyunKo:
@BrieucF, sorry for the long wait for this PR; this should now be ready to be merged (and hence IDEA o1). There are some TODO items (as seen in the code), but this is because we still don't have any digitization module (now I'm applying the SiPM efficiency on the fly during the simulation step). This should be resolved by a big chunk of upcoming development on the new digitization module - for this PR, I'd like to wrap up and move on to the next item.

atolosadelgado (approved):
When I run the steering file provided as example, I see a warning but it is probably ok

```
*** G4Exception : OpBoun06
      issued by : G4OpBoundaryProcess
G4OpBoundaryProcess: Opticalphoton step length: 1.00382e-09 mm.
This is larger than the threshold 1e-09 mm to set status StepTooSmall.
Boundary scattering may be incorrect. 

```

Please remember to put the efficiency array (and maybe the code that emulates the efficiency) somewhere else; for instance, the efficiency table may be placed in the compact file?

In addition, the steering file setups the range cut to None, which is interpreted as default value of 0.7 mm. Probably this is ok now, but it may play a role in the future.

atolosadelgado:
Hi @SanghyunKo 

Could you please resolve the rebase conflicts? Otherwise, I won’t be able to merge it.

SanghyunKo:
@atolosadelgado Apologies for the `push -f`... It was impossible to resolve the rebase conflict of this 7-month-old branch (it kept trying to undo my developments), so I finger-snapped and made all the commits into one giant commit.

As for the efficiency array, it's on my TODO list (also written inside the comment of the code); but to make this efficiency array sort of Python configuration to the input of a Gaudi algorithm, we first need to develop the digitization module first out-of-the-box. We need another round of large-scale development for this feature, so I'd be happy if we could live with a hard-coded efficiency array for a short period. Otherwise, we won't be able to simulate IDEAo1 forever unless we develop everything in advance.

PR #369, opened 2024-08-08
Add CLD_o2_v07

Zehvogel:
Improves load time from 33s to 1s on my machine. :) 

I am also planning to add layer assemblies for the endcaps, but that is less performance relevant. Additionally, I also want to add a new tracking region definition to exclude the LumiCal now that that is possible.

BEGINRELEASENOTES
- Added TrackerBarrel_o1_v06 using a stave assembly instead of directly placing the sensors into the layers
- Added CLD_o2_v07 using the new TrackerBarrel_o1_v06

ENDRELEASENOTES

TODO:
- [x] Check if surfaces are still pointing in the right directions
- [x] cleanup, add back the ZPlanarData and NeighbourSurfacesData
- [x] Physics validation
- [x] ~~Document better that a module in the xml is not the same as what is called a module in the encoding/structure~~
- [x] Add new tracking region
- [x] Update README
- [x] Run overlap check
- [x] Repair surfaces

Evaluation results:
- Only unrelated overlaps in the LumiCal
- Tracking surface origin, u and v vector differences between old and new TrackerBarrel less than 1 fm
- Able to run reconstruction on files produced with the previous version and reproduce exactly the same digitised hits and tracks

### DIFF
--- a/FCCee/CLD/compact/CLD_o2_v07/BeamInstrumentation_o3_v02_fitShield.xml
+++ b/FCCee/CLD/compact/CLD_o2_v07/BeamInstrumentation_o3_v02_fitShield.xml
@@ -1,0 +1,26 @@
+<lccdd>
+
+    <detectors>
+        <comment>Beampipe Instrumentation</comment>
+
+        <detector name="CompSol" type="DD4hep_Mask_o1_v01" insideTrackingVolume="true" vis="CompSolVis">
+            <parameter crossingangle="CrossingAngle" />
+            <envelope vis="CoilVis">
+                <shape type="Assembly"/>
+            </envelope>
+            <section type="Center"               start="CompSol_min_z"    end="QD0_min_z" rMin1="103*mm"  rMin2="180*mm"  rMax1="123*mm" rMax2="200*mm"    material="SolenoidMixture"  name="CompSol" />
+        </detector>
+
+        <detector name="ScreenSol" type="DD4hep_Mask_o1_v01" insideTrackingVolume="true" vis="ScreenSolVis">
+            <parameter crossingangle="CrossingAngle" />
+            <envelope vis="CoilVis">
+                <shape type="Assembly"/>
+            </envelope>
+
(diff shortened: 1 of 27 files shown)


Zehvogel:
I added a polycone for the tracking volume to exclude the LumiCal. It should still contain all of the beam pipe seen by prompt particles. Alternatively we could use a boolean shape where just the LumiCal envelope is removed from the cylindrical tracking volume, but I guess that is slightly slower?

I am also not sure yet what the correct setting for `connected` would be here:
```
    <!--
       If the volume should be connected to the world:   connected="true"
       else if the volume is part of the parallelworld:  connected="false"
       The volume is always connected to the top level!
    -->
    <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="true"
```

Zehvogel:
Some references for the people interested in the tracking region business (@atolosadelgado): 

https://github.com/AIDASoft/DD4hep/pull/1303
https://github.com/AIDASoft/DD4hep/pull/384

Zehvogel:
Size of the `MCParticles` collection when using `Geant4TCUserParticleHandler` (old) vs. `Geant4TVUserParticleHandler` (new) for a 100 electrons at 10-100 GeV and theta of 50-100 mrad, i.e. directly into the LumiCal. All the shower particles are no longer in the MCParticle collection :)

image

image

andresailer:
I have most of the overlaps related to the LumiCal fixed https://github.com/andresailer/lcgeo/commit/6e48fc5d92537ee9df70d7f732b91cad20e61fe0

There is one between the lumical and the innertracker envelope that I still need to work out.

andresailer:
image

andresailer:
image

No more overlaps https://github.com/andresailer/lcgeo/commit/df245895249e8cfd55d022fddd42471285bfe642

andresailer:
@Zehvogel Let's merge this, and I make a PR with my commits?

Zehvogel:
ok

PR #372, opened 2024-08-08
[FCCeeMDI] Use absolute path to import CAD files

BrieucF:
BEGINRELEASENOTES
- [FCCeeMDI] Use absolute path to import CAD files

ENDRELEASENOTES

It looks like, when importing CAD file with `<shape type="CAD_MultiVolume" ref="path_to_stl_file" unit="mm" material="LiquidNDecane">`, the path is relative to where the ddsim command is launched

### DIFF
--- a/FCCee/MDI/compact/MDI_o1_v01/Beampipe_CADimport_o1_v02.xml
+++ b/FCCee/MDI/compact/MDI_o1_v01/Beampipe_CADimport_o1_v02.xml
@@ -19,7 +19,7 @@
 
     <detector name="Beampipe_STL" type="DD4hep_TestShape_Creator">
       <check>
-        <shape type="CAD_MultiVolume" ref="./stl_files/Pipe_240430/AlBeMet162_30042024.stl"
+        <shape type="CAD_MultiVolume" ref="${K4GEO}/FCCee/MDI/compact/MDI_o1_v01/stl_files/Pipe_240430/AlBeMet162_30042024.stl"
 	       unit="mm" material="AlBeMet162">
 	  <volume id="0" name="s1" vis="AlBeMet_vis" material="AlBeMet162"> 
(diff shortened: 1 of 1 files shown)


andresailer:
This should work https://github.com/AIDASoft/DD4hep/pull/1175  with recent enough DD4hep versions?

BrieucF:
I have opened an issue in DD4hep (https://github.com/AIDASoft/DD4hep/issues/1313). Can we merge this regardless of the issue? I'd keep it like this even after the fix.

andresailer:
I think this rather is an issue for key4hep-spack though. Because it comes down to ROOT with pcre2 (I think), which we have no control over in DD4hep.

BrieucF:
I see. Does this prevent merging this PR?

andresailer:
Apart from the full path making it harder to changing the files and trying things out with ones own copy of a folder. No, nothing prevents merging these changes.

andresailer:
By the way

>It looks like, when importing CAD file with <shape type="CAD_MultiVolume" ref="path_to_stl_file" unit="mm" material="LiquidNDecane">, the path is relative to where the ddsim command is launched

Is this your interpretation of the TRegexp error mentioned in the DD4hep issue, or was this a different error message that caused the conclusion written here?

BrieucF:
> Is this your interpretation of the TPRegexp error mentioned in the DD4hep issue, or was this a different error message that caused the conclusion written here?

No no, it comes from the fact that running `examples/DDCAD/compact/Check_Shape_RelativePath.xml` from the DD4hep directory does not work, while going to `examples/DDCAD/compact/` and running it from there works, as I tried to explain in https://github.com/AIDASoft/DD4hep/issues/1313. The mention of the TPRegexp error was just a side note that it is not easy to understand that the problem is "not finding the CAD file" based on the error message.

andresailer:
It does not work in the way that there is the TPRegex error?

PR #374, opened 2024-08-09
Fix a few compiler warnings

jmcarcell:
BEGINRELEASENOTES
- Fix a few compiler warnings

ENDRELEASENOTES

I have enabled `-Werror` for the Key4hep CI so that warnings will now be promoted to errors. In k4geo warnings are introduced in PRs all the time, so this will make CI fail if that is the case. Because there are already some, this PR is going to be necessary to pass CI.

### DIFF
--- a/detector/calorimeter/ECalEndcap_Turbine_o1_v01_geo.cpp
+++ b/detector/calorimeter/ECalEndcap_Turbine_o1_v01_geo.cpp
@@ -112,9 +112,9 @@ namespace det {
     float LArgapi = nobleLiquidElem.attr<float>(_Unicode(gap));
     
     bool sameNUnitCells = genericBladeElem.attr<bool>(_Unicode(sameNUnitCells));
-    char* nUnitCellsStrArr = (char*)genericBladeElem.attr<std::string>(_Unicode(nUnitCells)).c_str();
-    char* nUnitCellsCStr = strtok(nUnitCellsStrArr, " ");   
-    int nUnitCells;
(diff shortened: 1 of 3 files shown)


detector/calorimeter/ECalEndcap_Turbine_o1_v01_geo.cpp
```diff
@@ -112,9 +112,9 @@ namespace det {
     float LArgapi = nobleLiquidElem.attr<float>(_Unicode(gap));
     
     bool sameNUnitCells = genericBladeElem.attr<bool>(_Unicode(sameNUnitCells));
-    char* nUnitCellsStrArr = (char*)genericBladeElem.attr<std::string>(_Unicode(nUnitCells)).c_str();
-    char* nUnitCellsCStr = strtok(nUnitCellsStrArr, " ");   
-    int nUnitCells;
+    auto nUnitCellsStrArr = genericBladeElem.attr<std::string>(_Unicode(nUnitCells));
+    char* nUnitCellsCStr = strtok(const_cast<char*>(nUnitCellsStrArr.c_str()), " ");
+    int nUnitCells = 0;
```
jmcarcell, line 117: @varnes 
I'm not sure if this was working as expected before. Doing `c_str()` on a temporary will give you a bad pointer as the temporary string is destroyed after the line ends. `nUnitCells` was not initialized so unless the block inside the if below is triggered, `nUnitCells` will have a random value. If the if is always triggering then it should be deleted.

jmcarcell:
Merging this to allow other PRs to pass

PR #375, opened 2024-08-13
Add `print_overlaps` util

Zehvogel:
BEGINRELEASENOTES
- Added `print_overlaps` util to printout all overlaps from an overlap dump file

ENDRELEASENOTES

prints out all `G4Exception-START / G4Exception-END` blocks in the give file. For people like me that can't remember how to do this with one magic `grep` line...

### DIFF
--- a/utils/print_overlaps.py
+++ b/utils/print_overlaps.py
@@ -1,0 +1,18 @@
+from argparse import ArgumentParser
+import re
+
+
+def main():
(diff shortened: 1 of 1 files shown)


jmcarcell:
I think this tool is useful but as you say it's possible to do it in a different way so I'm not sure if it would get a lot of usage. Closing...

PR #391, opened 2024-09-12
CLD_o2_v07: Fix lumi cal back shield, change envelopes to assemblies

### LINKED ISSUE #306: Geant4 complaints about LumiCalBackShield (CLD)
Hi,

I have run a geometry overlap check of CLD_o3_v01, with resolution of 300k points, latest key4hep stack. The output looks fine except for this issue,

```

*** G4Exception : GeomSolids1001

     issued by : G4BooleanSolid::GetPointOnSurface()

Solid - LumiCalBackShield_envelope_shape_0x1b644830

All 100k attempts to generate a point on the surface have failed!

The solid created may be an invalid Boolean construct!

````

The cause of this message may be:
* the result of the Boolean operation is a NULL (empty) object, 
* or the size of the resulting solid is extremely small compared to the size of the components of the Boolean operation.

I leave this note here, otherwise I will forget. If someone else wants to investigate it, let me know :)

andresailer:
BEGINRELEASENOTES
- CLD_o2_v07: change LumiCal envelopes from boolean of boolean to assembly, fixes #306, speeds up overlap check (of LumiCal only) with /geometry/test/resolution 300000 down to 13s instead of 3m10s

ENDRELEASENOTES

### DIFF
--- a/FCCee/CLD/compact/CLD_o2_v07/LumiCal_o3_v02_05.xml
+++ b/FCCee/CLD/compact/CLD_o2_v07/LumiCal_o3_v02_05.xml
@@ -21,17 +21,13 @@
         <detector name="LumiCal" type="LumiCal_o1_v01" vis="SeeThrough" id="DetID_LumiCal" readout="LumiCalCollection" insideTrackingVolume="false" >
 
             <envelope vis="LCALVis">
-                <shape type="BooleanShape" operation="union" material="Air">
-                    <shape type="BooleanShape" operation="Intersection">
-                        <shape type="Box"  dx="world_side" dy="world_side" dz="world_side"/>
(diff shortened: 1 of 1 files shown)


Zehvogel (approved):
LGTM :+1: 

I wonder what the point of the intersections was? Were they necessary to be put into the union?

andresailer:
> I wonder what the point of the intersections was? Were they necessary to be put into the union?

I was asking myself the same thing, of why using bool of bool instead of assembly.
Unions are supposed to be "well overlapping", so doing an union of just the tubes was giving an exception.

atolosadelgado:
I have run a quick simulation with geantinos, it seems to stay in the same place within 10 um tolerance, so it is ok for me :)

atolosadelgado (approved):

PR #392, opened 2024-09-17
Configurability for material plots

Victor-Schwan:
BEGINRELEASENOTES
- formatting and refactorization
- `material_plots_2D.py`:
  - `angleDef` is a choice parameter + `thetaRad` choice added
  - `outputDir` option added
- `material_scan_2d.py`:
  - argParser added (`compactFile`,`outputFileBase`, `angleDef`)
  - `outputDir` option added

ENDRELEASENOTES

### DIFF
--- a/utils/material_plots_2D.py
+++ b/utils/material_plots_2D.py
@@ -108,13 +108,7 @@ def main():
             entry_x0 += entry.nX0.at(i) * 100.0
             entry_lambda += entry.nLambda.at(i)
             entry_depth += entry.matDepth.at(i)
-            entry_x0 += entry.nX0.at(i) * 100.0
-            entry_lambda += entry.nLambda.at(i)
-            entry_depth += entry.matDepth.at(i)
 
-        h_x0.Fill(tree.angle, tree.phi, entry_x0)
-        h_lambda.Fill(tree.angle, tree.phi, entry_lambda)
-        h_depth.Fill(tree.angle, tree.phi, entry_depth)
         h_x0.Fill(tree.angle, tree.phi, entry_x0)
         h_lambda.Fill(tree.angle, tree.phi, entry_lambda)
         h_depth.Fill(tree.angle, tree.phi, entry_depth)
--- a/utils/material_plots_2D.py
+++ b/utils/material_plots_2D.py
@@ -17,17 +17,21 @@ from plotstyle import FCCStyle
 import ROOT
 
 
-def create_histogram(name_and_title: str, argument_name_space: argparse.Namespace) -> ROOT.TH2F:
+def create_histogram(
+    name_and_title: str,
+    angle_min: float,
+    angle_max: float,
+    angle_binning: float,
+    n_phi_bins: int,
+) -> ROOT.TH2F:
+    num_bins = int((angle_max - angle_min) / angle_binning)
     return ROOT.TH2F(
         name_and_title,
         name_and_title,
-        int(
-            (argument_name_space.angleMax - argument_name_space.angleMin)
-            / argument_name_space.angleBinning
-        ),
-        argument_name_space.angleMin,
-        argument_name_space.angleMax,
-        argument_name_space.nPhiBins,
+        num_bins,
+        angle_min,
+        angle_max,
+        n_phi_bins,
         -math.pi,
         math.pi,
     )
@@ -88,9 +92,9 @@ def main():
 
     ROOT.gROOT.SetBatch(1)
 
-    h_x0 = create_histogram("h_x0", args)
-    h_lambda = create_histogram("h_lambda", args)
-    h_depth = create_histogram("h_depth", args)
+    h_x0 = create_histogram("h_x0", args.angleMin, args.angleMax, args.angleBinning, args.nPhiBins)
+    h_lambda = create_histogram("h_lambda", args.angleMin, args.angleMax, args.angleBinning, args.nPhiBins)
+    h_depth = create_histogram("h_depth", args.angleMin, args.angleMax, args.angleBinning, args.nPhiBins)
 
     for angleBinning, entry in enumerate(tree):
         nMat = entry.nMaterials
--- a/utils/material_plots_2D.py
+++ b/utils/material_plots_2D.py
@@ -18,6 +18,24 @@ from plotstyle import FCCStyle
 import ROOT
 
 
+def create_histogram(
+    name_and_title: str, argument_name_space: argparse.Namespace
+) -> ROOT.TH2F:
+    return ROOT.TH2F(
+        name_and_title,
+        name_and_title,
+        int(
+            (argument_name_space.angleMax - argument_name_space.angleMin)
+            / argument_name_space.angleBinning
+        ),
+        argument_name_space.angleMin,
+        argument_name_space.angleMax,
+        argument_name_space.nPhiBins,
+        -math.pi,
+        math.pi,
+    )
+
+
 def main():
     parser = argparse.ArgumentParser(description="Material Plotter")
     parser.add_argument("--inputFile", "-f", type=str, help="relative path to the input file")
@@ -76,36 +94,9 @@ def main():
 
     ROOT.gROOT.SetBatch(1)
 
-    h_x0 = ROOT.TH2F(
-        "h_x0",
-        "h_x0",
-        int((args.angleMax - args.angleMin) / args.angleBinning),
-        args.angleMin,
-        args.angleMax,
-        args.nPhiBins,
-        -math.pi,
-        math.pi,
-    )
-    h_lambda = ROOT.TH2F(
-        "h_lambda",
-        "h_lambda",
-        int((args.angleMax - args.angleMin) / args.angleBinning),
-        args.angleMin,
-        args.angleMax,
-        args.nPhiBins,
-        -math.pi,
-        math.pi,
-    )
-    h_depth = ROOT.TH2F(
-        "h_depth",
-        "h_depth",
-        int((args.angleMax - args.angleMin) / args.angleBinning),
-        args.angleMin,
-        args.angleMax,
-        args.nPhiBins,
-        -math.pi,
-        math.pi,
-    )
+    h_x0 = create_histogram("h_x0", args)
+    h_lambda = create_histogram("h_lambda", args)
+    h_depth = create_histogram("h_depth", args)
 
     for angleBinning, entry in enumerate(tree):
         nMat = entry.nMaterials
--- a/utils/material_plots_2D.py
+++ b/utils/material_plots_2D.py
@@ -93,8 +93,12 @@ def main():
     ROOT.gROOT.SetBatch(1)
 
     h_x0 = create_histogram("h_x0", args.angleMin, args.angleMax, args.angleBinning, args.nPhiBins)
-    h_lambda = create_histogram("h_lambda", args.angleMin, args.angleMax, args.angleBinning, args.nPhiBins)
-    h_depth = create_histogram("h_depth", args.angleMin, args.angleMax, args.angleBinning, args.nPhiBins)
+    h_lambda = create_histogram(
+        "h_lambda", args.angleMin, args.angleMax, args.angleBinning, args.nPhiBins
+    )
+    h_depth = create_histogram(
+        "h_depth", args.angleMin, args.angleMax, args.angleBinning, args.nPhiBins
+    )
 
     for angleBinning, entry in enumerate(tree):
         nMat = entry.nMaterials
--- a/utils/material_plots_2D.py
+++ b/utils/material_plots_2D.py
@@ -144,8 +144,8 @@ def main():
 
         histograms[i].GetZaxis().SetTitle(axis_titles[i])
 
-        if args.x0Max != 0.0 and plot == "x0":
-            histograms[i].SetMaximum(args.x0Max)
+        if args.x0max != 0.0 and plot == "x0":
+            histograms[i].SetMaximum(args.x0max)
 
         histograms[i].GetXaxis().SetRangeUser(args.angleMin, args.angleMax)
 
--- a/utils/material_plots_2D.py
+++ b/utils/material_plots_2D.py
@@ -1,9 +1,17 @@
+"""
+This script must be called with python: 'python material_scan_2D.py --{argument} {value}'.
+The output files are saved in data/{outputDir}/name.suffix.
+If no outputDir is specified, it will be data/plots/name.suffix.
+"""
+
 from __future__ import print_function
 
 import argparse
 import math
-import os
 import sys
+from os import fspath
+from os.path import expandvars
+from pathlib import Path
 
 sys.path.append(os.path.expandvars("$FCCSW") + "/Examples/scripts")
 from plotstyle import FCCStyle
@@ -12,7 +20,8 @@ import ROOT
 
 def main():
     parser = argparse.ArgumentParser(description="Material Plotter")
-    parser.add_argument("--fname", "-f", dest="fname", type=str, help="name of file to read")
+    parser.add_argument("--inputFile", "-f", type=str, help="relative path to the input file")
+    parser.add_argument("--fname", "-f", dest="inputFile", type=str, help="name of file to read")
     parser.add_argument(
         "--angleMin", dest="angleMin", default=6, type=float, help="minimum eta/theta/cosTheta"
     )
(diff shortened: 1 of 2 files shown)


utils/material_plots_2D.py
```diff
@@ -1,80 +1,167 @@
...
 def main():
-    parser = argparse.ArgumentParser(description='Material Plotter')
-    parser.add_argument('--fname', "-f", dest='fname', type=str, help="name of file to read")
-    parser.add_argument('--angleMin', dest='angleMin', default=6, type=float, help="minimum eta/theta/cosTheta")
-    parser.add_argument('--angleMax', dest='angleMax', default=6, type=float, help="maximum eta/theta/cosTheta")
-    parser.add_argument('--angleDef', dest='angleDef', default="eta", type=str, help="angle definition to use: eta, theta or cosTheta, default: eta")
-    parser.add_argument('--angleBinning', "-b", dest='angleBinning', default=0.05, type=float, help="eta/theta/cosTheta bin width")
-    parser.add_argument('--nPhiBins', dest='nPhiBins', default=100, type=int, help="number of bins in phi")
-    parser.add_argument('--x0max', "-x", dest='x0max', default=0.0, type=float, help="Max of x0")                                                                                                                                                
+    parser = argparse.ArgumentParser(description="Material Plotter")
```
andresailer, line 23: Can we please not mix style changes and feature changes?

tmadlener:
[these are the changes with the formatting factored out](https://github.com/key4hep/k4geo/commit/b263d1e3ac34975c2267152f9f0d5da6ac1681ec)

andresailer:
> [these are the changes with the formatting factored out](https://github.com/key4hep/k4geo/commit/b263d1e3ac34975c2267152f9f0d5da6ac1681ec)

Now can we have those changes on top of the original style :zany_face:

tmadlener:
I am afraid that is not something I can do in 5 minutes ;)

Should we format all python sources and make a PR that we can then ignore in blames?

andresailer:
> Should we format all python sources and make a PR that we can then ignore in blames?

But something that reduces the amount of changes. Because #396 is going to be painful enough to merge.

utils/material_plots_2D.py
```diff
@@ -1,80 +1,167 @@
...
+        type=float,
+        help="Eta/theta/cosTheta bin width",
+    )
+    parser.add_argument(
+        "--nPhiBins",
+        default=100,
+        type=int,
+        help="Number of bins in phi",
+    )
+    parser.add_argument("--x0Max", "-x", default=0.0, type=float, help="Max of x0")
```
tmadlener, line 59:
```suggestion
    parser.add_argument("--x0max", "-x", default=0.0, type=float, help="Max of x0")
```
Can we keep the original capitalization? There is no real need to change it as far as I can see.

utils/material_plots_2D.py
```diff
@@ -1,80 +1,167 @@
...
+    h_depth = ROOT.TH2F(
+        "h_depth",
+        "h_depth",
+        int((args.angleMax - args.angleMin) / args.angleBinning),
+        args.angleMin,
+        args.angleMax,
+        args.nPhiBins,
+        -math.pi,
+        math.pi,
+    )
```
tmadlener, line 110: Apart from the names & titles these are the same histograms, this could be moved into a separate function that only takes the name as argument.

utils/material_plots_2D.py
```diff
@@ -1,80 +1,167 @@
...
-    parser.add_argument('--fname', "-f", dest='fname', type=str, help="name of file to read")
-    parser.add_argument('--angleMin', dest='angleMin', default=6, type=float, help="minimum eta/theta/cosTheta")
-    parser.add_argument('--angleMax', dest='angleMax', default=6, type=float, help="maximum eta/theta/cosTheta")
-    parser.add_argument('--angleDef', dest='angleDef', default="eta", type=str, help="angle definition to use: eta, theta or cosTheta, default: eta")
-    parser.add_argument('--angleBinning', "-b", dest='angleBinning', default=0.05, type=float, help="eta/theta/cosTheta bin width")
-    parser.add_argument('--nPhiBins', dest='nPhiBins', default=100, type=int, help="number of bins in phi")
-    parser.add_argument('--x0max', "-x", dest='x0max', default=0.0, type=float, help="Max of x0")                                                                                                                                                
+    parser = argparse.ArgumentParser(description="Material Plotter")
+    parser.add_argument(
+        "--inputFile", "-f", type=str, help="relative path to the input file"
```
andresailer, line 25: No pressing need to change the argument from --fname. Changing user interfaces should be avoided unless absolutely necessary., add --inputFile as an alternate if you really want to?

utils/material_plots_2D.py
Victor-Schwan: That was not my intention. I have activated auto-formatting (black + isort) in code and as soon as I open a script it is also formatted. In general, I think autoformatting makes a lot of sense in a project like this...

utils/material_plots_2D.py
```diff
@@ -1,80 +1,167 @@
...
+If no outputDir is specified, it will be data/plots/name.suffix.
+"""
+
 from __future__ import print_function
+
 import argparse
 import math
+import sys
+from os import fspath
+from os.path import expandvars
```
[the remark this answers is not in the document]
Victor-Schwan: I think it is more pythonic, shortens the following code (preventing newlines), and prevents namespace clutter.  One can still search for `fspath` and `expandvars`. But it is not a matter of life-and-death for me

utils/material_plots_2D.py
Victor-Schwan: I thought of it to use camelCase consistently. Then I do not have to remember which arguments do and which do not

utils/material_plots_2D.py
Victor-Schwan: I thought of it for two reasons:
1. That way, the argument name is more consistent with [ILDConfig](https://github.com/iLCSoft/ILDConfig/blob/d730298e40b55ff1fa5073a56f9ae510b93be6bc/StandardConfig/production/ILDReconstruction.py#L57). I prefer if scripts name similar arguments similarly
2. The name does not conform to camelCase which most arguments do

If it is used frequently, keeping `--fname` as well is certainly an option

utils/material_plots_2D.py
Victor-Schwan: good idea. done

utils/material_plots_2D.py
tmadlener: The issue in this case is not the re-formatting itself, but rather that the formatting changes and feature changes are mixed in one commit. In case there is an auto-formatter, the way to go would be: auto-format only first in a separate commit and then make a second commit with the feature changes.

That makes it easier to review the feature changes only, because we have a diff with only that. Additionally, git can be configured to ignore format only commits in `git blame`, which again makes it easier to figure out what actually changed in the future.

andresailer: > That was not my intention. I have activated auto-formatting (`black` + `isort`) in code and as soon as I open a script it is also formatted. In general, I think autoformatting makes a lot of sense in a project like this...

To put it bluntly and exaggerating :): No, it makes no sense like this, everyone using a different auto linter leads to way too much noise. When in Rome do as the Romans. When there is no linter config, don't lint by default.

So in this case the issue is the re-formatting :)

Yes auto-formatting is nice, but sadly there is a cost for "legacy" code.

Victor-Schwan: I see. I think I managed to add a commit that isolates the style changes :)

utils/material_plots_2D.py
andresailer: Let me say that I can see where you are coming from, but really the big picture matters here, and you have to consider other people as well. Your preferences are not supreme. 

So, no changes to user interfaces unless absolutely necessary. There might be existing documentation somewhere, people have scripts that call this script with this argument. Or the original author just preferred fname and things are as they are, that ship has sailed.

Or as [Linus](https://lkml.org/lkml/2012/12/23/75) puts it "WE DO NOT BREAK USERSPACE!"
(the expletive in that link is not aimed at you), but I have / had to deal with too many callous interface changes in the past.

So, if you want to add `inputName` that is OK, you have a valid argument for that. But you do not have any valid argument to break existing usage.

utils/material_plots_2D.py
andresailer: If you really want to, add an alternative flag, but see the [other comment.](https://github.com/key4hep/k4geo/pull/392/files#r1776702228)

utils/material_plots_2D.py
Victor-Schwan: Fair point. Thanks for the explanation. I will make sure that the user interface is not broken :)

andresailer: Thanks for your understanding!

tmadlener:
Rebased this after #398 has been merged. Please check again that I didn't mess up in the process.

`--inputFile` is now available as well as the original `--fname` which now simply has `inputFile` as destination.

utils/material_plots_2D.py
```diff
@@ -1,17 +1,45 @@
...
+        -math.pi,
+        math.pi,
+    )
+
+
 def main():
     parser = argparse.ArgumentParser(description="Material Plotter")
-    parser.add_argument("--fname", "-f", dest="fname", type=str, help="name of file to read")
+    parser.add_argument("--inputFile", "-f", type=str, help="relative path to the input file")
+    parser.add_argument("--fname", "-f", dest="inputFile", type=str, help="name of file to read")
```
andresailer, line 42:
```suggestion
    parser.add_argument("--inputFile", "-f", "--fname", type=str, help="relative path to the input file")
```

utils/material_plots_2D.py
```diff
@@ -98,16 +110,31 @@ def main():
...
         h_depth.Fill(tree.angle, tree.phi, entry_depth)
+        h_x0.Fill(tree.angle, tree.phi, entry_x0)
+        h_lambda.Fill(tree.angle, tree.phi, entry_lambda)
+        h_depth.Fill(tree.angle, tree.phi, entry_depth)
 
-    # go through the
+    # go through the plots
     plots = ["x0", "lambda", "depth"]
     histograms = [h_x0, h_lambda, h_depth]
+<<<<<<< HEAD
```
andresailer, line 127: Conflict marker detected

utils/material_plots_2D.py
```diff
@@ -116,22 +143,25 @@ def main():
...
             title = "cos(#theta)"
         histograms[i].GetXaxis().SetTitle(title)
         histograms[i].GetYaxis().SetTitle("#phi")
 
         histograms[i].GetZaxis().SetTitle(axis_titles[i])
 
-        if args.x0max != 0.0 and plots[i] == "x0":
-            histograms[i].SetMaximum(args.x0max)
+        if args.x0Max != 0.0 and plot == "x0":
+            histograms[i].SetMaximum(args.x0Max)
```
andresailer, line 156:
```suggestion
        if args.x0max != 0.0 and plot == "x0":
            histograms[i].SetMaximum(args.x0max)
```
We still have the lowercase max, I think.

utils/material_scan_2D.py
```diff
@@ -1,7 +1,15 @@
-import os
-from Gaudi.Configuration import *
+<"""
```
andresailer, line 1:
```suggestion
"""
```

utils/material_plots_2D.py
andresailer: PS: Pretty sure the alternative gives
```
argparse.ArgumentError: argument --fname/-f: conflicting option string: -f
```

```python
import argparse
parser = argparse.ArgumentParser(description="foo")
parser.add_argument("--inputFile", "-f", type=str, help="relative path to the input file")
parser.add_argument("--fname", "-f", dest="inputFile", type=str, help="name of file to read")
```

utils/material_plots_2D.py
tmadlener: :see_no_evil: Thanks.

tmadlener:
Thanks for checking.

utils/material_plots_2D.py
```diff
@@ -1,17 +1,44 @@
...
+from os import fspath
+from os.path import expandvars
+from pathlib import Path
 
 sys.path.append(os.path.expandvars("$FCCSW") + "/Examples/scripts")
 from plotstyle import FCCStyle
 import ROOT
 
 
+def create_histogram(name_and_title: str, argument_name_space: argparse.Namespace) -> ROOT.TH2F:
```
tmadlener, line 21:
```suggestion
def create_histogram(name_and_title: str, argument_name_space: argparse.Namespace) -> ROOT.TH2F:
```
Can we make this take the number of bins as well as the min/max values instead of an `argument_name_space`? That would make this a bit more re-usable and less dependent to work only for these specific arguments.

utils/material_scan_2D.py
```diff
@@ -10,9 +18,42 @@
 # DD4hep geometry service
 from Configurables import GeoSvc
 
+parser.add_argument(
+    "--compactFile",
+    help="Compact detector file to use",
+    type=str,
+    default=fspath(
+        Path(environ["k4geo_DIR"]) / "ILD" / "compact" / "ILD_sl5_v02" / "ILD_l5_v02.xml"
```
tmadlener, line 26: This technically changes the default geometry from `IDEA_o1_v02.xml` to an ILD geometry. However, it also looks like the previous version didn't work without editing the file?

Technically this should be a positional (required) argument, since the script can't do anything useful without it. However, at this point `--compactFile` is also pretty much a convention for several scripts.

For me these changes are OK.

Victor-Schwan: I don't care which detector is the default as long as the path is absolute or relative to `$K4GEO` or `$k4geo_DIR`. Originally the path was relative and Daniel and I had trouble getting it to work. I think this should work out of the box without assuming a current working dir

tmadlener (approved):
I am OK with this. Anything else from your side @andresailer?

utils/material_scan_2D.py
andresailer: Originally which path was relative?

tmadlener: The path to the `IDEA_o1_v02.xml` compact file.

andresailer: Ahh, ok, the detector file was just hard coded. I guess the compactFile will accept relative or absolute paths.

tmadlener: Yes, exactly. And yes, absolute and relative paths will work. It is directly passed to the `GeoSvc.detectors`.

armin-ilg:
Thank you for the work @Victor-Schwan, it works fine for me! I think it would be nice to apply the same improvements to the 1D mat. budget scripts (in the same place) and update the [scripts/check_material_budget.sh](https://github.com/key4hep/k4geo/blob/main/scripts/check_material_budget.sh) script accordingly. What do you think @andresailer @tmadlener ?

tmadlener:
In the long run, I think harmonizing all of these scripts would be a good idea. Whether we do it in this PR or in follow up PRs depends a bit on the availability of @Victor-Schwan. I the bash script does not work any longer at the moment, we should fix it. @armin-ilg do you have some estimates how long this runs, and whether we could put this into a CI workflow (maybe in a slightly slimmed down version)?

Victor-Schwan:
I agree that it is a good idea to harmonize as much as possible. I'd prefer to do the harmonization in another PR. In principle, one could also turn the scripts in the scripts folder into python scripts to have the same `--compactFile` interface (keeping a bash script calling them not to change the ui). Is the linked script affected by the changes? The main change is the different default detector but the script forwards a compact file anyway...

armin-ilg:
> In the long run, I think harmonizing all of these scripts would be a good idea. Whether we do it in this PR or in follow up PRs depends a bit on the availability of @Victor-Schwan. I the bash script does not work any longer at the moment, we should fix it. @armin-ilg do you have some estimates how long this runs, and whether we could put this into a CI workflow (maybe in a slightly slimmed down version)?

On lxplus the plots take O(5) minutes (running all angle defs), so that could definitely go into a CI workflow

Victor-Schwan:
Should this script be added to a CI workflow in this PR or in another one?
Assuming that the 1D mat. budget scripts will be changed in another PR, is an update of the https://github.com/key4hep/k4geo/blob/main/scripts/check_material_budget.sh script required @armin-ilg?

armin-ilg:
> Should this script be added to a CI workflow in this PR or in another one? Assuming that the 1D mat. budget scripts will be changed in another PR, is an update of the https://github.com/key4hep/k4geo/blob/main/scripts/check_material_budget.sh script required @armin-ilg?

I was wrong, sorry, the bash script still works fine, no change needed.

Let me know if I shall make a CI workflow at some point @tmadlener  (will not be able to do it soon though)

tmadlener:
@Victor-Schwan can you estimate how much work it would be to simply run the bash script in CI? If it's only a few more lines, I would appreciate it if it's added in this PR, otherwise, I am also fine with a follow up.

Victor-Schwan:
> @Victor-Schwan can you estimate how much work it would be to simply run the bash script in CI? If it's only a few more lines, I would appreciate it if it's added in this PR, otherwise, I am also fine with a follow up.

I took another look at the script and noticed two things:

1. According to a comment, for over a year it only runs on the stable key4hep release (https://github.com/key4hep/k4geo/blob/main/scripts/check_material_budget.sh#L57)
2. In line 7, `IDEA` is hardcoded in a name (https://github.com/key4hep/k4geo/blob/main/scripts/check_material_budget.sh#L7). In my opinion, in the utils, scripts, etc, the detector choice should not be hardcoded/assumed; such a name can confuse.

I only use the nightlies for ILD@FCCee, hence, I cannot test the script easily with my detectors (I have never run it before). I suggest fixing the script and adding the CI in another PR

Victor-Schwan:
> > Should this script be added to a CI workflow in this PR or in another one? Assuming that the 1D mat. budget scripts will be changed in another PR, is an update of the https://github.com/key4hep/k4geo/blob/main/scripts/check_material_budget.sh script required @armin-ilg?
> 
> I was wrong, sorry, the bash script still works fine, no change needed.
> 
> Let me know if I shall make a CI workflow at some point @tmadlener (will not be able to do it soon though)

@armin-ilg Am I right in assuming that there is nothing left to stop you from approving this PR?

armin-ilg (approved):
All fine with me, thanks!

PR #396, opened 2024-09-24
Improvements to vertex/silicon wrapper

armin-ilg:
BEGINRELEASENOTES
- No more warnings in silicon wrapper
- Improvements in vertex builder printouts
- Adding all materials of beam pipe also to material_plots_2D.py, as without having the beam pipe enabled, the vertex material budget estimation will fail.
- Changed paths to .stl files in vertex to use https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/IDEA/IDEA_o1_v03/STL_FILES/ (still commented out due to overlaps)
ENDRELEASENOTES

@BrieucF: As you requested I fixed the warnings in the silicon wrapper.

P.s:
I don't know how to squash all the commits, sorry that this is such a mess

### DIFF
--- a/utils/material_plots.py
+++ b/utils/material_plots.py
@@ -34,11 +34,8 @@ def main():
             material = material.replace("66D","")
             material = material.replace("Vtx","")
 
-            if material == "Air": continue
-            if material == "Tungsten": continue
-            if material == "Copper": continue
-            if material == "beam": continue
-            if material in ["LiquidNDecane", "AlBeMet162", "Gold"]:
+            # Ignore materials used in beam pipe as vertex material budget does not work without beam pipe. You can add more materials to ignore here
+            if material in ["Air","Tungsten","Copper","beam","LiquidNDecane", "AlBeMet162", "Gold"]:
                 continue
             if material not in histDict.keys():
                 histDict[material] = {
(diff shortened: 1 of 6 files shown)


andresailer:
Can you stop touching your branch for 5 minutes?

utils/material_plots_2D.py
```diff
@@ -36,8 +36,9 @@ def main():
 
         entry_x0, entry_lambda, entry_depth = 0.0, 0.0, 0.0
         for i in range(nMat):
-            if entry.material.at(i) == "Air": continue
-
+            # Ignore materials used in beam pipe as vertex material budget does not work without beam pipe. You can add more materials to ignore here
```
andresailer, line 39: This makes this script very non-generic and confusing to use.
Maybe you add an option to ignore a list material names?

andresailer: Yes! Thanks!

utils/material_plots.py
```diff
@@ -30,16 +32,14 @@ def main():
         for i in range(nMat):
             material = entry.material.at(i)
 
-            # If you need to replace some string in the material, add that here
-            material = material.replace("66D","")
-            material = material.replace("Vtx","")
+            # Removing substrings from materials
+            for substring in args.removeMatsSubstrings:
+                material = material.replace(substring,"")
```
BrieucF, line 37: Isn't there an indentation problem here?

armin-ilg: I don't see a problem, the indentation is the same as everywhere else (four spaces) and the "Removing substrings from materials" and "Ignore certain materials if specified" functionalities should be at the same level

BrieucF: Indeed, I mis-read, sorry

BrieucF:
Shall we merge this before or after https://github.com/key4hep/k4geo/pull/392?

tmadlener:
There will most likely be a conflict either way. So, I don't have any strong preferences.

PR #397, opened 2024-09-25
[IDEA Pre-shower] Adopting IDEA muon system builder for pre-shower

mahmoudali2:
BEGINRELEASENOTES
- Generalizing the muon system builder to adopt pre-shower description, the changes include:
     - Making the variable names more general, not specific only for muon system.
     - Changing the detector side's volume thicknesses in case there is only one chambers row in the side (like the pre-shower case), and it's in general it is the case for any -almost- circular shape for the detector. 
     - Disallowing the overlap rotation in case of single chamber side.
- Overall, the changes make the builder more general to adopt different cases with different structures (polyhedron & cylinder of chambers).

ENDRELEASENOTES

### DIFF
--- a/FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml
+++ b/FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml
@@ -60,6 +60,9 @@
   <!-- Import Endcap plate absorber -->
   <include ref="EndPlateAbsorber_o1_v01.xml"/>
 
+  <!-- Import Endcap pre-shower -->
+  <include ref="Preshower_o1_v01.xml"/>
+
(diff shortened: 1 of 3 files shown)


FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml
```diff
@@ -60,6 +60,9 @@
   <!-- Import Endcap plate absorber -->
   <include ref="EndPlateAbsorber_o1_v01.xml"/>
 
+  <!-- Import Endcap pre-shower -->
```
BrieucF, line 63:
```suggestion
  <!-- Import pre-shower -->
```

mahmoudali2:
Overlap check is clean for both Pre-shower, and Muon-system.

PR #398, opened 2024-09-25
CI: Add python formatting to pre-commit and make all files comply

tmadlener:
BEGINRELEASENOTES
- Switch the `pre-commit` action to run in a Key4hep environment
- Add `ruff` formatting to `pre-commit`
- Fix a few python2 style `print` statements
- Fix format of all python files

ENDRELEASENOTES

As discussed in #392 and #396. Given that some of the files still had python2 style `print` statements, `ruff` didn't work on all of them without intervention. I have made a separate commit that only fixes those (semi-automated), and then formatted everything afterwards.


tmadlener:
@andresailer should we try formatting with longer lines to reduce a few of the line breaks? We could also configure a linter, but I am not sure how well that deals with all the dynamic imports from DD4hep and/or Gaudi. Plus, I am not sure I want to go fix all the ILD scripts that have probably been unused for quite some time.

andresailer:
> @andresailer should we try formatting with longer lines to reduce a few of the line breaks? We could also configure a linter, but I am not sure how well that deals with all the dynamic imports from DD4hep and/or Gaudi. Plus, I am not sure I want to go fix all the ILD scripts that have probably been unused for quite some time.

Yes, longer line would probably be nice. Linter also, but agreed that it would be annoying to fix unused scripts from years gone by.

We do flake8 in DD4hep https://github.com/AIDASoft/DD4hep/blob/master/.github/workflows/python.yml so that seems to be not too bad, unless I missed a log of `ignore` statements in our files.

example/ddsim.py
andresailer, line 1: I think we should drop this file, because it might get confused with ddsim from DD4hep.

tmadlener: Done.

tmadlener:
I set the line length to 99, mainly because we have the same in podio and there is no other value yet in Key4hep. Diff seems to become slightly smaller, but I suppose the differences are rather minimal wrt default line length of 88.

andresailer (approved):
OK with 99, it is good at least to keep things consistent with podio.

Any objections by anyone?

.ruff.toml
```diff
@@ -0,0 +1,10 @@
+target-version = "py311"
```
andresailer, line 1:
```suggestion
target-version = "py310"
```

tmadlener: Since we have adopted `py311` in EDM4hep, I think we can also leave it here.

andresailer: Yes

tmadlener:
Looks like no one is unhappy enough to complain, so I would suggest to merge this.

PR #400, opened 2024-09-29
Improve check for a header file from DD4hep

jmcarcell:
BEGINRELEASENOTES
- Improve the check for DCH_info.h from DD4hep. It was added in version 1.29 so if the version is below that don't use it.

ENDRELEASENOTES

Also CMake hangs for a while on that check, maybe it's checking a lot of paths. Since DDRec is in the defaults and always built in the stack, I think it's fine not to check for this or if needed to check for DDRec. Eventually the list of files that depend on this header file will be wrong.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,6 +97,14 @@ if(NOT K4GEO_USE_LCIO)
   message(STATUS "Use of LCIO is DISABLED, some detectors that depend on LCIO will not be built: ${lcio_sources}")
 endif()
 
+if(${DD4hep_VERSION} VERSION_LESS 1.29)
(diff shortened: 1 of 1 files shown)


atolosadelgado:
@jmcarcell 

this check avoids compilation error when old versions of the stack are used, I would prefer keeping that check... 

The code was actually introduced by @andresailer , what do you think?

andresailer:
As long as we don't require a DD4hep version that includes this file we shouldn't drop the check. We can't check for DDRec because the file was introduced long after DDRec

jmcarcell:
Indeed, now this checks on the version of DD4hep where that header was introduced and if it doesn't exist it will not build the file.

PR #401, opened 2024-10-02
[IDEA] Fix overlaps with DR-Calo and between SiliconWrapper and endplate absorbers

mahmoudali2:
BEGINRELEASENOTES
- IDEA_o1_v03: Moving IDEA muon system starting point in both barrel (rmin) and endcap (z-offset) 30 mm, backwards to avoid overlapping with Dual readout calorimeter.
  - The change including also updating chamber names to distinguish between muon chambers and pre shower chamber, since they are using the same uURWELL chamber.
  - Change dimensions of solenoid/endplate and move the pre-shower to avoid overlaps with dual readout calorimeter

ENDRELEASENOTES
I made sure that the overlap check for the muon system is clean.


BrieucF:
Merging this as it is now to have IDEA free from overlaps (without the DR calo) in the Key4hep release, and prepare a more thorough discussion on how to best solve this for the next Full Sim meeting.

PR #404, opened 2024-10-17
Fix some material definitions for ILD models

danieljeans:
BEGINRELEASENOTES
- fix some inconsistent definitions of materials used in ILD simulation (sum of fractions != 1.000)
- adjust composition of RPC gas to R134a:0.93 + CO2:0.05 + SF6:0.02 [exchange 5% isoButane for CO2, as suggested by G. Grenier]

ENDRELEASENOTES


tmadlener:
Does this have any implications for the final physics?

danieljeans:
it should be completely negligible.

PR #407, opened 2024-11-04
make material in 1st layer of ECAL absorber configurable

giovannimarchiori:
Make material in 1st layer of noble-liquid ECAL absorber configurable.
Set by default to G10 rather than LAr. 
Also add more comments and fix some whitespaces

BEGINRELEASENOTES
- Make material in 1st layer of noble-liquid ECAL absorber configurable. Set by default to G10 rather than LAr. 
ENDRELEASENOTES

Here are a couple of plots that compare cellID and energy with the previous code and with the new code when the material in 1st layer is set to LAr via xml, to confirm that the code reshuffling itself has no impact on physics

ECalBarrelModuleThetaMergedPositioned cellID_comparison_old_vs_new
ECalBarrelModuleThetaMergedPositioned energy_comparison_old_vs_new

### DIFF
--- a/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/ECalBarrel_thetamodulemerged.xml
+++ b/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/ECalBarrel_thetamodulemerged.xml
@@ -126,11 +126,11 @@
         </active>
         <passive>
           <rotation angle="InclinationAngle"/>  <!-- inclination angle -->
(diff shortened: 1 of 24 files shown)


BrieucF:
Thanks Giovanni! You prefer to have the change to have all layers with the same "radial" extent in another PR?

giovannimarchiori:
Yes let's reserve that for another PR (are we sure we want to have both changes applied in the same detector model)?

PR #406, opened 2024-10-25
Segmented Crystal ECAL (SCEPCal) addition

wonyongc:
BEGINRELEASENOTES
- Added geometry, segmentation, sensitive action, and hit classes for the Segmented Crystal ECAL (SCEPCal).

ENDRELEASENOTES

Hi, I have been developing a full simulation for a segmented crystal ECAL (CalVision/MAXICC collaborations) and am now submitting PRs to get the simulation into key4hep.
I should note that I needed to make a new datatype to store the scintllation/cerenkov hit information for dual-readout. So far I have simply extended the edm4hep schema with a custom 'edm4dr' schema in the simulation repo, but to merge into k4geo, it seems like it would be a better option (or the only one) to have this new schema merged into the main edm4hep. I've opened a PR to do that here: https://github.com/key4hep/EDM4hep/pull/380
Please let me know if I should take a different approach or if you have suggestions or changes I should make.
Thanks,
Wonyong

PR #411, opened 2024-11-22
Include dual-readout endcap tubes calorimeter in IDEA_o2 xml file

lopezzot:
BEGINRELEASENOTES
- A new subdetector for the dual-readout capillary-tubes-based endcap calorimeter is included as discussed [here](https://indico.cern.ch/event/1439207/contributions/6056623/attachments/2903299/5092292/lopezzot_fccsim_2472024.pdf)
- As INFN recently changed the baseline of the IDEA detector, a new IDEA_o2 XML file is created accordingly. It includes the new `DREndcapTubes` subdetector, changes the solenoid inner radius and removes the preshower. XML files for IDEA_o2 dimensions and materials are included as well. IDEA_o2 will be completed with the new crystal em calorimeter and the dual-readout capillary-tubes-based barrel calorimeter.
- A new example `example/SteeringFile_IDEA_o2_v01.py` is included to run a simulation with the IDEA_o2 concept.
- A new SDAction (`DRTubesSDAction`) is included and applied to the `DREndcapTubes` subdetector via `regexSD` in `example/SteeringFile_IDEA_o2_v01.py`, this speeds up the event rate w.r.t. previous dual-readout calorimetry simulations and reduces the memory fooprint as explained [here](https://indico.cern.ch/event/1457476/contributions/6136107/attachments/2929865/5144647/lopezzot_fccsim_1892024.pdf)

ENDRELEASENOTES

Hello,

This PR should be the starting point for the new IDEA concept as proposed by INFN. The inner part of IDEA up to the Silicon Wrapper is not touched. The solenoid inner radius is increased and between the SiWrapper and the solenoid a new crystal dual-readout em calorimeter will be included. When PR  #406 is merged crystals will be included in the IDEA_o2 XML file. The preshower is removed. The hadronic endcap calorimeter is the new `DREndcapTubes` which uses the INFN Hidra-like technology. A new PR will be created with the [barrel hadronic calorimeter](https://github.com/s6anloes/DDDRCaloTubes). The muon system inner radius will have to be changed accordingly and for the moment it is commented-out in the IDEA_o2 XML file.

Testing:
- I produced 1000 10 GeV electrons events shot directly in the middle of the endcap calorimeter. They were simulated on lxplus9 at a rate of 0.8 s/evt with no warnings. Distributions from the calohits look good.
- Thanks to `regexSD` the memory footprint of this partial IDEA_o2 configuration is 4.5 Gb (previously it was 15 Gb).
- I run an overlap check with a 10 um tolerance and no overlaps were detected. However, checking for overlaps over 30M tubes takes forever, so the check was done without fibers inside the calorimeter towers and each standalone tower was checked for overlaps independently.

To do in the next PRs:
- In case this PR is merged the next steps for IDEA_o2 are to include the dual-readout em crystal calorimeter from #406 and the dual-readout capillary-tubes barrel calorimeter, to apply `DRTubesSDAction` to the barrel calorimeter and to adjust the muon system dimensions to avoid overlaps with the barrel calorimeter at theta=90 degree.

Please let me know if you agree with this strategy for IDEA option2. Thanks!


BrieucF:
Hi Lorenzo, great! 

Can you add a test similar to https://github.com/key4hep/k4geo/blob/main/test/CMakeLists.txt#L93 ? 

Since https://github.com/key4hep/k4geo/pull/406 is orthogonal, I think we can proceed with this PR regardless and add the crystals when ready.

lopezzot:
@BrieucF I added a test for IDEA_o2_v01.

I agree this PR can proceed regardless of the crystal one.

atolosadelgado:
Hi @lopezzot 

thanks for this PR, looks great!

I have only 2 questions:
* I understand that the SD action is setup in the steering file, but that file is not provided when running the test. Do you think is it worth it to use the steering file when running the test?
* I have run the simulation using different particles, and no tracking warnings are thrown. When plotting the hit pattern radius vs z, there is a pattern, is it ok? The picture below show the hit pattern in the endcap for kaons at 50GeV
DR_endcap_kaons

Thank you for your time.
Alvaro
PS: while running your geometry, I spotted a minor bug in DDSim, so double thanks.

BrieucF:
Can you also explain the hot spot at z ~2800? I naively would not expect that with 50 GeV K_longs...

FCCee/IDEA/compact/IDEA_o2_v01/DREndcapTubes_o1_v01.xml
```diff
@@ -0,0 +1,278 @@
...
+    </detector>
+  </detectors>
+
+  <readouts>
+    <readout name="DREndcapTubesRO">
+      <segmentation type="CartesianGridXY" 
+             grid_size_x="1*mm" 
+             grid_size_y="1*mm"/>
+      <!-- NO SPACES BETWEEN ID DESCRIPTORS!!! -->
+      <id>stave:10,tower:6,air:1,col:16,row:16,clad:1,core:1,cherenkov:1</id>
```
andresailer, line 274: This should start with `system:6` (or `:5`?) and maybe also `side:-2`?

Instead of air/clad/core/cherenkov with one bit, why not 2 bits (of they are exclusive), or 4 bits to encode the flag?

lopezzot:
Hello @atolosadelgado and @BrieucF, thank you for your time and very good questions.

1. Yes, I think it is worth using the steering file in the test. The only caveat is that regexSD takes a while to perform searches over so many volumes so the test time will increase. I will add a commit with this change and comment on the test time.

2. About hit positions (x,y,z). They are not the global position of the step inside a fiber but the position of the inner tip of the fiber where the step is taking place. If a track crosses the same fiber twice a single hit will be created with (x,y,z) equal to the the fiber inner tip and the signal of the corresponding fiber will be incremented twice. That is why there is a spot at 280 cm which is where most of fiber tips are (the endcap itself starts at 280 cm). The patterns Alvaro sees are due to the fact that new fibers (with shorter lengths) are included along the side of the tower (trapezoid) as soon as towers open up. Therefore some fibers have an inner tip not starting at 280 cm. These fibers are positioned in rows and columns around the tower at given depths long towers and this is the reason why some clusters appear in the scatter plot. In other words, not every fiber starts at 280 cm but those that don't are not randomly positioned, so some spots are visible even for them. This is my understanding for the moment and looks consistent with the geometry implemented (of course experience may prove me wrong).

plugins/DRTubesSDAction.cpp
```diff
@@ -0,0 +1,267 @@
...
+  // We recreate the TubeID from the tube copynumber:
+  // fist16 bits for the columnID and second 16 bits for the rowID
+  auto TubeID = aStep->GetPreStepPoint()->GetTouchable()->GetCopyNumber(2);
+  unsigned int ColumnID = TubeID >> 16;
+  unsigned int RawID = TubeID & 0xFFFF;
+  auto TowerID = static_cast<unsigned int>(aStep->GetPreStepPoint()->GetTouchable()->GetCopyNumber(3));
+  auto StaveID = static_cast<unsigned int>(aStep->GetPreStepPoint()->GetTouchable()->GetCopyNumber(4));
+
+  VolumeID VolID = 0; // recreate the 64-bit VolumeID
+  BitFieldCoder bc("stave:10,tower:6,air:1,col:16,row:16,clad:1,core:1,cherenkov:1");
```
andresailer, line 151: If this has to match with the readout `<id>` this should have a lot of comments in the XML

plugins/DRTubesSDAction.cpp
```diff
@@ -0,0 +1,267 @@
...
+  //
+  Geant4Calorimeter::Hit* hit = coll->findByKey<Geant4Calorimeter::Hit>(VolID);  // the hit
+  if (!hit) {  // if the hit does not exist yet, create it
+    hit = new Geant4Calorimeter::Hit();
+    hit->cellID = VolID;  // this should be assigned only once
+    G4ThreeVector FiberVec = DRTubesSglHpr::CalculateFiberPosition(aStep);
+    Position FiberPos(FiberVec.x(), FiberVec.y(), FiberVec.z());
+    hit->position = FiberPos;  // this should be assigned only once
+    // Note, when the hit is saved in edm4hep format the energyDeposit is
+    // divided by 1000, i.e. it translates from MeV (Geant4 unit) to GeV (DD4hep unit).
```
andresailer, line 243:
```suggestion
    // divided by 1000, i.e. it translates from MeV (Geant4 unit) to GeV (EDM4hep unit).
```

plugins/DRTubesSglHpr.hh
```diff
@@ -0,0 +1,178 @@
...
+    // Methods
+    //
+    // Apply Briks Law
+    static constexpr G4double ApplyBirks(const G4double& de, const G4double& steplength)
+    {
+      return (de / steplength) / (1 + fk_B * (de / steplength)) * steplength;
+    }
+
+    // Smear S signal according to Poissonian fluctuations and light yield
+    static G4int SmearSSignal(const G4double& satde) { return G4Poisson(satde * 9.5); }
```
andresailer, line 43: Where does the 9.5 come from?

plugins/DRTubesSglHpr.hh
```diff
@@ -0,0 +1,178 @@
...
+
+    // Smear C signal according to Poissonian fluctuations and light yield
+    static G4int SmearCSignal() { return G4Poisson(0.177); }
```
andresailer, line 46: Where does the 0.177 come from?

plugins/DRTubesSDAction.cpp
```diff
@@ -0,0 +1,267 @@
...
+  if (IsScin) {  // it is a scintillating fiber
+
+    //m_userData.fEvtAction->AddEdepScin(Edep);
+
+    if (aStep->GetTrack()->GetDefinition()->GetPDGCharge() == 0 || steplength == 0.) {
+      return true;  // not ionizing particle
+    }
+    G4double distance_to_sipm = DRTubesSglHpr::GetDistanceToSiPM(aStep);
+    signalhit =
+      DRTubesSglHpr::SmearSSignal(DRTubesSglHpr::ApplyBirks(Edep, steplength));
```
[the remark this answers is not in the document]
andresailer: Are you sure about the application to the scintillation process only, and not to the energy deposit? I am pretty sure that this affected the deposited energy in the past, and this is what CALICE scintillator test beam simulations used, and what went into the scintillator calo simulation in DD4hep as well...

https://github.com/AIDASoft/DD4hep/blob/c9023f116f4f161f21d244fccdbc971753bd748f/DDG4/src/Geant4StepHandler.cpp#L90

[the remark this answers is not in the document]
lopezzot: I am reviewing the Geant4 code for Birks Law right now and there is no trace of something that could affect the energy deposited. I also discussed with Geant4 colleagues, Briks Law in Geant4 can only affect the light emission process but not the energy deposited in the step. Of course surprises can happen, if you have evidence of something different I would be interested.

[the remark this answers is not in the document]
andresailer: This code in geant4 https://github.com/Geant4/geant4/blob/e58e650b32b961c8093f3dd6a2c3bc917b2552be/source/processes/electromagnetic/utils/src/G4EmSaturation.cc#L79 that is called from the step handler in DD4hep.

plugins/DRTubesSglHpr.hh
lopezzot: Added a comment to explain it. Thanks.

plugins/DRTubesSglHpr.hh
lopezzot: Added a comment to explain it. Thanks.

plugins/DRTubesSDAction.cpp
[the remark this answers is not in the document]
lopezzot: Ok I'm learning something here, unfortunately in Geant4 we have a lot of code that should not be exposed to users. If I understand it correctly, the Geant4StepHandler applies Birks Law to every step (or at least every time it is applied to a G4Step in the SDAction code). The point is, for almost every material (step) it will find that Birks Constant is 0 (default) and do absolutely nothing. It seems a very inefficient thing to do. Why not let the users define the Birks function in their sensitive detector code for their specific materials? This is the standard I encountered so far in LHC experiments simulations. Anyhow, in my case this is not a problem because in the geometry materials Briks Constants are never set and in my SDAction I do not use Geant4StepHandler... We could also have a chat in person about this.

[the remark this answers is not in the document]
andresailer: No, the step handler is only called by the Scintillator Calo SensitiveDetector.
https://github.com/AIDASoft/DD4hep/blob/c9023f116f4f161f21d244fccdbc971753bd748f/DDG4/plugins/Geant4SDActions.cpp#L445 So it will be applied selectively to steps in the sensitive material, and Birks constant should be defined for the sensitive material, otherwise a different sensitive detector could be chosen.

Why not re-use what is available in Geant4 for Applying Birks' Law though? Which is a more complete implementation than what you have?

But it is true that we don't apply Birks' Law twice accidentally because it comes via the sensitive detector code.

lopezzot:
Hi @andresailer, @atolosadelgado, @BrieucF while I was trying to add a "system" <id> for the DREndcapTubes subdetector I think I spotted a problem with DD4hep Assembly volumes. In my code, I use an assembly to place inside phi-slices volumes. I do this placement in a for loop over j and assign j as geant4 copynumber and dd4hep volID to the volumes placed inside the Assembly https://github.com/lopezzot/k4geo/blob/lp-idea_o2/detector/calorimeter/dual-readout-tubes/src/DREndcapTubes_o1_v01.cpp#L167.
However, when I get the Geant4 copynumber in the DRTubesSDAction code for these phi-slices volumes, the copynumber returned is not the one I set in the DD4hep geometry code. This reminds me of this problem https://github.com/aidasoft/DD4hep/pull/1277 (if i am not wrong) related to an LHCb issue with reproducibility of IDs. As far as I understand, while DD4hep converts the Assembly geometry node in geant4 node, copynumbers are internally changed and this is a problem because I need geant4 copynumbers to recreate the 64-bits volume IDs in my DRTubesSDAction code. What do you think? should I open an issue for this? thank you.

andresailer:
> should I open an issue for this? thank you.

Yes, please open an issue for this in DD4hep.

FCCee/IDEA/compact/IDEA_o2_v01/DREndcapTubes_o1_v01.xml
lopezzot: Right! I included a new `system:5` id and assigned it to the highest volume in the geometry node.
`side:-2` is not needed here because in the SDAction I check if the step is in endcap right (z>0) or left (z<0) and two hit collections (left and right) are used to store hits from the two endcaps, so there is no need for the `side` id in the hit-id.
About condensing 1-bit IDs into a larger id, it is certainly possible. However, in this implementation I have to use g4 copynumbers, retrieve them in the SDAction, and from them create the 64-bit id. If I condense more ids into a larger id, I have to include more functions in the geometry cpp code to encapsulate these short ids into a larger one and assign it to volumes as g4 copynumbers. The code will become longer and less readable both in the geometry part and in the SDAction part. If it is not a problem, I would prefer to keep this as simple as possible.

BrieucF:
> Hello @atolosadelgado and @BrieucF, thank you for your time and very good questions.
> 
>     1. Yes, I think it is worth using the steering file in the test. The only caveat is that regexSD takes a while to perform searches over so many volumes so the test time will increase. I will add a commit with this change and comment on the test time.
> 
>     2. About hit positions (x,y,z). They are not the global position of the step inside a fiber but the position of the inner tip of the fiber where the step is taking place. If a track crosses the same fiber twice a single hit will be created with (x,y,z) equal to the the fiber inner tip and the signal of the corresponding fiber will be incremented twice. That is why there is a spot at 280 cm which is where most of fiber tips are (the endcap itself starts at 280 cm). The patterns Alvaro sees are due to the fact that new fibers (with shorter lengths) are included along the side of the tower (trapezoid) as soon as towers open up. Therefore some fibers have an inner tip not starting at 280 cm. These fibers are positioned in rows and columns around the tower at given depths along towers and this is the reason why some clusters appear in the scatter plot. In other words, not every fiber starts at 280 cm but those that don't are not randomly positioned, so some spots are visible even for them. This is my understanding for the moment and looks consistent with the geometry implemented (of course experience may prove me wrong).
> 
> 
> EDIT: I added a commit below to use the steering file in the test. It takes about 350 s on lxplus9, I believe it is still acceptable.

Thanks for the explanation. Some more questions. Do you also store the single hits into CaloHitContributions? Do you set the position to the "real" hit position (instead of the fiber front) for those? Will this be enough to perform studies on possible longitudinal information from timing?

lopezzot:
> Thanks for the explanation. Some more questions. Do you also store the single hits into CaloHitContributions? Do you set the position to the "real" hit position (instead of the fiber front) for those? Will this be enough to perform studies on possible longitudinal information from timing?

Good point. For the moment I do not save calohitcontributions. Our experience with test-beam data, and standalone simulations of the 4pi geometry, show that having only the fibre tip position is already good enough for the vast majority of studies we want to do. Also note that in IDEA_o2 crystals will be positioned before this calo, and the need for longitudinal segmentation in the fiber-dr calorimeter is then reduced. In any case, for longitudinal studies it would certainly be possible to add calohitcontributions and save the true hit position and timing of each hit inside fibers. I believe a dedicated PR with some full-sim studies could be open just after this one.

BrieucF:
> > Thanks for the explanation. Some more questions. Do you also store the single hits into CaloHitContributions? Do you set the position to the "real" hit position (instead of the fiber front) for those? Will this be enough to perform studies on possible longitudinal information from timing?
> 
> Good point. For the moment I do not save calohitcontributions. Our experience with test-beam data, and standalone simulations of the 4pi geometry, show that having only the fibre tip position is already good enough for the vast majority of studies we want to do. Also note that in IDEA_o2 crystals will be positioned before this calo, and the need for longitudinal segmentation in the fiber-dr calorimeter is then reduced. In any case, for longitudinal studies it would certainly be possible to add calohitcontributions and save the true hit position and timing of each hit inside fibers. I believe a dedicated PR with some full-sim studies could be open just after this one.

Alright, for another PR then! Thanks

BrieucF:
Can you update the branch? Are we good to go?

lopezzot:
> Can you update the branch? Are we good to go?

This PR https://github.com/AIDASoft/DD4hep/pull/1362 fixes the problem I mentioned above. Therefore for what concerns my code, I believe we are good to go. Just two last questions:
1. Is it ok if I clang-format this code before merging? have you got a .clang-format file example in case or I can go with mine?
2. I have to rebase before merging, right?

BrieucF:
> > Can you update the branch? Are we good to go?
> 
> This PR [AIDASoft/DD4hep#1362](https://github.com/AIDASoft/DD4hep/pull/1362) fixes the problem I mentioned above. Therefore for what concerns my code, I believe we are good to go. Just two last questions:
> 
>     1. Is it ok if I clang-format this code before merging? have you got a .clang-format file example in case or I can go with mine?
> 
>     2. I have to rebase before merging, right?

1. Yes it is ok for me. You could potentially use this one: https://github.com/key4hep/k4FWCore/blob/main/.clang-format . I do not have a strong opinion, maybe @andresailer has ?
2. Yes, please rebase, I think only `test/CMakeList.txt` was touched both here and in the PR you are missing (https://github.com/key4hep/k4geo/pull/409/files#diff-33394812ba204689144fd2f80832db83853ba1cb32403edb4e15fe4893e675fd)

lopezzot:
Formatted and rebased.

lopezzot:
Just a small correction to the previous testing: the memory consumption is now around 2.5 Gb, that's because the tower granularity was increased from 36 to 72 rotations around the beam axis. Less unique tubes are now inserted inside a phi-slice and more unique tubes are replicated in phi-slices, therefore the memory footprint is reduced.
Can we proceed with merging?

BrieucF:
Yes, I was leaving some time for people to comment about the clang formatting. Looks like there is no complain.

Thanks Lorenzo!

FCCee/IDEA/compact/IDEA_o2_v01/DREndcapTubes_o1_v01.xml
BrieucF: @andresailer  can we resolve this?

andresailer: If you are convinced the two sides never end up in the same collection, yes.

lopezzot: Yes, here is the code that separates left and right hits in the correct collections. Thanks! https://github.com/key4hep/k4geo/pull/411/files#diff-ec6c4af548e901207b2fbad9d312c13896804af83afaa86aa7a27872fb03b625R191-R201

BrieucF: Ok, thanks. Resolving for now, we will see when using it in subsequent algorithms if it is annoying to have them in two separated collections. Indeed we usually have one collection for both endcaps.

PR #361, opened 2024-07-30
Update version 2, 3 and 4 of ECalBarrel_NobelLiquid_InclinedTrapezoid_o1_v0*_geo.cpp by including layered calorimeter data

SwathiSasikumar:
BEGINRELEASENOTES
- This updates version ECalBarrel_NobelLiquid_InclinedTrapezoid_o1_v02_geo.cpp and ECalBarrel_NobelLiquid_InclinedTrapezoid_o1_v03_geo.cpp with layered calorimeter data for PandoraPFA as given in ECalBarrel_NobelLiquid_InclinedTrapezoid_o1_v01_geo.cpp
- Update on the compact file LAr_ECalBarrel.xml based on the latest parameters and names...

ENDRELEASENOTES

PR #412, opened 2024-11-28
CLD_o4_v05: fixing LAr_ECalBarrel.xml

faltovaj:
BEGINRELEASENOTES

- CLD_o4_v05: Fixing LAr_ECalBarrel.xml constants name (was changed by mistake in #361) 

ENDRELEASENOTES

### DIFF
--- a/FCCee/CLD/compact/CLD_o4_v05/LAr_ECalBarrel.xml
+++ b/FCCee/CLD/compact/CLD_o4_v05/LAr_ECalBarrel.xml
@@ -71,7 +71,7 @@
     <constant name="Glue_thickness" value="0.1*mm"/>
     <!-- readout in between two absorber plates -->
     <constant name="readout_thickness" value="1.2*mm"/>
-    <constant name="ECal_cell_size" value="5.1*mm"/>
+    <constant name="ECal_cell_size" value="20*mm"/>
   </define>
 
   <display>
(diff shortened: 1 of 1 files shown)


SwathiSasikumar:
Thanks, Jana for correcting this. Two comments from my side: 1. ```ECal_cell_size``` should be 20 mm because 5 mm is the value for CALICE detectors. 2. ```ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v03``` was used on purpose because Brieuc wanted to update this XML file to the latest version. However, in the CLD+LAr version, ```ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v03``` doesn't seem like a good idea because this version requires the new 11-layered calorimeter instead of 12 layered calorimeter and breaks while running for CLD+LAr. So it is good to switch back to version 1.

BrieucF (approved):

SwathiSasikumar (approved):

jmcarcell:
I assume someone tested this because in its current state `CLD_o4_v05` fails when loaded by `DD4hep`.

atolosadelgado:
> I assume someone tested this. Merging; in its current state `CLD_o4_v05` fails when loaded by `DD4hep`.

What do you mean by "fails when loaded by DD4hep"? could you share a reproducer? 

I tried the following and it runs, which means the geometry is translated into Geant4:
```
source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh
git clone -b forPandora_v1 --depth 1 https://github.com/faltovaj/k4geo.git
cd k4geo/
cmake -B build -S . -D CMAKE_INSTALL_PREFIX=install
cmake --build build -- -j 68 install
k4_local_repo 
ddsim --compactFile ./FCCee/CLD/compact/CLD_o4_v05/CLD_o4_v05.xml -N 1 -G 
```

andresailer:
He means that the master branch fails and we need this PR to make it work again.

atolosadelgado:
> He means that the main branch fails and we need this PR to make it work again.

fair enough, thanks for the clarification :)

PR #416, opened 2024-12-16
DCH patch

atolosadelgado:
BEGINRELEASENOTES
- The drift chamber cell shape has been updated to a composite design, improving navigation speed in simulations by factor 10. To revert to the original twisted tube shape, add the optional `useG4TT` tag in the detector section.

ENDRELEASENOTES

Hi,

After the patch I did for Geant4 twisted tube, it seems there is still an error when shooting exactly at theta 90 degrees. While this bug is being fixed, I implemented an alternative shape. Original twisted tube shape can be used if optional tag is added within the detector section.

The change is transparent to the user, so I did not increase the version number of the detector, what do you think?

### DIFF
--- a/detector/tracker/DriftChamber_o1_v02.cpp
+++ b/detector/tracker/DriftChamber_o1_v02.cpp
@@ -578,7 +578,7 @@ dd4hep::Solid CompositeTT(double twist_angle,  double cell_rin_z0,  double cell_
   // ROOT hyperboloid require stereoangles stin and stout
   //-- stereo for rmin
   double stin  = DCH_i.stereoangle_z0(cell_rin_z0);
-  //-- stereo for rour
+  //-- stereo for rout
   double stout = DCH_i.stereoangle_z0(cell_rout_z0);
 
   // make hyperboloid longer, later intersection with gtrap will lead to proper length
--- a/detector/tracker/DriftChamber_o1_v02.cpp
+++ b/detector/tracker/DriftChamber_o1_v02.cpp
@@ -25,6 +25,8 @@ using DCH_length_t = dd4hep::rec::DCH_info_struct::DCH_length_t;
 using DCH_angle_t  = dd4hep::rec::DCH_info_struct::DCH_angle_t;
 using DCH_layer    = dd4hep::rec::DCH_info_struct::DCH_layer;
 
+dd4hep::Solid CompositeTT(double twist_angle,  double cell_rin_z0,  double cell_rout_z0, double dz, double dphi, const dd4hep::rec::DCH_info & DCH_i);
+
 /// Function to build DCH
 static dd4hep::Ref_t create_DCH_o1_v02(dd4hep::Detector &desc, dd4hep::xml::Handle_t handle, dd4hep::SensitiveDetector sens)
 {
@@ -75,6 +77,7 @@ static dd4hep::Ref_t create_DCH_o1_v02(dd4hep::Detector &desc, dd4hep::xml::Hand
     }
 
     bool debugGeometry = detElem.hasChild(_Unicode(debugGeometry));
+    bool useG4TT = detElem.hasChild(_Unicode(useG4TT));
     auto gasElem    = detElem.child("gas");
     auto gasvolMat  = desc.material(gasElem.attr<std::string>(_Unicode(material)));
     auto gasvolVis  = desc.visAttributes(gasElem.attr<std::string>(_Unicode(vis)));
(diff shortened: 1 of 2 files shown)


andresailer:
Which "optional" tag should one use?

detector/tracker/DriftChamber_o1_v02.cpp
```diff
@@ -285,7 +288,9 @@ static dd4hep::Ref_t create_DCH_o1_v02(dd4hep::Detector &desc, dd4hep::xml::Hand
         DCH_length_t cell_rout_zLhalf  = DCH_i->Radius_zLhalf(cell_rout_z0);
         DCH_length_t cell_dz           = DCH_i->Lhalf;
         DCH_angle_t cell_phi_width     = phi_step - safety_phi_interspace;
-        dd4hep::TwistedTube cell_s( cell_twistangle, cell_rin_zLhalf, cell_rout_zLhalf, cell_dz, 1, cell_phi_width);
+        dd4hep::Solid cell_s;
+        if( useG4TT ) cell_s = dd4hep::TwistedTube ( cell_twistangle, cell_rin_zLhalf, cell_rout_zLhalf, cell_dz, 1, cell_phi_width);
+        else          cell_s = CompositeTT(cell_twistangle, cell_rin_z0, cell_rout_z0, cell_dz, cell_phi_width, *DCH_i);
```
andresailer, line 293:
```suggestion
        dd4hep::Solid cell_s = useG4TT ?
                               dd4hep::TwistedTube(cell_twistangle, cell_rin_zLhalf, cell_rout_zLhalf, cell_dz, 1, cell_phi_width):
                               CompositeTT(cell_twistangle, cell_rin_z0, cell_rout_z0, cell_dz, cell_phi_width, *DCH_i);
```

detector/tracker/DriftChamber_o1_v02.cpp
```diff
@@ -510,6 +515,81 @@ static dd4hep::Ref_t create_DCH_o1_v02(dd4hep::Detector &desc, dd4hep::xml::Hand
...
+                                          fZneg.D.x, fZneg.D.y
+                                        };
+
+  dd4hep::EightPointSolid gtrap_shape(dz, vertices_array.data() );
+
+  //----------------- Hyperboloid -------------
+  // ROOT hyperboloid require stereoangles stin and stout
+  //-- stereo for rmin
+  double stin  = DCH_i.stereoangle_z0(cell_rin_z0);
+  //-- stereo for rour
```
andresailer, line 581:
```suggestion
  //-- stereo for rout
```

detector/tracker/DriftChamber_o1_v02.cpp
```diff
@@ -510,6 +515,81 @@ static dd4hep::Ref_t create_DCH_o1_v02(dd4hep::Detector &desc, dd4hep::xml::Hand
...
+    point2d A;
+    point2d B;
+    point2d C;
+    point2d D;
+  };
+
+  face fZneg;
+  face fZpos;
+
+  fZpos.A = { trap_rin*cos(  poly_angle + twist_angle_half ), trap_rin*sin(  poly_angle + twist_angle_half ) };
```
andresailer, line 553: can we get some general comments what this is doing? Why it is doing it?

detector/tracker/DriftChamber_o1_v02.cpp
```diff
@@ -75,6 +77,7 @@ static dd4hep::Ref_t create_DCH_o1_v02(dd4hep::Detector &desc, dd4hep::xml::Hand
     }
 
     bool debugGeometry = detElem.hasChild(_Unicode(debugGeometry));
+    bool useG4TT = detElem.hasChild(_Unicode(useG4TT));
```
lopezzot, line 80: Hi, do you think it is worth adding a comment in the geometry section explaining what this does and how to set it? or better is users do not try it out?

atolosadelgado: it is better users do not try

detector/tracker/DriftChamber_o1_v02.cpp
atolosadelgado: For the sake of readability, I only use a ternary if it fits into one 80-char line, which is not the case

detector/tracker/DriftChamber_o1_v02.cpp
atolosadelgado: thanks, I missed that one!

detector/tracker/DriftChamber_o1_v02.cpp
atolosadelgado: I have added the following comments, is it enough? I think it is challenging to explain without a picture
```
/// Solid equivalent to a twisted tube, resulting from the intersection of an hyperboloid and a generic trapezoid
/// the hyperboloid provides the hyperboloidal surfaces, the trapezoid provides the other two types of surfaces
```

atolosadelgado: I am writing a technical note with plots and equations, I think it is better suited place to explain what is happening in that piece of code. I can add later a reference to the note once it is uploaded to CDS

PR #418, opened 2025-01-13
Avoid potential use of dangling temporary objects.

scott-snyder:
In something like

```
  const auto& transformMatrix = detelement.nominal().worldTransformation();
```
nominal() returns a temporary Alignment object by value, and worldTransformation() returns a reference within that Alignment object. So we should make a copy.
(The C++ rules for extending the lifetime of a temporary bound to a reference don't help here because the object bound to a reference is different from the object whose lifetime needs to be extended.)

BEGINRELEASENOTES
- Fix potential uses of dangling temporaries.
ENDRELEASENOTES


jmcarcell:
Newer versions of clang report that, but this is only available in the LCG stacks I think

tmadlener:
k4geo should build on an LCG stack, right? It would be nice to avoid re-introducing this in the future.

Issue #421, opened 2025-01-21
problems using MDI_o1_v01


danieljeans:
### Check duplicate issues.

- [x] Checked for duplicates

### Goal

Hi,

I'm having problems using the CAD-based models of the FCCee MDI under MDI_o1_v01 .

I get an error like:

> Compact          ERROR ++ FAILED    to convert subdetector: Beampipe_STL: Cannot find a material referenced by name:DefaultMaterial

It seems the specified materials are not being picked up correctly, and being replaced by "DefaultMaterial"?
I found the [discussion](https://github.com/AIDASoft/DD4hep/issues/1076) , as far as I can see from that discussion, the current code follows the recommendations.

I believe that I could use these models several months ago (~September 2024), but I'm afraid I cannot remember in exactly which setup...

Thanks for your help,
Cheers,
Daniel.

### Operating System and Version

RH9

### compiler

gcc11.4.1

### The version of the key4hep stack

/cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh -r 2025-01-20

### Package Version

latest k4geo and/or dd4hep 

### Reproducer

A minimal example:

```
$ source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh -r 2025-01-20

$ wget https://fccsw.web.cern.ch/fccsw/filesForSimDigiReco/MDI/MDI_o1_v01/stl_files/Pipe_240430/AlBeMet162_30042024.stl

$ cat << EOF > gg.xml
<lccdd>
  <includes>
    <gdmlFile  ref="\${K4GEO}/FCCee/CLD/compact/CLD_o4_v05/elements.xml"/>
    <gdmlFile  ref="\${K4GEO}/FCCee/CLD/compact/CLD_o4_v05/materials.xml"/>
  </includes>
  <define>
    <constant name="world_side"             value="30*m"/>
    <constant name="world_x"                value="world_side/2"/>
    <constant name="world_y"                value="world_side/2"/>
    <constant name="world_z"                value="world_side/2"/>
  </define>
  <detectors>
    <detector name="Beampipe_STL" type="DD4hep_TestShape_Creator">
      <check>
        <shape type="CAD_MultiVolume" ref="AlBeMet162_30042024.stl"
               unit="mm" material="AlBeMet162">
          <volume id="0" name="s1" vis="AlBeMet_vis" material="AlBeMet162">
            <position x="0*cm"  y="0*cm"  z="0*cm"/>
[...]

PR #422, opened 2025-01-21
add Default material to CLD o4 v5 material file

atolosadelgado:
This solves https://github.com/key4hep/k4geo/issues/421

BEGINRELEASENOTES
- Add Default material to CLD o4 v5 material file, so it can be used with CAD models. 
- Add simulation test of 1 event for CLD o4 v5

ENDRELEASENOTES


jmcarcell:
What happens if there is no default material defined?

atolosadelgado:
> What happens if there is no default material defined?

DDCad requires a material named "Default material" to work, even if it is not used. In our case, it prevents the CAD-based MDI description to be used by CLD.

jmcarcell (approved):

PR #432, opened 2025-02-14
Pandora plugin to calculate material properties of calorimeters dynamically

SwathiSasikumar:
BEGINRELEASENOTES
- PandoraPFA uses ```LayeredCalorimeterData``` class from DD4hep to store layer-by-layer information e.g radiation length., essential for shower reconstruction and energy measurement. 
- In this plugin,  using another class from DD4hep, MaterialManager, LayeredCalorimeterData object is dynamically built based on the detector geometry ensuring each calorimeter layer correctly reflects its material composition. 
- This plugin can be used for any calorimeter irrespective of its geometry. It is essential to have LayeredCalorimeterData information to run Pandora.
- The changes in the geometry source file ```ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v01_geo.cpp``` and xml file ```CLD_o4_v05.xml``` should also be noted to accommodate the plugin. 
ENDRELEASENOTES

### DIFF
--- a/FCCee/CLD/compact/CLD_o4_v05/LAr_ECalBarrel.xml
+++ b/FCCee/CLD/compact/CLD_o4_v05/LAr_ECalBarrel.xml
@@ -82,15 +82,15 @@
     <!-- readout for the simulation -->
     <!-- offset in eta is eta max value including cryostat -->
     <readout name="ECalBarrelEta">
-        <!-- segmentation type="GridTheta_k4geo" grid_size_theta="0.5625" offset_theta="-0.83"/ -->
-        <segmentation type="GridEta_k4geo" grid_size_eta="0.01" offset_eta="-1.2"/>
+      <!-- segmentation type="GridTheta_k4geo" grid_size_theta="0.5625" offset_theta="-0.83"/ -->
+      <segmentation type="GridEta_k4geo" grid_size_eta="0.01" offset_eta="-1.2"/>
       <id>system:5,cryo:1,type:3,subtype:3,layer:8,module:11,eta:9</id>
     </readout>
     <!-- readout for the reconstruction -->
     <!-- phi position is calculated based on the centre of volume (hence it cannot be done in the simulation from energy deposits position) -->
     <readout name="ECalBarrelPhiEta">
-        <!-- segmentation type="FCCSWGridPhiTheta" grid_size_theta="0.5625" phi_bins="768" offset_theta="-0.83" offset_phi="-pi+(pi/768.)"/ -->
-        <segmentation type="FCCSWGridPhiEta_k4geo" grid_size_eta="0.01" phi_bins="768" offset_eta="-1.2" offset_phi="-pi+(pi/768.)"/>
+      <!-- segmentation type="FCCSWGridPhiTheta" grid_size_theta="0.5625" phi_bins="768" offset_theta="-0.83" offset_phi="-pi+(pi/768.)"/ -->
+      <segmentation type="FCCSWGridPhiEta_k4geo" grid_size_eta="0.01" phi_bins="768" offset_eta="-1.2" offset_phi="-pi+(pi/768.)"/>
       <id>system:5,cryo:1,type:3,subtype:3,layer:8,eta:9,phi:10</id>
     </readout>
   </readouts>
--- a/detector/calorimeter/ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v01_geo.cpp
+++ b/detector/calorimeter/ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v01_geo.cpp
@@ -559,6 +559,7 @@ static dd4hep::detail::Ref_t createECalBarrelInclined(dd4hep::Detector& aLcdd,
   dd4hep::PlacedVolume envelopePhysVol = motherVol.placeVolume(envelopeVol);
   envelopePhysVol.addPhysVolID("system", xmlDetElem.id());
   caloDetElem.setPlacement(envelopePhysVol);
+  dd4hep::xml::setDetectorTypeFlag(xmlDetElem, caloDetElem);
 
  /* // Create caloData object
   auto caloData = new dd4hep::rec::LayeredCalorimeterData;
--- a/detector/calorimeter/ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v01_geo.cpp
+++ b/detector/calorimeter/ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v01_geo.cpp
@@ -559,7 +559,6 @@ static dd4hep::detail::Ref_t createECalBarrelInclined(dd4hep::Detector& aLcdd,
   dd4hep::PlacedVolume envelopePhysVol = motherVol.placeVolume(envelopeVol);
   envelopePhysVol.addPhysVolID("system", xmlDetElem.id());
   caloDetElem.setPlacement(envelopePhysVol);
-  dd4hep::xml::setDetectorTypeFlag(xmlDetElem, caloDetElem);
 
  /* // Create caloData object
   auto caloData = new dd4hep::rec::LayeredCalorimeterData;
--- a/plugins/Pandora_LayeredCalorimeterPlugin.cpp
+++ b/plugins/Pandora_LayeredCalorimeterPlugin.cpp
@@ -1,3 +1,34 @@
+/*
+ * Copyright (c) 2020-2024 Key4hep-Project.
+ *
+ * This file is part of Key4hep.
+ * See https://key4hep.github.io/key4hep-doc/ for further info.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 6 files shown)


FCCee/CLD/compact/CLD_o4_v05/LAr_ECalBarrel.xml
```diff
@@ -82,15 +82,15 @@
...
+      <!-- segmentation type="GridTheta_k4geo" grid_size_theta="0.5625" offset_theta="-0.83"/ -->
+      <segmentation type="GridEta_k4geo" grid_size_eta="0.01" offset_eta="-1.2"/>
       <id>system:5,cryo:1,type:3,subtype:3,layer:8,module:11,eta:9</id>
     </readout>
     <!-- readout for the reconstruction -->
     <!-- phi position is calculated based on the centre of volume (hence it cannot be done in the simulation from energy deposits position) -->
     <readout name="ECalBarrelPhiEta">
-        <!-- segmentation type="FCCSWGridPhiTheta_k4geo" grid_size_theta="0.5625" phi_bins="768" offset_theta="-0.83" offset_phi="-pi+(pi/768.)"/ -->
-        <segmentation type="FCCSWGridPhiEta_k4geo" grid_size_eta="0.01" phi_bins="768" offset_eta="-1.2" offset_phi="-pi+(pi/768.)"/>
+      <!-- segmentation type="FCCSWGridPhiTheta" grid_size_theta="0.5625" phi_bins="768" offset_theta="-0.83" offset_phi="-pi+(pi/768.)"/ -->
```
atolosadelgado, line 92: please, forgive my ignorance, but I have to ask: is this change of segmentation required to run Pandora?

detector/calorimeter/ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v01_geo.cpp
```diff
@@ -560,7 +560,7 @@ static dd4hep::detail::Ref_t createECalBarrelInclined(dd4hep::Detector& aLcdd,
   envelopePhysVol.addPhysVolID("system", xmlDetElem.id());
   caloDetElem.setPlacement(envelopePhysVol);
 
-  // Create caloData object
+ /* // Create caloData object
```
atolosadelgado, line 563: I would suggest unstaging this file from the PR. Please, see below the option of adding a few lines to the plugin so that it can remove any assigned data extension

plugins/Pandora_LayeredCalorimeterPlugin.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    // Create LayeredCalorimeterData
+    auto caloData = new dd4hep::rec::LayeredCalorimeterData;
+    caloData->layoutType = dd4hep::rec::LayeredCalorimeterData::BarrelLayout;
+    caloData->extent[0] = Rmin;
+    caloData->extent[1] = Rmax;
+    caloData->extent[2] = 0.0; // Barrel-specific value
+    caloData->extent[3] = half_length;
+
+    caloDetElem.addExtension<dd4hep::rec::LayeredCalorimeterData>(caloData);
```
atolosadelgado, line 101: could you move this line after the print out "Successfully added LayeredCalorimeterData", at the end of the function?

In addition, to be discussed... do you think it is acceptable to remove any data extension attached by the detector constructor itself? To do it:
```
    if( ! caloDetElem->extensions.empty()) {
      dd4hep::printout(PrintLevel::WARNING, LOG_SOURCE, "Removing data extensions from %s", subDetectorName.c_str());
      caloDetElem->extensions.clear();
    }
```
This also requires an additional header:
```
#include <DD4hep/detail/DetectorInterna.h>
```

This way, the plugin becomes more robust against potential cases where other data extension is hardcoded in the detector constructor

plugins/Pandora_LayeredCalorimeterPlugin.cpp
```diff
@@ -0,0 +1,166 @@
...
+    double dR =  Rmax - Rmin;
+    double scale_fact = dR / (-Rmin * cos(angle) + sqrt(pow(Rmax, 2) - pow(Rmin * sin(angle), 2)));
+
+    // Loop through provided layer heights
+    for (double layerHeight : layerHeights) {
+      rad_last = rad_first + (layerHeight * scale_fact);
+      dd4hep::rec::Vector3D ivr1 = dd4hep::rec::Vector3D(0., rad_first, 0); // defining starting vector points of the given layer
+      dd4hep::rec::Vector3D ivr2 = dd4hep::rec::Vector3D(0., rad_last, 0);  // defining end vector points of the given layer
+     
+      const dd4hep::rec::MaterialVec &materials = matMgr.materialsBetween(ivr1, ivr2); // calling material manager(MM) to get material info between two points
```
atolosadelgado, line 116: do you think it would be good idea to protect against empty vectors?
```
// skip layer if no materials found
if( materials.empty() ) continue;
```

atolosadelgado (commented):
Hi @SwathiSasikumar 

Thank you for your PR! It is well-written.

I have one question: Would it be possible or advisable for the plugin to erase any previous data extension? If so, this can be done by adding a few lines I mentioned in the comments, and the detector constructor can be unstaged from the PR

Thank you for your time!

plugins/Pandora_LayeredCalorimeterPlugin.cpp
andresailer: Wouldn't this cause a memory leak?
```
      caloDetElem->extensions.clear();
```

andresailer: > ... do you think it is acceptable/advisable to remove any data extension attached 

Actually, it is not advisable to remove all extensions. It might make sense to remove the extension of the type we add here. It might also make sense to throw an error and let the user figure out which extension they want to keep. Maybe add a flag to replace or not an existing extension.

andresailer: We need to add a convenience function in dd4hep to make it easier to remove specific extensions. This issue brought up here will be addressed in a later PR then.

plugins/Pandora_LayeredCalorimeterPlugin.cpp
andresailer: I think unless `rad_first == rad_last` it is impossible for there to be no materials between two points? There must at least be the world volume. Or something is more wrong in the geometry than we can protect against here.

plugins/Pandora_LayeredCalorimeterPlugin.cpp
```diff
@@ -0,0 +1,166 @@
...
+    // Material manager
+    rec::MaterialManager matMgr(caloDetElem.volume());
+    double rad_first = Rmin;
+    double rad_last = 0;
+    double dR =  Rmax - Rmin;
+    double scale_fact = dR / (-Rmin * cos(angle) + sqrt(pow(Rmax, 2) - pow(Rmin * sin(angle), 2)));
+
+    // Loop through provided layer heights
+    for (double layerHeight : layerHeights) {
+      rad_last = rad_first + (layerHeight * scale_fact);
```
giovannimarchiori, line 112: if this is to be applied also to the ecal with inclined electrodes, please note that these formulae are not correct, as the relation is not linear between layer length along the electrode and deltaR. The correct way handling these is implemented in https://github.com/key4hep/k4geo/blob/0a6afaea82216144d2c558c684cbc59579b56e29/detector/calorimeter/ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v03_geo.cpp#L703-L711

SwathiSasikumar: The plugin is a general tool and any detector irrespective of its geometry should be able to use it.  The formulas are completely correct. The only difference is that the layer heights provided inside the xml file as the plugin arguments should be given radially and not along the electrode.

giovannimarchiori: Hi @SwathiSasikumar ,

ok, I guess that with "the layer heights should be given radially" you mean not the radial depth of each layer (otherwise you would not need the scale_fact) but the electrode segmentation, rescaled to the total radial thickness of the active ECAL volume, as was done in ECalBarrel_NobleLiquid_InclinedTrapezoids_o1_v01/v02_geo.cpp 

I would suggest to at least make this very clear in the code (add a comment to state that the layer heights are the electrode segmentation, rescaled to the total radial thickness of the active ECAL volume (not the radial depth)).

SwathiSasikumar: @giovannimarchiori there is an update to this PR which is due in next days. The layer heights will not need any scaling factor and will also include some other changes. I am waiting to correct an overlap in the detector geometry.
In the update there already is a comment on the XML file which says that the layer height needs to be added radially.

jmcarcell:
This PR has been stale and has now conflicts. I'm closing, if it's needed we can reopen.

PR #434, opened 2025-02-14
Add LANGUAGES CXX to CMakeLists.txt

jmcarcell:
BEGINRELEASENOTES
- Add LANGUAGES CXX to CMakeLists.txt to disable checks for a C compiler

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 set( PackageName k4geo )
 #---------------------------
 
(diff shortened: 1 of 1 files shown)


tmadlener:
Just for completness, and since there was some discussion on this on [spack]https://github.com/spack/spack/pull/48963#issuecomment-2648199643), we are not running into any problems with this downstream?

jmcarcell:
If we don't compile C code then the check is not needed, something else is that this change may not do anything if something else makes this check happen anyway which I haven't checked... But this should be the correct setting for k4geo and others.

PR #440, opened 2025-03-12
Modifying CTests for IDEA & Allegro to not to shoot particles to (0,0,1)

SanghyunKo:
BEGINRELEASENOTES
- Modified CTests for IDEA & Allegro to not to shoot particles to (0,0,1)

ENDRELEASENOTES

Currently, some CTests run with a default configuration (shooting `mu-` to the direction `(0,0,1)`), which effectively nullifies the test. Instead, the tests for IDEA and Allegro should now shoot many Geantinos (for testing the geometry navigation) or a handful of other particles (for testing the simulation).

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -124,7 +124,7 @@ endif()
 if(DCH_INFO_H_EXIST)
 SET( test_name "test_IDEA_with_DRC_o1_v03" )
 ADD_TEST( t_${test_name} "${CMAKE_INSTALL_PREFIX}/bin/run_test_${PackageName}.sh"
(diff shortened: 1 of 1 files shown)


atolosadelgado:
Hi @SanghyunKo 

Thanks for this PR! Out of curiosity, the test for the DRC with electrons, do you think it is worth it to force the particle type in the ctest command (by adding `--gun.particle electron`) to ensure the comment and the behavior remain consisten?

Otherwise, it looks fine for me.

SanghyunKo:
@atolosadelgado No problem, I've just added the line (it was my laziness for not checking that the particle name should be `e-`)

PR #443, opened 2025-03-18
SCEPCal addition

wonyongc:
SCEPCal addition

BEGINRELEASENOTES
- Added Segmented Crystal EM Precision Calorimeter (SCEPCal)
- Separate Main/Timing layer subdetectors
- Separate readout collections for edep, S/C photon counts using Geant4CalorimeterHit/edm4hep::SimCalorimeterHit
- Support different F/R crystal granularities in main layer
- Single, pure-projective timing layer
- New main layer projective crack migitation scheme with angular offset
- Added OpticalSurfaces
ENDRELEASENOTES

### DIFF
--- a/example/SteeringFile_IDEA_o2_v01.py
+++ b/example/SteeringFile_IDEA_o2_v01.py
@@ -603,31 +603,33 @@ def setupCerenkov(kernel):
     from DDG4 import PhysicsList
 
     seq = kernel.physicsList()
-    cerenkov = PhysicsList(kernel, 'Geant4CerenkovPhysics/CerenkovPhys')
+    cerenkov = PhysicsList(kernel, "Geant4CerenkovPhysics/CerenkovPhys")
     cerenkov.MaxNumPhotonsPerStep = 1000
     # cerenkov.MaxBetaChangePerStep = 10.0
     # cerenkov.TrackSecondariesFirst = True
     cerenkov.VerboseLevel = 0
     cerenkov.enableUI()
     seq.adopt(cerenkov)
-    ph = PhysicsList(kernel, 'Geant4OpticalPhotonPhysics/OpticalGammaPhys')
-    ph.addParticleConstructor('G4OpticalPhoton')
+    ph = PhysicsList(kernel, "Geant4OpticalPhotonPhysics/OpticalGammaPhys")
+    ph.addParticleConstructor("G4OpticalPhoton")
     ph.VerboseLevel = 0
     ph.enableUI()
     seq.adopt(ph)
     return None
 
+
 def setupCerenkovScint(kernel):
     from DDG4 import PhysicsList
+
     seq = kernel.physicsList()
 
-    scint = PhysicsList(kernel, 'Geant4ScintillationPhysics/ScintillationPhys')
+    scint = PhysicsList(kernel, "Geant4ScintillationPhysics/ScintillationPhys")
     scint.VerboseLevel = 0
     scint.TrackSecondariesFirst = True
     scint.enableUI()
     seq.adopt(scint)
 
-    cerenkov = PhysicsList(kernel, 'Geant4CerenkovPhysics/CerenkovPhys')
+    cerenkov = PhysicsList(kernel, "Geant4CerenkovPhysics/CerenkovPhys")
     cerenkov.VerboseLevel = 0
     cerenkov.MaxNumPhotonsPerStep = 10
     cerenkov.MaxBetaChangePerStep = 10.0
@@ -635,13 +637,14 @@ def setupCerenkovScint(kernel):
     cerenkov.enableUI()
     seq.adopt(cerenkov)
 
-    ph = PhysicsList(kernel, 'Geant4OpticalPhotonPhysics/OpticalGammaPhys')
-    ph.addParticleConstructor('G4OpticalPhoton')
+    ph = PhysicsList(kernel, "Geant4OpticalPhotonPhysics/OpticalGammaPhys")
+    ph.addParticleConstructor("G4OpticalPhoton")
     ph.VerboseLevel = 0
     ph.enableUI()
     seq.adopt(ph)
 
     return None
+
 
 SIM.physics.setupUserPhysics(setupCerenkov)
 


detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -0,0 +1,690 @@
...
+    xml_comp_t compXml,
+    dd4hep::Volume assemblyVol,
+    dd4hep::PlacedVolume *thePlacedVol=nullptr
+  ) {
+      dd4hep::Polyhedra theShape(nSegments,phiOffset,phiExtent,zPoints,rMin,rMax);
+      dd4hep::Volume theVolume(volName, theShape, theDetector.material("Vacuum"));
+      theVolume.setVisAttributes(theDetector,compXml.visStr());
+      *thePlacedVol=assemblyVol.placeVolume(theVolume);
+      return theVolume;
+  };
```
SanghyunKo, line 216: I'm not sure how `dd4hep::Volume` will behave in the `return`... Is this lambda function really necessary?

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -0,0 +1,690 @@
...
+      double rE=r0e+(XTAL_LEN_F+XTAL_LEN_R)/2.;
+      RotationZYX rotE(M_PI/2,thC,0);
+      Position    dispE(rE*sin(thC)-PROJ_OFFSET_R,0,rE*cos(thC));
+      
+      dd4hep::EightPointSolid barrelThetaAssemblyShape((XTAL_LEN_F+XTAL_LEN_R)/2, verticesE);
+      dd4hep::Volume barrelThetaAssemblyVolume("barrelThetaAssembly", barrelThetaAssemblyShape, theDetector.material("Vacuum"));
+      barrelThetaAssemblyVolume.setVisAttributes(theDetector,barrelAssemblyThetaVisXML.visStr());
+      barrelPhiAssemblyVolume.placeVolume(barrelThetaAssemblyVolume,Transform3D(rotE,dispE));
+
+      for (int nGamma=0; nGamma<N_GAMMA_BARREL; nGamma++) {
```
SanghyunKo, line 395: What is `Gamma`?? (need comments)

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -0,0 +1,690 @@
...
+
+      dd4hep::Volume endcapThetaAssemblyVolume("endcapThetaAssembly", endcapThetaAssemblyShape, theDetector.material("Vacuum"));
+      endcapThetaAssemblyVolume.setVisAttributes(theDetector,endcapAssemblyThetaVisXML.visStr());
+      endcapPhiAssemblyVolume.placeVolume(endcapThetaAssemblyVolume,Transform3D(rotE,dispE));      
+
+      dd4hep::Volume endcapThetaAssemblyVolume_1("endcapThetaAssembly_1", endcapThetaAssemblyShape, theDetector.material("Vacuum"));
+      endcapThetaAssemblyVolume_1.setVisAttributes(theDetector,endcapAssemblyThetaVisXML.visStr());
+      endcapPhiAssemblyVolume_1.placeVolume(endcapThetaAssemblyVolume_1,Transform3D(rotE,dispE));   
+
+      for (int nGamma=0;nGamma<nGammaEndcap;nGamma++) {
```
SanghyunKo, line 565: Same (need comments)

detector/calorimeter/SCEPCal_TimingLayer.cpp
```diff
@@ -0,0 +1,495 @@
...
+    xml_comp_t compXml,
+    dd4hep::Volume assemblyVol,
+    dd4hep::PlacedVolume *thePlacedVol=nullptr
+  ) {
+      dd4hep::Polyhedra theShape(nSegments,phiOffset,phiExtent,zPoints,rMin,rMax);
+      dd4hep::Volume theVolume(volName, theShape, theDetector.material("Vacuum"));
+      theVolume.setVisAttributes(theDetector,compXml.visStr());
+      *thePlacedVol=assemblyVol.placeVolume(theVolume);
+      return theVolume;
+  };
```
SanghyunKo, line 231: Same comments with the main layer

detectorSegmentations/include/detectorSegmentations/SCEPCal_MainSegmentationHandle_k4geo.h
```diff
@@ -0,0 +1,72 @@
...
+
+    inline CellID setCellID(int System, int Phi, int Eta, int Gamma, int Epsilon, int Depth) const {
+        return access()->implementation->setCellID(System, Phi, Eta, Gamma, Epsilon, Depth);
+    }
+
+    inline int System(const CellID& aCellID) const { return access()->implementation->System(aCellID); }
+    inline int Phi(const CellID& aCellID) const { return access()->implementation->Phi(aCellID); }
+    inline int Eta(const CellID& aCellID) const { return access()->implementation->Eta(aCellID); }
+    inline int Gamma(const CellID& aCellID) const { return access()->implementation->Gamma(aCellID); }
+    inline int Epsilon(const CellID& aCellID) const { return access()->implementation->Epsilon(aCellID); }
```
SanghyunKo, line 54: Also needs some comments what are `Gamma` and `Epsilon`.

detectorSegmentations/include/detectorSegmentations/SCEPCal_TimingSegmentation_k4geo.h
```diff
@@ -0,0 +1,113 @@
...
+#include "DDSegmentation/Segmentation.h"
+#include "Math/Vector3D.h"
+#include "DD4hep/DetFactoryHelper.h"
+#include <vector>
+#include <cmath>
+
+namespace dd4hep {
+namespace DDSegmentation {
+
+class SCEPCal_TimingSegmentation_k4geo : public Segmentation {
```
SanghyunKo, line 16: Seems like the segmentation for the timing layer is almost same as the main layer. Does it make sense to have only one segmentation and let `Epsilon=0` (or whatever) for the timing layer?

plugins/SCEPCal_MainSDAction.cpp
```diff
@@ -0,0 +1,83 @@
...
+      if(track->GetDefinition()==G4OpticalPhoton::OpticalPhotonDefinition()) {
+        auto procName =track->GetCreatorProcess()->GetProcessName();
+        bool isCerenkov       =(procName =="CerenkovPhys");
+        bool isScintillation  =(procName =="ScintillationPhys");
+        if (!isCerenkov && !isScintillation) return true;
+
+        if (track->GetCurrentStepNumber()==1) {
+          auto* hitSC =newOrExistingHitIn(isScintillation? m_userData.m_collectionID_scint:
+                                                           m_userData.m_collectionID_ceren);
+          hitSC->energyDeposit+=(int)1000; // edep is in MeV, readout is in GeV, therefore a count of 1 is 1000
```
SanghyunKo, line 70: AVOID HARD-CODING UNITS

plugins/SCEPCal_TimingSDAction.cpp
```diff
@@ -0,0 +1,83 @@
...
+      if(track->GetDefinition()==G4OpticalPhoton::OpticalPhotonDefinition()) {
+        auto procName =track->GetCreatorProcess()->GetProcessName();
+        bool isCerenkov       =(procName =="CerenkovPhys");
+        bool isScintillation  =(procName =="ScintillationPhys");
+        if (!isCerenkov && !isScintillation) return true;
+
+        if (track->GetCurrentStepNumber()==1) {
+          auto* hitSC =newOrExistingHitIn(isScintillation? m_userData.m_collectionID_scint:
+                                                           m_userData.m_collectionID_ceren);
+          hitSC->energyDeposit+=(int)1000; // edep is in MeV, readout is in GeV, therefore a count of 1 is 1000
```
SanghyunKo, line 70: SAME (see this example: https://github.com/AIDASoft/DD4hep/blob/master/DDG4/edm4hep/Geant4Output2EDM4hep.cpp#L628)

SanghyunKo (changes requested):
In general, the geometry construction of the main layer's barrel part is hard to read (and may need some comments to understand).

SanghyunKo (changes requested):

detector/calorimeter/SCEPCal_TimingLayer.cpp
```diff
@@ -0,0 +1,495 @@
...
+        double x1y2l=x1y2*tan(gamma-D_GAMMA_TLBARREL/2);
+        double x1y2r=x1y2*tan(gamma+D_GAMMA_TLBARREL/2);
+
+        double verticesT[]={x0y0r,y0e,x1y0r,-y0e,x1y0l,-y0e,x0y0l,y0e,
+                            x0y2r,y2e,x1y2r,-y2e,x1y2l,-y2e,x0y2l,y2e};
+
+        Position dispT(0,0,0);
+
+        XYZVector dispGlobal(rE*sin(thC),rE*sin(thC)*tan(gamma),rE*cos(thC));
+        XYZVector posGlobal=(rotZphiGlobal*dispGlobal)*CLHEP::cm;
```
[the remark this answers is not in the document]
andresailer: There probably shouldn't be any units here, since the original value comes from 
this line 
https://github.com/key4hep/k4geo/pull/443/files/75144c4c03a13c7dee2f628971665fcb160ead1a#diff-01ef598fcddb56791d84deeb7fe87481bdbb686fd587f41f231fbd117c1f687fR45
?
(and some calculations), which is already in dd4hep units?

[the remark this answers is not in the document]
SanghyunKo: @andresailer sorry, you're right - I overlooked they're coming from the xml.

detectorSegmentations/include/detectorSegmentations/SCEPCal_TimingSegmentation_k4geo.h
SanghyunKo: Hi @wonyongc, thanks for the updates! Now it's much better for reading the code. I had a chance to look into the code more carefully, and still have some questions related to this part (and how you retrieve `cellID` in the sensitive detector).

It seems that you're properly setting `PhysVolID` for all crystals and timing layers, and you're also using the 'standard' SD action rather than `regexSensitiveDetectorAction` - which means that you can utilize full 64bits of your `cellID` with `volumeID(step)` instead of using G4 copy number.

For example, in the SD action, instead of 
```c++
G4StepPoint        *thePrePoint =step->GetPreStepPoint();
G4TouchableHandle  thePreStepTouchable =thePrePoint->GetTouchableHandle();

auto cellID =thePreStepTouchable->GetCopyNumber(0);
```
you can use
```c++
auto cellID = volumeID(step)
```
Would you mind checking whether the above option works for you? Then you don't need to struggle to fit your `cellID` into 32 bits.

detector/calorimeter/SCEPCal_TimingLayer.cpp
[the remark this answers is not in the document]
SanghyunKo: I'm still not sure that's the case... For example
```c++
double rGlobal=r0e+XTAL_LEN_F/2;
XYZVector dispGlobal(-center[1],center[0],rGlobal);
XYZVector posGlobal=(rotZphiGlobal*(rotYthGlobal*dispGlobal+DISP_PROJ_R))/dd4hep::mm;
```
If we ignore the dimensionless rotation matrices, the input parameters that decide the dimension are `XTAL_LEN_F`, `DISP_PROJ_R`, etc. If I look up the xml file, they corresponds to
```xml
<constant name="scepcal_projective_offset_r" value="10*cm"/>
<constant name="scepcal_xtal_length_f" value="5*cm"/>
```
so they have proper dimensions in [cm]. Also this will result in different consequences depending on the build option (`dd4hep::mm` is 0.1 in most cases, but it can also be 1 if DD4hep is built with `DD4HEP_USE_GEANT4_UNITS=ON`, see [here](https://github.com/AIDASoft/DD4hep/blob/772ae1561e47961a93e49af6dd0e0a10bfd8fb9b/DDParsers/include/Evaluator/DD4hepUnits.h#L37-L40))

SanghyunKo:
In the meantime: could you also update the README files? ([here](https://github.com/key4hep/k4geo/tree/main/FCCee/IDEA/compact#idea_o2_v01) and [here](https://github.com/key4hep/k4geo/tree/main/detector/calorimeter#calorimeters))

atolosadelgado (commented):
Dear Wonyong

Thank you for your PR. I left some comments along the code. 

I noticed some compilation warnings, I believe they are easy to fix. Could you please fix them? 
[scepcal_warnings_override.txt](https://github.com/user-attachments/files/19483725/scepcal_warnings_override.txt)

I was running some simulations with particle gun (geantinos and protons at 20GeV)  to get an idea of the initialization time and the time per event. Do you have some numbers to crosscheck? 

I got the following, but maybe it is worth to try out with other particles or generated events:
* for geantinos and protons, initalization time of full IDEA around 22 min, SCEPCAL alone 2 min
* for geantinos, full IDEA, 2 ms/geantino
* for protons at 20GeV, full IDEA 8.4s/proton, SCEPCAL alone 6.2 s/proton

Thank you for your time!

FCCee/IDEA/compact/IDEA_o2_v01/SCEPCal_o1_v01.xml
```diff
@@ -0,0 +1,1170 @@
...
+        author="Wonyong Chung"
+        url="https://github.com/wonyongc/SCEPCal"
+        status="release"
+        version="1.0">
+  </info>
+
+  <readouts>
+    <readout name="SCEPCal_Main">
+      <segmentation type="SCEPCal_MainSegmentation_k4geo"/>
+      <id>system:5,phi:7,eta:11,gamma:4,epsilon:4,depth:1</id>
```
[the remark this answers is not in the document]
BrieucF: Thanks for describing the bit field! Can we change from `eta` to `theta`? I understand from the rest of the xml that this is indeed a theta based segmentation.

plugins/SCEPCal_TimingSDAction.cpp
```diff
@@ -0,0 +1,83 @@
...
+      };
+
+      // edep hits
+      auto* hitedep =newOrExistingHitIn(m_collectionID);
+      hitedep->energyDeposit+=edep;
+
+      // Scintillation and Cerenkov hits
+      if(track->GetDefinition()==G4OpticalPhoton::OpticalPhotonDefinition()) {
+        auto procName =track->GetCreatorProcess()->GetProcessName();
+        bool isCerenkov       =(procName =="CerenkovPhys");
```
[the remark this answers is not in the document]
atolosadelgado: for the record, the type/subtype of scintillation is 2/22, and for Cherenkov 2/21. For other EM subprocesses, please check this list [[link](https://github.com/Geant4/geant4/blob/df176550b3e44cdd70406ff2b581896426b5bcbe/source/processes/electromagnetic/utils/include/G4EmProcessSubType.hh#L68)]. I think comparing integers is way faster than a string

detectorSegmentations/include/detectorSegmentations/SCEPCal_TimingSegmentation_k4geo.h
SanghyunKo: @wonyongc `regexSD` is kind of tweak, so I'd recommend the standard one if you're able to manage with it. What is the memory footprint of the standalone SCEP cal?

SanghyunKo: Oh, 10 GB definitely sounds too much... Then it might worth considering `regexSD`. But before we put decent amount of effort on converting SD actions, one more question: How the memory looks like after you comment out every `volume.setSensitiveDetector(sens)` and `placedVolume.addPhysVolID()`? I'd like to know whether it's the issue of SD action (like it was for fiber DRC) or it's the issue of geometry itself.

FCCee/IDEA/compact/IDEA_o2_v01/SCEPCal_o1_v01.xml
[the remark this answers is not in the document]
BrieucF: Thanks!

detectorSegmentations/include/detectorSegmentations/SCEPCal_TimingSegmentation_k4geo.h
SanghyunKo: Sure, we can leave two segmentations separate, I have no problem with it. On the other hand, the memory of the geometry itself needs a certain amount of optimization, but I'd opt to merge this PR first (we have disabled CI for IDEA_o2) and seek for another PR for the optimization since it'll take some time. @BrieucF what do you think?

detector/calorimeter/SCEPCal_TimingLayer.cpp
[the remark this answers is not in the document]
SanghyunKo: I think I have figured out what's going on. We have two bugs, one in the geometry construction, and one in the SD action so that the factor 10 cancels out by coincident.

First things first, the default unit for DD4hep is cm (unless we set `DD4HEP_USE_GEANT4_UNITS=ON`), while the default unit for Geant4 (or CLHEP) is mm. The geometry construction itself purely belongs to DD4hep and irrelevant to Geant4. So the unit of length throughout the whole detector construction must be in cm (so we don't want to divide by `dd4hep::mm` in the geometry construction).

Nevertheless, DDG4 expects mm as the unit length stored in `Geant4Calorimeter::Hit`. If the user takes the number from Geant4 itself (e.g. `G4StepPoint::GetPosition()`), there is no need of unit conversion. However, if it is taken from the `DDSegmentation`, an explicit unit conversion happens in the SD action because the segmentation will give the number in cm ([here](https://github.com/AIDASoft/DD4hep/blob/3466ee762ed148ff4f798292c90e229e53b6c0bd/DDG4/plugins/Geant4SDActions.cpp#L62-L63), for example). However, this doesn't seem to be your case, as the position given by the segmentation is substituted to the `Geant4Calorimeter::Hit` as it is without any unit conversion (see [here](https://github.com/wonyongc/k4geo/blob/fc628ce586845aced0eca6515bbe582cb4e80fc0/plugins/SCEPCal_MainSDAction.cpp#L42-L56), and this is the place where you need to divide by `dd4hep::mm`).

The current output might look okay, but it could be a time bomb ticking away because if someone calls `segmentation->position()` out of the box (during clustering, for example), then it'll give us a number with a factor 10 larger and the downstream algorithms will fail.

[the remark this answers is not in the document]
SanghyunKo: @wonyongc yes, you need to do an explicit conversion if and only if you move the quantity to different type of containers from one another (DD4hep to Geant4/Edm4hep).

SanghyunKo (approved):

SanghyunKo:
Now I'm fine with this (as the first round of development; it has to be optimized further later)

SanghyunKo (commented):
(I guess changes are code-formatting?)

BrieucF:
There seem to remain just one small pre-commit complain: https://github.com/key4hep/k4geo/actions/runs/15708967866/job/44262131538?pr=443#step:4:128

BrieucF:
Thanks Wonyong, can you rebase?

BrieucF:
One more https://github.com/key4hep/k4geo/actions/runs/15870344776/job/44745380827?pr=443#step:4:139 ;-)

PR #446, opened 2025-03-19
update tracker region extent in ALLEGRO xml

giovannimarchiori:
The constants tracker_region_zmax and tracker_region_rmax are used in reconstruction to define the tracking volume.
Currently these are set equal to the DCH outer R and Z
Since there is also the wrapper in the tracking volume, these constants are redefined to include also the wrapper
- [X] the code is sufficiently well documented (e.g. inline comments)
- [X] the PR source branch has been rebased (`--ff-only`) to `k4geo/main`
- [X] the PR does not contain any additions or modifications that do not belong to it
- [X] The release notes below contain a succinct and comprehensive description of the changes that are sufficiently self-explanatory

BEGINRELEASENOTES
- Update ALLEGRO tracker_region zmaz, rmax constants to also include the wrapper

ENDRELEASENOTES


giovannimarchiori:
just rebased

Issue #449, opened 2025-03-24
Add FCCconfig tests to k4geo


atolosadelgado:
Hi,

Would it be worth integrating downstream CI tests into k4geo to prevent issues like this [1]? I'm specifically thinking about FCCconfig (and maybe CLDconfig), which manage the configuration for running the full chain.

After a quick chat with @jmcarcell, it seems the fastest and simplest solution would be to add tests in k4geo equivalent to those in FCCconfig.

@BrieucF @andresailer, what do you think?

[1] https://github.com/key4hep/k4geo/pull/438

PR #450, opened 2025-03-24
Exit a loop earlier and cache a map lookup in TrackerBarrel_o1_v06_geo

jmcarcell:
BEGINRELEASENOTES
- Exit a loop earlier and cache a map lookup in TrackerBarrel_o1_v06_geo

ENDRELEASENOTES

Behaviour is the same and it is slightly faster in Release builds when caching the map lookup even though the cellID doesn't change, it seems the compiler isn't smart enough. Having a look around it seems basically nothing is using the `NeighbourSurfacesData` except here: https://github.com/iLCSoft/MarlinTrkProcessors/blob/master/source/Refitting/src/ExtrToTracker.cc#L814 that is not run for CLD. Maybe it could be considered to be removed?

### DIFF
--- a/detector/tracker/TrackerBarrel_o1_v06_geo.cpp
+++ b/detector/tracker/TrackerBarrel_o1_v06_geo.cpp
@@ -42,9 +42,10 @@ using dd4hep::Volume;
 using dd4hep::rec::NeighbourSurfacesData;
 using dd4hep::rec::ZPlanarData;
 
-void populateNeighbourData(NeighbourSurfacesData* neighbourSurfacesData, UTIL::BitField64& encoder, int module_idx,
-                           int sensor_idx, int nphi, int nz) {
+void populateNeighbourData(NeighbourSurfacesData* neighbourSurfacesData, UTIL::BitField64& encoder,
+                           const int module_idx, const int sensor_idx, const int nphi, const int nz) {
   const dd4hep::CellID cellID = encoder.lowWord(); // 32 bits
+  auto& neighbours = neighbourSurfacesData->sameLayer[cellID];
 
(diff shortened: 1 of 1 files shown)


Zehvogel (approved):
Looks good to me. How large is the performance improvement?

Regarding the actual usage of the `neighbourSurfacesData` I don't know more.

atolosadelgado:
Hi @jmcarcell @Zehvogel 

Thanks for this PR and reviewing it! 

Have you tried your modifications in a full chain of simulation+reconstruction? I think at the moment we do not include these kind of tests in k4geo (see https://github.com/key4hep/k4geo/issues/449)

jmcarcell:
> Looks good to me. How large is the performance improvement?
> 
> Regarding the actual usage of the `neighbourSurfacesData` I don't know more.

I think it's in the order of magnitude of 100 ms in the debug builds I use for development, 10 ms in release builds. I think anyway most of the work happens in DD4hep. Maybe @andresailer knows if we need `neighbourSurfacesData` or is it a relic from the past? In any case it doesn't matter for this PR.


> Hi @jmcarcell @Zehvogel
> 
> Thanks for this PR and reviewing it!
> 
> Have you tried your modifications in a full chain of simulation+reconstruction? I think at the moment we do not include these kind of tests in k4geo (see #449)

Indeed that is how it should be done even if in principle no behavior is changed in this PR. I ran a simulation with the same seed before and after this and the resulting files are identical, so I'm happy about that. Ready to merge I would say.

andresailer:
Yes, we can remove it.

jmcarcell:
Let's starting merging this while we have it and then remove it in another PR.

PR #451, opened 2025-03-25
Fix a few compiler warnings

jmcarcell:
BEGINRELEASENOTES
- Fix compiler warnings caught with clang 19
- Add a missing `#include <list>` that will be needed with https://github.com/AIDASoft/DD4hep/pull/1432

ENDRELEASENOTES

The changes in the muon system have appeared before, @mahmoudali2, in
https://github.com/key4hep/k4geo/pull/353.

Not all the warnings are fixed here since others may require some validation:

@varnes, this could be wrong since whatever the pointer points to may or may not be that string:
```
/k4geo/detector/calorimeter/ECalEndcap_Turbine_o1_v02_geo.cpp:125:37: warning: object backing the pointer will be destroyed at the end of the full-expression [-Wdangling-gsl]
  125 |     char* nUnitCellsStrArr = (char*)genericBladeElem.attr<std::string>(_Unicode(nUnitCells)).c_str();
```

@Archil-AD, it complains about hiding the original `neighbours` in the base class Segmentation . If that was not intended I would simply suggest to change the name to something else like `getNeighbours`. In general it's good practice to call different things witht different names.

```
/k4geo/detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiRow_k4geo.h:54:27: warning: 'dd4hep::DDSegmentation::FCCSWHCalPhiRow_k4geo::neighbours' hides overloaded virtual function [-Woverloaded-virtual]
   54 |     std::vector<uint64_t> neighbours(const CellID& cID) const;
      |                           ^
/DD4hep/install/include/DDSegmentation/Segmentation.h:89:20: note: hidden overloaded virtual function 'dd4hep::DDSegmentation::Segmentation::neighbours' declared here: different number of parameters (2 vs 1)
   89 |       virtual void neighbours(const CellID& cellID, std::set<CellID>& neighbours) const;
```

### DIFF
--- a/detector/calorimeter/ECalEndcap_Turbine_o1_v01_geo.cpp
+++ b/detector/calorimeter/ECalEndcap_Turbine_o1_v01_geo.cpp
@@ -4,15 +4,6 @@
 #include "XML/Utilities.h"
 #include <DDRec/DetectorData.h>
 
-// todo: remove gaudi logging and properly capture output
-#define endmsg std::endl
-#define lLog std::cout
-namespace MSG {
-const std::string ERROR = " Error: ";
-const std::string DEBUG = " Debug: ";
-const std::string INFO = " Info: ";
-} // namespace MSG
-
 namespace det {
 
 unsigned ECalEndCapElementCounter = 0;
--- a/detector/calorimeter/ECalEndcap_Turbine_o1_v02_geo.cpp
+++ b/detector/calorimeter/ECalEndcap_Turbine_o1_v02_geo.cpp
@@ -5,17 +5,6 @@
 #include "XML/Utilities.h"
 #include <DDRec/DetectorData.h>
 
-// todo: remove gaudi logging and properly capture output
-#define endmsg std::endl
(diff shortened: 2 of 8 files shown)


jmcarcell:
Yes, we should merge this one because the `#include <list>` will prevent `k4geo` from compiling with the current version of DD4hep. The other two warnings we should also try to fix them with @varnes and @Archil-AD, but we can do so in a later PR.

varnes:
Hi @jmcarcell ,

 I'd like to fix the warning in my code, but I can't see to reproduce it (i.e. when I check out and compile, there are no warnings).  Is there a certain compiler flag I need to set?

Thanks,
Erich

jmcarcell:
Hi @varnes, 

if you are using the stack, then I think the easiest is simply to switch to clang by running:
```
export CXX=$(which clang++)
```
And then clean the build directory and compile, you'll see the warning I'm mentioning (among other warnings). I couldn't find a similar warning for GCC :shrug:

PR #453, opened 2025-03-26
CLD_o2_v07: make tracking volume actually parallel

Zehvogel:
BEGINRELEASENOTES
- CLD_o2_v07: made tracking volume actually parallel

ENDRELEASENOTES

looks like I overlooked this :(
with `connected="true"` the volume is placed in the world not the parallelworld
https://github.com/AIDASoft/DD4hep/blob/f775dbf817dde0e6c1685bd8dccb87dc25136af5/DDCore/src/plugins/Compact2Objects.cpp#L1587

### DIFF
--- a/FCCee/CLD/compact/CLD_o2_v07/CLD_o2_v07.xml
+++ b/FCCee/CLD/compact/CLD_o2_v07/CLD_o2_v07.xml
@@ -262,7 +262,7 @@
        else if the volume is part of the parallelworld:  connected="false"
        The volume is always connected to the top level!
     -->
-    <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="true"
+    <parallelworld_volume name="tracking_volume" anchor="/world" material="Air" connected="false"
         vis="VisibleBlue">
         <shape type="Polycone" material="Air">
             <!-- small-angle approximation for tan(theta) -->


atolosadelgado:
Hi @Zehvogel 

Thank you for your PR! Is there a k4geo test that covers your changes? If not, what can I run to cross-check that your changes do not affect other repos? I was thinking of running the full CLD reconstruction chain, but I wasn't sure it would cover everything.

Zehvogel:
> Thank you for your PR! Is there a k4geo test that covers your changes? If not, what can I run to cross-check that your changes do not affect other repos? I was thinking of running the full CLD reconstruction chain, but I wasn't sure it would cover everything.

No. The only way to validate this at the moment is manually. This will also not change until someone finally sets up some extensive validation for CLD.

I am also not sure how you would even test for this change specifically.

jmcarcell:
Can you explain why are you making this change (link the DD4hep issue?) and what visible change will there be for someone working on CLD in the release notes?

atolosadelgado:
thanks @Zehvogel for the clarification!

shall we merge @jmcarcell ?

jmcarcell:
I get the same results as before if I run conformal tracking after this change.

andresailer:
> I get the same results as before if I run conformal tracking after this change.

Did you re-run simulation?

jmcarcell:
Yes, I got exactly the same output when simulating some muons and then also ran conformal tracking (the gaudi native one) with and without the change.

PR #458, opened 2025-04-04
Allow crosstalk coefficients in the input to the creation of crosstalk neighbour mapping

zwu0922:
BEGINRELEASENOTES
- Crosstalk coefficients can be provided as input argument in the `det::crosstalk::getNeighboursModuleThetaMerged` function
- No impact on other parts of k4geo, but [k4RecCalorimeter](https://github.com/HEP-FCC/k4RecCalorimeter/blob/main/RecFCCeeCalorimeter/src/components/CreateFCCeeCaloXTalkNeighbours.cpp) needs to be updated accordingly.

ENDRELEASENOTES


detectorCommon/include/detectorCommon/xtalk_neighbors_moduleThetaMergedSegmentation.h
```diff
@@ -23,7 +23,11 @@ namespace crosstalk {
    *   return Vector of neighbours and their crosstalk coefficients.
    */
   std::vector<std::pair<uint64_t, double>>
-  getNeighboursModuleThetaMerged(const dd4hep::DDSegmentation::FCCSWGridModuleThetaMerged_k4geo& aSeg,
+  getNeighboursModuleThetaMerged(const double xtalk_coef_radial,
```
andresailer, line 26: * Please update according to code style
* Please add parameters to documentation
* Why not move the new parameters to the end of the function arguments and use default values that were in the code before for compatibility with the existing code?

zwu0922: * Please update according to code style

The code is updated to conform the coding style (indents and naming convention of variables)


  * Please add parameters to documentation

Done


 * Why not move the new parameters to the end of the function arguments and use default values that were in the code before for compatibility with the existing code?

The order of function arguments is adjusted, with default values assigned for cross-talk coefficients, so as to be compatible with the existing code.

andresailer (approved):

PR #467, opened 2025-04-17
Enforce clang format

tmadlener:
BEGINRELEASENOTES
- Add a `clang-format` hook to our pre-commit configuration to actually enforce the formatting that we expect via `.clang-format`.

ENDRELEASENOTES

I am not entirely sure whether these files have gone out of conformance due to changes after the `.clang-format` file was added in https://github.com/key4hep/k4geo/commit/b53e9c6d4eb3aa47838d237654ba204f6a952022 or whether these were not included in the formatting in https://github.com/key4hep/k4geo/commit/221a469b46992dba67ed902455876b4ffb46ce5b . Anyhow, I think there is little meaning for having a `.clang-format` config if we don't enforce it.

This has the potential to introduce merge conflicts in other PRs that also touch these files, I haven't yet checked whether that is the case.

### DIFF
--- a/.github/scripts/clang-format-hook
+++ b/.github/scripts/clang-format-hook
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Thin wrapper around clang-format for easier to parse output from the
+# pre-commit hook.
+#
+# Needs to work with multiple input files as pre-commit passes multiple files to
+# the "executables"
+
+# Make sure that diff is actually recent enough (diffutils >= 3.4) to support
+# colored output
+COLOR_OUTPUT=$(diff --color=always <(echo) <(echo) > /dev/null 2>&1 && echo "--color=always")
+
+success=0
+for file in ${@}; do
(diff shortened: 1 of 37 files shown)


.github/scripts/clang-format-hook
```diff
@@ -0,0 +1,21 @@
...
+# Needs to work with multiple input files as pre-commit passes multiple files to
+# the "executables"
+
+# Make sure that diff is actually recent enough (diffutils >= 3.4) to support
+# colored output
+COLOR_OUTPUT=$(diff --color=always <(echo) <(echo) > /dev/null 2>&1 && echo "--color=always")
+
+success=0
+for file in ${@}; do
+    if ! $(clang-format --style=file --Werror -i ${file} > /dev/null 2>&1); then
```
tmadlener, line 15:
```suggestion
    if ! $(clang-format --style=file --Werror --dry-run ${file} > /dev/null 2>&1); then
```

jmcarcell (approved):

PR #468, opened 2025-04-17
[WIP] Add the MAIA detector concept geometry

tmadlener:
BEGINRELEASENOTES
- Add the `MAIA` muon collider detector concept

ENDRELEASENOTES

This is the first of a series of PRs that aim at migrating existing muon collider geometries from different sources (forks and other repositories) back to k4geo. The main thing that we should settle early in the process is probably the name of the muon collider subfolder, which I currently simply put to `MuColl`

The origin of this geometry is: https://github.com/madbaron/detector-simulation/tree/KITP_10TeV/geometries/MAIA_v0

It has been moved to k4geo using `git-filter-repo` to keep the original history and currently there is no from of cleanup done.


MuColl/MAIA/compact/MAIA_v0/HCalEndcap_o1_v01_01.xml
```diff
@@ -0,0 +1,163 @@
...
+		<!--readout name="HCalRingCollection">
+			<segmentation type="CartesianGridXY" grid_size_x="HCal_cell_size" grid_size_y="HCal_cell_size" />
+			<id>${GlobalCalorimeterReadoutID}</id>
+		</readout-->
+	</readouts>
+
+
+	<detectors>
+
+		<detector name="HCalEndcap" type="GenericCalEndcap_o2_v01" id="DetID_HCAL_Endcap" readout="HCalEndcapCollection" vis="HCALVis" >
```
tmadlener, line 55: This is where the discussion connects to https://github.com/key4hep/k4geo/pull/341

tmadlener:
I have removed the WIP tag as this is essentially ready (following the discussion in #469 about naming conventions). It introduces a dedicated `MAIA` subfolder for MAIA inside `MuColl`. Ideally we also have a README in this folder. @madbaron is there one we can already use which briefly summarizes the MAIA concept (and potentially how it came to be from a technical level)?

We could also think about adding a small test using a particle gun that would at least ensure the geometry loads and can be used for simulation. Something like:

https://github.com/key4hep/k4geo/blob/d233e48862c253274c4e3a54257900691e389d72/test/CMakeLists.txt#L49-L53

madbaron:
Hi @tmadlener,

we don't have a good README to pick, but we can make one.
Can you point me to a good example?

tmadlener:
There is a very comprehensive one for ILD: https://github.com/key4hep/k4geo/tree/main/ILD/compact

For MAIA it might be good enough to start with something like IDEA: https://github.com/key4hep/k4geo/tree/main/FCCee/IDEA/compact

madbaron:
Could you please pick up the latest commits for the KITP_10TeV branch (https://github.com/madbaron/detector-simulation/tree/KITP_10TeV) from the past few days?

They get rid of a few remaining volume overlaps we had in the v0 geometry, as well as delete some of the unused CLIC systems

madbaron (approved):
Looks good to me! Thanks!

PR #469, opened 2025-04-17
Add all historic muon collider geometries

tmadlener:
BEGINRELEASENOTES
- Add the "historic" (i.e. not MAIA or MuSIC) muon collider geometries to `MuColl/MuColl/compact`

ENDRELEASENOTES

This is the second in the (now admittedly short) series of PRs to bring back muon collider detector concept geometries to the central k4geo repositories with the goal to retire all other existing forks and sources. The other one is #468.

These are the ones that currently live in https://github.com/MuonColliderSoft/lcgeo

Main things that are still to do:
- [x] harmonize between this and #468 regarding the substructure below the top-level `MuColl` directory.
  - [x] Should `MuSIC` and `MAIA` and the 3 TeV models be separate sub-folders in `MuColl` (similar to what [FCCee](https://github.com/key4hep/k4geo/tree/main/FCCee) currently has? -> **YES**
- [x] Decide whether we want to keep all of the geometries or whether this is a point in time to get rid of some. -> **YES, but only test MuColl_v1**
- [x] Includes #341 

FYI @madbaron (feel free to tag whoever needs to be tagged here).

Similar to #468 git history is kept in tact via `git-filter-repo` and some rebasing to have some form of sane git history (as far as possible).


madbaron:
mmh probably adopting an FCCee-like directory structure is a little tidier, so I'd lean towards doing that.

Adding also @bartosik-hep and @davzuliani for their info

bartosik-hep:
There are some things to be fixed yet in the MuSIC geometries, which also should be added to the `README.md` file. And I agree with adding sub-folders for separating CLIC, MuColl and MuSIC geometries.
@tmadlener Do you think it would be easier to create a dedicated `mucoll` branch in this repo and use it as a target for this PR, such that me and @davzuliani can clean them up before merging into `main` branch?

tmadlener:
> @tmadlener Do you think it would be easier to create a dedicated mucoll branch in this repo and use it as a target for this PR, such that me and @davzuliani can clean them up before merging into main branch?

whatever is easiest for you. If you want you can also finish that work in the muon collider fork and then I can play the same trick again as I did before to get them here.

In any case, if we create dedicated subfolders for the different concepts, I can split off the 3 TeV models into another PR and we make a dedicated one for MuSIC once things are ready.

tmadlener:
I have split off the MuSIC geometries from this PR so that we can merge those in later once they are ready. This PR now only contains the "historic" muon collider geometries. I have also adapted the directory structure to be `MuColl/MuColl/compact` similar to what FCC does and also to what we already do for MAIA in #468. 

Are there particular models for which there should be basic tests that ensure that the geometry loads and can be used for at least some particle gun tests?

bartosik-hep:
`MuColl_v1` was several production campaigns, which could be worth including in a test. The rest are not essential for testing

tmadlener:
Do you think a simple particle gun is enough for testing? Alternatively we could use some input file that covers a few more cases, e.g. a ttbar event or something in that direction.

tmadlener:
Added a simple test for `MuColl_v1`, that effectively checks that the geometry loads within ddsim.

tmadlener:
Rebased this onto #341 and switched over to use the `o1_v01` with a `conicalCutout` to get some CI running already now

PR #478, opened 2025-05-26
CLD_o2_v08

andresailer:

BEGINRELEASENOTES
- CLD_o2_v08: new detector model taking into account Birks Law for the scintillator hcal.

ENDRELEASENOTES


Zehvogel (approved):
Looks good to me! Wish we would have any validation plot for this...

jmcarcell:
I think for the future, when adding new options or versions we could divide the process in two PRs, one copying some existing option or version (in this case CLD_o2_v07) and the other one with the changes, then it would be possible to see what has changed in the diff.

PR #483, opened 2025-06-16
correct x/y components extracted from Br-Z field map

danieljeans:
BEGINRELEASENOTES
- FieldMapBrBz: Correct bug in the conversion of a radial field component to x-y components. Previously x and y were interchanged.
ENDRELEASENOTES

### DIFF
--- a/detector/other/FieldMapBrBz.cpp
+++ b/detector/other/FieldMapBrBz.cpp
@@ -139,8 +139,8 @@ void FieldMapBrBz::fieldComponents(const double* pos, double* globalField) {
   field[1] = (1.0 - rd) * (1.0 - zd) * B_r0z0.Bz + rd * (1.0 - zd) * B_r1z0.Bz + (1.0 - rd) * zd * B_r0z1.Bz +
              rd * zd * B_r1z1.Bz;
 
-  globalField[0] += field[0] * sin(phi);
-  globalField[1] += field[0] * cos(phi);
+  globalField[0] += field[0] * cos(phi);
+  globalField[1] += field[0] * sin(phi);
   globalField[2] += field[1];
 
   /*


danieljeans:
an independent check of my logic would be much appreciated (but I've convinced myself)

danieljeans:
> Here: /cvmfs/ilc.desy.de/sw/x86_64_gcc44_sl6/v01-17-06/Mokka/mokka-08-04/source/
> 
> /cvmfs/ilc.desy.de/sw/x86_64_gcc44_sl6/v01-17-06/Mokka/mokka-08-04/source/Geometry/Tesla/src/FieldX03.cc

so we should blame @gaede ;-) ?

andresailer:
Probably much older than that, and we should check that there is no even number of "mistakes"?
Also blame the collective review process!

tmadlener:
I checked the geometry in which this is *actively* used (actually included and loaded in the xml config somehow, and not leftover in an un-included xml file or commented out). As far as I figured out now, it's only ILD geometries (apart from a mention in `MuColl_v1`, which is however not included in the loaded geometry, [here](https://github.com/key4hep/k4geo/pull/469/files#diff-720182142c16102d1f9544dfd18ed2b0afb461a7722dd73345feb71ff4c5509b)

- ILD_l5_v06
- ILD_s5_v04
- ILD_s5_v08
- ILD_s5_o1_v07
- ILD_l5_v05
- ILD_l5_o1_v05
- ILD_l5_v07
- ILD_s5_o1_v04
- ILD_l5_o1_v04
- ILD_s5_v05
- ILD_s5_o1_v08
- ILD_l5_o1_v07
- ILD_s5_v07
- ILD_l5_o1_v03
- ILD_l5_v04
- ILD_l5_v03
- ILD_l5_v08
- ILD_s5_o1_v03
- ILD_l5_o1_v08
- ILD_s5_o1_v06
- ILD_l5_o1_v06
- ILD_s5_o1_v05
- ILD_s5_v03
- ILD_s5_v06

andresailer:
Should make a drawing of some slice of the field to visualise the impact?

jmcarcell:
Basically, exchanging `sin` and `cos` is a reflection (X or Y axis, it doesn't matter) plus a rotation of 90 degrees, for example with y
`y = cos(phi), y'=-cos(phi), y''=-cos(phi+90)=sin(phi)`

PR #486, opened 2025-06-19
Fiber DRC segmentation and SD action update for downstream digitization and reconstruction

SanghyunKo:

BEGINRELEASENOTES
- Use the front end instead of the rear end for the default DRC calo hit position
- Added a dedicated function to calculate the rear end position of the fiber DRC cells (only needed for the DRC digitizer)
- Made the neighborhood definition of the fiber DRC configurable via a compact file
- Fixed a bug that the fiber DRC SD action did not take the full active volume of the scintillation fiber for the energy deposit

ENDRELEASENOTES

1. Currently, we're using the rear end (where the readout electronics will be located) for the calo hit position of the DRC. However, this creates an issue when it comes to the track-cluster matching because it uses the distances between the extrapolated track state and the calo hit position, which would be ~ 2m in average if we use the rear end for the calo hit position. So I made a change to use the front end for the calo hit position.
2. The previous method (returning the rear end position) has been implemented into a dedicated function of `GridDRcalo_k4geo`. It is needed (and only needed) for the digitizer of the DRC that estimates the pulse timing based on the distance between the `step` and the readout.
3. Updated the neighborhood finding function in a way that the size of the neighborhood and the switch to include both types of cells (scintillation and Cherenkov) are configurable via a compact file.
4. Fixed a bug that the previous fiber DRC SD action did not take the full active volume of the scintillation fiber for the energy deposit.

Depends on https://github.com/HEP-FCC/k4RecCalorimeter/pull/167

### DIFF
--- a/detectorSegmentations/src/GridDRcalo_k4geo.cpp
+++ b/detectorSegmentations/src/GridDRcalo_k4geo.cpp
@@ -63,6 +63,8 @@ namespace DDSegmentation {
     dd4hep::Position localPos = dd4hep::Position(0., 0., 0.);
 
     // get local coordinate
+    // if IsSiPM is false, the local position is (0, 0, 0)
+    // and the function returns the tower's position
     if (IsSiPM(cID))
       localPos = dd4hep::Position(localPosition(cID));
 
@@ -109,6 +111,8 @@ namespace DDSegmentation {
     dd4hep::Position localPos = dd4hep::Position(0., 0., 0.);
 
     // get local coordinate
+    // if IsSiPM is false, the local position is (0, 0, 0)
+    // and the function returns the wafer's position
     if (IsSiPM(cID))
       localPos = dd4hep::Position(localPosition(cID));
 


detectorSegmentations/src/GridDRcalo_k4geo.cpp
```diff
@@ -179,31 +194,20 @@ namespace DDSegmentation {
     auto southEast = setCellID(isRHS, systemId, noEta, noPhi, nX + 1, nY - 1);
     auto southWest = setCellID(isRHS, systemId, noEta, noPhi, nX - 1, nY - 1);
     auto northWest = setCellID(isRHS, systemId, noEta, noPhi, nX - 1, nY + 1);
-    // dist=2 cells
-    auto north = setCellID(isRHS, systemId, noEta, noPhi, nX, nY + 2);
-    auto east = setCellID(isRHS, systemId, noEta, noPhi, nX + 2, nY);
-    auto south = setCellID(isRHS, systemId, noEta, noPhi, nX, nY - 2);
-    auto west = setCellID(isRHS, systemId, noEta, noPhi, nX - 2, nY);
+    // dist=1 cells
+    auto north = setCellID(isRHS, systemId, noEta, noPhi, nX, nY + 1);
```
andresailer, line 198: Is this change related to the change in the chosen calo hit position?

andresailer: Thanks!

FCCee/IDEA/compact/IDEA_o1_v03/FiberDualReadoutCalo_o1_v01.xml
```diff
@@ -594,6 +594,11 @@
       </sipmDim>
       <structure>
         <dim distance="1.5*mm" dx="1.0*mm"/>
+        <!--parameters for neighborhood finding algorithm-->
+        <!--rmax1/rmax2: radius of the neighborhood in units of fiber-fiber distance in barrel/endcap-->
+        <!--x1/x2: margin to incorporate satelite cells at the edge of the tower in barrel/endcap-->
+        <!--delta: flag to remove different type of cells in the neighborhood-->
```
BrieucF, line 600: can you specify how it is interpreted? Adding e.g. "If set to 0, cherenkov and scintillating fibers will be considered as neighbors". Maybe we could also have a more explicit name than "delta" for this parameter?

detectorSegmentations/src/GridDRcalo_k4geo.cpp
```diff
@@ -68,13 +70,64 @@ namespace DDSegmentation {
     auto rot = paramBase->GetRotationZYX(noPhi);
     auto translation = rot * localPos;
 
+    // we want the cell position to be the front end of the tower
+    // retrieve the fiber length
+    int row = y(cID);
+    int col = x(cID);
+    double fiberLen = paramBase->GetShortFibers(noEta).retrieveFiberLength(row, col);
```
BrieucF, line 77: I found a bit confusing to always call this through the short fibers, while some fibers here have the full length. Can you add here a comment stating that this method returns the full length fiber dimension if the row col do not correspond to any short fiber?

detectorSegmentations/src/GridDRcalo_k4geo.cpp
```diff
@@ -68,13 +70,64 @@ namespace DDSegmentation {
...
+    // and vector to the front end
+    auto diff = paramBase->GetTowerPos(noPhi) - waferPos;
+    auto unitVec = diff / std::sqrt(diff.Mag2());
+
+    // total vector is sum of the vector to the wafer center + rotated local coordinate
+    // + the translation to the front end of the tower
+    auto total = translation + waferPos + unitVec * fiberLen;
+
+    // if LHS rotate by 180 deg w.r.t. X axis (on par to the DRconstructor)
```
BrieucF, line 86: If not there already, can you add a comment in the DRConstructor, where you do the rotation, that downstream code rely on this i.e. if for some reason we change this in the future, the person doing so will know that more modifications are needed?

detectorSegmentations/src/GridDRcalo_k4geo.cpp
```diff
@@ -68,13 +70,64 @@ namespace DDSegmentation {
...
+    bool isRHS = IsRHS(cID);
+
+    // first get a vector to the center of the wafer (per tower)
+    DRparamBase_k4geo* paramBase = setParamBase(noEta);
+    auto waferPos = paramBase->GetSipmLayerPos(noPhi);
+
+    dd4hep::Position localPos = dd4hep::Position(0., 0., 0.);
+
+    // get local coordinate
+    if (IsSiPM(cID))
```
BrieucF, line 110: why is this `if` needed?

FCCee/IDEA/compact/IDEA_o1_v03/FiberDualReadoutCalo_o1_v01.xml
SanghyunKo: Added a line of comment as suggested. As for the parameter name, IIUC the name should match one of the available names in the [`DD4hep/DDCore/include/XML/UnicodeValues.h`](https://github.com/AIDASoft/DD4hep/blob/master/DDCore/include/XML/UnicodeValues.h). It won't compile otherwise (correct me if I'm wrong).

detectorSegmentations/src/GridDRcalo_k4geo.cpp
SanghyunKo: This comes from a historical reason - this function was designed to return the position of the tower (precisely "wafer" behind the tower) rather than the individual hit position when `IsSiPM` returns 0.

SanghyunKo: Sure, I've just done it now

SanghyunKo:
The downstream test failure is expected (needs HEP-FCC/k4RecCalorimeter#167)

PR #493, opened 2025-07-14
Fix build failure when `K4GEO_USE_LCIO=OFF`

SanghyunKo:
BEGINRELEASENOTES
- Fix build failure when `K4GEO_USE_LCIO=OFF`

ENDRELEASENOTES

Fixing build failure because of the two source files requiring `UTIL/LCTrackerConf.h` that are still included even if `K4GEO_USE_LCIO=OFF`.


andresailer (approved):
Thanks!

PR #499, opened 2025-07-25
SCEPCal partial construction

Robyn-Evren:
BEGINRELEASENOTES
- New variables added for partial construction of SCEPCal 

ENDRELEASENOTES

- 3 Files changed 
- Able to control phi, theta and gamma 
- Phi can loop around same as for H-cal, recommend only changing gamma if you are loading only 1 phi slice.

### DIFF
--- a/FCCee/IDEA/compact/IDEA_o2_v01/DectDimensions_IDEA_o2_v01.xml
+++ b/FCCee/IDEA/compact/IDEA_o2_v01/DectDimensions_IDEA_o2_v01.xml
@@ -233,9 +233,9 @@
     <constant name="scepcal_timing_xtal_length" value="10*cm"/>
 
     <!-- SCEPCal variables for partial construction -->
-    <!-- Barrel phi segment loading, no. segments defined above, base start="0", main layer only -->
-    <constant name="scepcal_barrel_phi_load_start" value="0"/>
-    <constant name="scepcal_barrel_phi_load_end" value="127"/>
+    <!-- Barrel and endcap phi segment loading, no. segments defined above, base start="0", main layer only -->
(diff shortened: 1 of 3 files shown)


SanghyunKo (approved):
Also good to go from my side @BrieucF

wonyongc:
I am going to have to revert this as a bunch of my runs were made invalid (testing crystal sizes smaller than 1cm) because of missing chunks of the detector by the hardcoded theta bound numbers in the xml, of which I was not aware before starting the runs. 
There is no indication or expectation that one should be aware of hardcoded geometry bounds in the xml before running.
Furthermore, my original comment also stands. There is no reason for the newly added 'load_start/load_end' variables. It is literally the exact thing as the 'PHI_START/END' and 'N_THETA_BARREL/GAMMA' variables already in there. All that is needed is to point the xml attribute to these, instead of making new variables. This change did not even introduce the ability to specify the range of degrees to load, which was the only actual addition that was proposed. It still uses the index numbers, same as before.
By making new 'load' variables this 1) adds unnecessary redundancy in loop control, and 2) is wasteful, because the 'load' variables block out only the volume placement but leaves the rest of the creation loop intact. The proper way to skip the entire loop for unwanted sections is to make the change in the existing variables, as I said.
In fact, I have done, un-done, and re-done this exact thing many, many times in my private fork while testing the geometry build to do partial visualizations. In the end, I decided to leave it out of the main repo because the thought was that when running the full sim in production, one does not want to perturb global geometry build boundaries with hardcoded, arbitrary values.
No one is checking to make sure the entire detector is in fact being built. That should be a given.
You do not want people to find out they wasted hundreds of hours of compute on batch runs for a broken geometry after the fact.

PR #502, opened 2025-08-01
Update tests to allow running them without installing

jmcarcell:
BEGINRELEASENOTES
- Update tests to allow running them without installing (i.e. we don't need to run `make install` or `ninja install` before running the tests
- Remove "test_IDEA_with_DRC_o1_v03" that was duplicated, keep only the one that simulates electrons

ENDRELEASENOTES

Using add_test(NAME ...) complains when there is already another test with the given name, while what we were using before does not so that's an advantage.

Tagging @SanghyunKo since he was the last one modifying the IDEA test

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -118,21 +118,12 @@ add_test_with_env(t_${test_name} ddsim --compactFile=${CMAKE_CURRENT_SOURCE_DIR}
 SET_TESTS_PROPERTIES( t_${test_name} PROPERTIES FAIL_REGULAR_EXPRESSION  " Exception; EXCEPTION;ERROR;Error" TIMEOUT 900)
 endif()
 
-#--------------------------------------------------
-# test for IDEA o1 v03, with DRC, with Geantinos
-if(DCH_INFO_H_EXIST)
-SET( test_name "test_IDEA_with_DRC_o1_v03" )
-ADD_TEST( t_${test_name} "${CMAKE_CURRENT_BINARY_DIR}/run_test_${PackageName}.sh"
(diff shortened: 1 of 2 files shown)


test/CMakeLists.txt
```diff
@@ -1,199 +1,152 @@
...
-
-macro(SET_TEST_ENV test_name)
+function(SET_TEST_ENV test_name)
   # Make sure to point our current installation for the compact files
-  SET_TESTS_PROPERTIES( t_${test_name}
+  SET_TESTS_PROPERTIES( ${test_name}
     PROPERTIES
       FAIL_REGULAR_EXPRESSION  "Exception;EXCEPTION;ERROR;Error"
       ENVIRONMENT
         K4GEO=${CMAKE_INSTALL_PREFIX}/share/k4geo
```
andresailer, line 7: This points to the install location, so it should probably be something else?
Also LD_LIBRARY_PATH is no longer being set?

test/CMakeLists.txt
```diff
@@ -1,199 +1,169 @@
...
     )
-
 endmacro()
 
+function(set_test_env testname)
+  message(WARNING ${PROJECT_BINARY_DIR})
```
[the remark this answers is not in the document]
jmcarcell: Removed

test/CMakeLists.txt
jmcarcell: I removed `K4GEO`, which seems not to be needed. 
`LD_LIBRARY_PATH` is set at the bottom for all tests. Maybe I can do the same for `FAIL_REGULAR_EXPRESSION` and remove the function

PR #510, opened 2025-09-10
Remove a default destructor that is not needed in DRparamBase_k4geo.h

jmcarcell:
BEGINRELEASENOTES
- Remove a default destructor that is not needed in `DRparamBase_k4geo.h`

ENDRELEASENOTES

Can be seen with Clang:

``` c
/k4geo/detectorSegmentations/include/detectorSegmentations/DRparamBase_k4geo.h:103:7: warning: definition of implicit copy constructor for 'shortFibers' is deprecated because it has a user-declared destructor [-Wdeprecated-copy-with-dtor]
  103 |       ~shortFibers() = default;
```

In this case it doesn't need to be defined, `default` is the default.

Tagging @SanghyunKo

### DIFF
--- a/detectorSegmentations/include/detectorSegmentations/DRparamBase_k4geo.h
+++ b/detectorSegmentations/include/detectorSegmentations/DRparamBase_k4geo.h
@@ -100,7 +100,6 @@ namespace DDSegmentation {
     struct shortFibers {
     public:
       shortFibers(const double towerH) : fTowerH(towerH) {}
-      ~shortFibers() = default;
 
       void addShortFibers(const int row, const int col, const double len) {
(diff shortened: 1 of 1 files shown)


SanghyunKo (approved):

SanghyunKo:
Have absolutely no problem with it

SanghyunKo:
Btw, I fear that the downstream test will fail today since I merged #486. This should be fixed once we test & merge HEP-FCC/k4RecCalorimeter#167 tomorrow...

jmcarcell:
It does fail as expected: https://github.com/key4hep/k4geo/actions/runs/17618023623/job/50056273528?pr=510

PR #511, opened 2025-09-12
Properly override neighbours() virtual methods.

scott-snyder:
Some segmentation classes defined a neighbours() method with a signature different from that defined in the Segmentation base class. In those cases, also override the base class signature in terms of what the derived class provides.

BEGINRELEASENOTES
- Make sure segmentation classes properly override the virtual neighbours() method for the FCCSWHcalPhi segmentations.

ENDRELEASENOTES

### DIFF
--- a/detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiRow_k4geo.h
+++ b/detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiRow_k4geo.h
@@ -74,6 +74,11 @@ namespace DDSegmentation {
      */
     void defineCellIndexes(const unsigned int layer) const;
 
+    /**  Find neighbours of the cell.
+     *   Implement the signature from the Segmentation base class.
+     */
+    virtual void neighbours(const CellID& cellID, std::set<CellID>& neighbours) const override;
+
     /**  Determine the azimuthal angle of HCal cell based on the cellID.
      *   @param[in] aCellId ID of a cell.
      *   return Phi.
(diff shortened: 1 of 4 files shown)


detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiRow_k4geo.h
```diff
@@ -74,6 +74,11 @@ namespace DDSegmentation {
      */
     void defineCellIndexes(const unsigned int layer) const;
 
+    /**  Find neighbours of the cell.
+     *   Implement the signature from the Segmentation base class.
+     */
+    virtual void neighbours(const CellID& cellID, std::set<CellID>& neighbours) const override;
+
```
tmadlener, line 81: Super nitpicky here, but could we move this up to the other definition of `neighbours`, so that they appear next to each other? I was just wondering what the other signature was to check if it's only marginally of, or whether it's completely different (which it seems to be).

detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiTheta_k4geo.h
```diff
@@ -70,6 +70,11 @@ namespace DDSegmentation {
      */
     void defineCellEdges(const unsigned int layer) const;
 
+    /**  Find neighbours of the cell.
+     *   Implement the signature from the Segmentation base class.
+     */
+    virtual void neighbours(const CellID& cellID, std::set<CellID>& neighbours) const override;
+
```
tmadlener, line 77: Same here.

tmadlener (commented):
Purely from a code point of view this looks reasonable, but I will let people more familiar with this judge whether using the default `Segmentation::neighbours` was intentional or not.

PR #512, opened 2025-09-12
Don't use exceptions to test for the presence of an attribute.

scott-snyder:
Exceptions should be exceptional: don't rely on exceptions for control flow for cases which are not actually errors.

BEGINRELEASENOTES
- Small code cleanup in a few files to remove the use of exceptions for normal control flow.

ENDRELEASENOTES


jmcarcell:
This is happening in a few more places:
```
detector/fcal/BeamCal_o1_v02_geo.cpp
158:        try {
159:          const std::string& sliceType = compSlice.attr<std::string>(_Unicode(layerType));

detector/fcal/BeamCal_o1_v01_geo.cpp
168:        try {
169:          const std::string& sliceType = compSlice.attr<std::string>(_Unicode(layerType));

detector/tracker/ZSegmentedPlanarTracker_geo.cpp
64:  try {
65:    isStripDetector = x_det.attr<bool>(_Unicode(isStripDetector));
```
so I have done the changes for these files.

PR #514, opened 2025-09-12
Fix overlaps in HCalTileBarrel

scott-snyder:
HCalTileBarrel has overlaps between the volumes representing the first (minimum z) row and the end plate.

Specifically, the geometry places the center of each row (also called a sequence, consisting of absorber+scintillator+spacing) by

```
      double zOffset = -dzDetector + (2 * numSeq + 1) * (dzSequence * 0.5);
```

where numSeq is the index of the row/sequence and dzSequence is the z-width of a row/sequence.  This can be rewritten probably more clearly as

```
      double zOffset = -dzDetector + numSeq*dzSequence + dzSequence/2;
```

dzDetector is calculated as

```
  double dzDetector = (numSequencesZ * dzSequence) / 2 + dZEndPlate + space;
```

so this is the total half-length of the barrel, including the sensitive area and the end plate.  But the placement above starts putting the rows right at -dzDetector ... overlapping the end plate.  Further, since (dZEndPlate+space) is not divisible by dzSequence, then the row edges do not line up with z=0, and they are asymmetric with respect to z inversion.

We can fix this by simply changing the offset calculation to

```
      double zOffset = -dzDetector + numSeq*dzSequence + dzSequence/2 + dZEndPlate + space;
```

This gets rid of the overlaps and makes the layout symmetric.

To be consistent, we also need to change the width_z parameter for the FCCSWHCalPhiRow_k4geo segmentation.  This was set to a hardcoded value of 2795.5*2*mm, which is just twice dzDetector.  So we should also subtract (twice) the end plate width + space.  Before this, the segmentation cell edges were also offset --- not lining up with 0 and asymmetric. With this change, they are symmetric.

This will of course change simulation results for the tile hcal barrel.

BEGINRELEASENOTES
- Fix overlaps in the tile hcal barrel.
ENDRELEASENOTES

### DIFF
--- a/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/HCalBarrel_TileCal_v03.xml
+++ b/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/HCalBarrel_TileCal_v03.xml
@@ -31,6 +31,9 @@
     <constant name="BarHCal_layer_size1" value="50*mm" />
     <constant name="BarHCal_layer_size2" value="100*mm" />
     <constant name="BarHCal_layer_size3" value="200*mm" />
+    <!-- division into rows -->
+    <constant name="BarHCal_sequence_thickness" value="2*BarHCal_master_plate_thickness + BarHCal_spacer_plate_thickness + BarHCal_scintillator_thickness + 2*BarHCal_air_space_thickness" />
+    <constant name="BarHCal_numSequencesZ" value="floor(2*(BarHCal_dz-BarHCal_end_plate_thickness-BarHCal_plate_space)/BarHCal_sequence_thickness)" />
   </define>
 
   <display>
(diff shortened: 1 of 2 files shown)


scott-snyder:
Mark as draft while investigating test failure.

scott-snyder:
Test failure was not reproducible.

giovannimarchiori:
@Archil-AD could you please have a look?
Thanks!

scott-snyder:
You're right, thanks for pointing that out.
It makes rather less difference for the theta segmentation of course.
There only changes the extent of the outermost cells, while for the
row segmentation, changing the width shifts the entire grid.

Anyway, that's changed.

giovannimarchiori:
Hi @scott-snyder 
is there a way to replace the hardcoded 2795.5 with something derived from BarHCal_dz (2800) and the other detector dimensions? or maybe at least replace 2795.5 with BarHCal_dz - BarHCal_<xxx> (xxx = gap? margin? ) so that the code is more readable and if we change the dimensions in DectDimensions.xml these will propagate here?

giovannimarchiori (approved):
OK, PR looks good to me - after merging please open a new small PR to read the number of sequences from the compact file
Thanks a lot!

PR #517, opened 2025-09-25
Some segmentation cleanup.

scott-snyder:
- Avoid directly using protected members from base classes when there is a public accessor.  Change protected members to private.
 - Remember field indices, to avoid having to always do string lookups.
 - Factor out duplicated code between constructors.
 - Naming convention consistency (use m_ for class data members).
 - Prefer to pass simple types by value, and to return vector members as const references.
 - Use override keywords.

BEGINRELEASENOTES
- Code cleanups in the segmentation classes.
ENDRELEASENOTES

### DIFF
--- a/detectorSegmentations/include/detectorSegmentations/FCCSWGridPhiEtaHandle_k4geo.h
+++ b/detectorSegmentations/include/detectorSegmentations/FCCSWGridPhiEtaHandle_k4geo.h
@@ -63,10 +63,10 @@ public:
   /// Equality operator
   bool operator==(const FCCSWGridPhiEta_k4geo& seg) const { return m_element == seg.m_element; }
   /// determine the position based on the cell ID
-  inline Position position(const CellID& id) const { return Position(access()->implementation->position(id)); }
+  inline Position position(const CellID id) const { return Position(access()->implementation->position(id)); }
 
   /// determine the cell ID based on the position
-  inline dd4hep::CellID cellID(const Position& local, const Position& global, const VolumeID& volID) const {
+  inline dd4hep::CellID cellID(const Position& local, const Position& global, const VolumeID volID) const {
     return access()->implementation->cellID(local, global, volID);
   }
 
@@ -109,7 +109,7 @@ public:
       -# size in phi
       -# size in eta
   */
-  inline std::vector<double> cellDimensions(const CellID& /*id*/) const {
+  inline std::vector<double> cellDimensions(const CellID /*id*/) const {
     return {access()->implementation->gridSizePhi(), access()->implementation->gridSizeEta()};
   }
 };
--- a/detectorSegmentations/include/detectorSegmentations/GridDRcalo_k4geo.h
+++ b/detectorSegmentations/include/detectorSegmentations/GridDRcalo_k4geo.h
@@ -20,7 +20,7 @@ namespace DDSegmentation {
     // Determine the global (local) position based on the cell ID.
     // position of the front end of the fiber
     virtual Vector3D position(const CellID& aCellID) const override;
-    Vector3D localPosition(const CellID& aCellID) const;
+    Vector3D localPosition(const CellID aCellID) const;
     Vector3D localPosition(int numx, int numy, int x_, int y_) const;
 
     // position of the rear end (readout location) of the fiber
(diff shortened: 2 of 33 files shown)


detectorSegmentations/include/detectorSegmentations/GridDRcalo_k4geo.h
```diff
@@ -32,91 +32,110 @@ namespace DDSegmentation {
...
-    int y(const int& aId32) const { return y(convertLast32to64(aId32)); }
+    int x(const int aId32) const { return x(convertLast32to64(aId32)); }
+    int y(const int aId32) const { return y(convertLast32to64(aId32)); }
 
-    bool IsCerenkov(const int& aId32) const { return IsCerenkov(convertLast32to64(aId32)); }
+    bool IsCerenkov(const int aId32) const { return IsCerenkov(convertLast32to64(aId32)); }
 
-    bool IsTower(const int& aId32) const { return IsTower(convertLast32to64(aId32)); }
+    bool IsTower(const int aId32) const { return IsTower(convertLast32to64(aId32)); }
     bool IsSiPM(const int& aId32) const { return IsSiPM(convertLast32to64(aId32)); }
```
andresailer, line 80:
```suggestion
    bool IsSiPM(const int aId32) const { return IsSiPM(convertLast32to64(aId32)); }
```
?

detectorSegmentations/src/GridTheta_k4geo.cpp
```diff
@@ -4,19 +4,12 @@ namespace dd4hep {
...
-    registerIdentifier("identifier_theta", "Cell ID identifier for theta", m_thetaID, "theta");
-  }
+  GridTheta_k4geo::GridTheta_k4geo(const BitFieldCoder* decoder) : Segmentation(decoder) { commonSetup(); }
 
-  GridTheta_k4geo::GridTheta_k4geo(const BitFieldCoder* decoder) : Segmentation(decoder) {
+  /// Initialization common to all ctors.
+  void GridTheta_k4geo::commonSetup() {
     // define type and description
     _type = "GridTheta_k4geo";
     _description = "Etheta segmentation in the global coordinates";
```
andresailer, line 15:
```suggestion
    _description = "Theta segmentation in the global coordinates";
```
?

detectorSegmentations/include/detectorSegmentations/FCCSWGridPhiEtaHandle_k4geo.h
```diff
@@ -109,7 +109,7 @@ class FCCSWGridPhiEta_k4geo : public FCCSWGridPhiEtaHandle_k4geo {
       -# size in phi
       -# size in eta
   */
-  inline std::vector<double> cellDimensions(const CellID& /*id*/) const {
+  inline std::vector<double> cellDimensions(const CellID /*id*/) const {
```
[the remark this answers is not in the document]
scott-snyder: No... this class derives from Handle<...> not Segmentation, so this method is not virtual.

giovannimarchiori:
Thumbs up from my side!

jmcarcell (commented):
Two comments that are not urgent (so merge if you want):
- Several member getters could return a `const` object (for example `Phi()` or `phi()` returning a `double` instead of a `const double`
- There are two members that could by statically allocated in `GridDRcalo_k4geo.cpp`: https://github.com/key4hep/k4geo/blob/main/detectorSegmentations/src/GridDRcalo_k4geo.cpp#L26-L27

tmadlener:
Merging this to make a new tag after the nightlies have passed.

scott-snyder:
> Two comments that are not urgent (so merge if you want):

Thanks for the comments.

>     * Several member getters could return a `const` object (for example `Phi()` or `phi()` returning a `double` instead of a `const double`

That actually makes no difference.  (cv-qualifiers are ignored for prvalues that aren't class or array types.)

>     * There are two members that could by statically allocated in `GridDRcalo_k4geo.cpp`, then the destructor doesn't need to be defined: https://github.com/key4hep/k4geo/blob/550e75e33069420c92d9bbd87d35a6538e92fd32/detectorSegmentations/src/GridDRcalo_k4geo.cpp#L26-L27

You're right that that bit can/should be cleaned up further, but it's a bit more involved than you suggest due to constness issues.  It's anyway wrong the way it is; i'll take care of this in a future PR.

PR #519, opened 2025-10-02
Add pixels tracker

jessy-daniel:
BEGINRELEASENOTES
- Add a pixel/strips segmentation in vertex and tracking detectors. This modification is necessary to implement a detailled tracking digitizer which needs the info on hit pixels and strips. 

ENDRELEASENOTES

### DIFF
--- a/detector/tracker/TrackerEndcap_o2_v06_geo.cpp
+++ b/detector/tracker/TrackerEndcap_o2_v06_geo.cpp
@@ -312,7 +312,6 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
     thisLayer.zPosition = sumZ / ring_no; // calc average z
     thisLayer.distanceSensitive = innerR;
     thisLayer.lengthSensitive = outerR - innerR;
-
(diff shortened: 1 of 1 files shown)


Zehvogel (changes requested):
Thanks, I think overall these changes look correct :)

My main question is: does the current default `CLDConfig/CLDReconstruction.py` still work as before? (I would expect so but one can never be sure...)

In addition, I have some change requests regarding the form:

Please fix the indentation in basically every line you touched to conform to the rest of the file (misconfigured editor?). Like it is at the moment, even some lines where the content did not change show up in the diff :(

Also, this is not quite the correct spot to make these changes. Adhering to our usual policy, you would have to make a `CLD_o2_v09` for these changes **and also bump up the version numbers of the changed files**. Maybe we could also just put them into `CLD_o2_v08` as this model was not used much yet, but I would be against putting them into `CLD_o2_v07` now.

andresailer:
There is a discussion [here](https://github.com/key4hep/k4RecTracker/pull/64#discussion_r2398673687) that might make these changes obsolete

PR #527, opened 2025-10-13
FCC MDI: new CAD model with rotated notch

aciarma:
BEGINRELEASENOTES

CAD version of MDI had the crotch rotated 180deg, so that the notch for the SR mask was on the wrong side.
The SR mask is only on the lower part of the upstream beampipes - negative x coordinates.

Created new folder MDI_o1_CADBased_v01 with the following fixes:
- fixed orientation of crotch beam pipe
- added MDI_standalone_v01.xml files to ease debug of MDI geometries

New naming convention decided:
- MDI_o1_ShapeBased_vXX for shape based geometries
- MDI_o1_CADBased_vXX    for CAD based geometries

ENDRELEASENOTES

OLD version of MDI_o1_v01
image

FIXED version of MDI_o1_CADBased_v01
image

### DIFF
--- a/FCCee/IDEA/compact/IDEA_o1_v02/BeamInstrumentation_o1_v01.xml
+++ b/FCCee/IDEA/compact/IDEA_o1_v02/BeamInstrumentation_o1_v01.xml
@@ -6,7 +6,7 @@
         url="no"
         status="development"
         version="1.0">
-    <comment>COmpensating and screening solenoids for FCCee</comment>
+    <comment>Compensating and screening solenoids for FCCee</comment>
   </info>
 
     


BrieucF:
Thanks Andrea, can you refactor the RELEASE NOTE to reflect the fact that you created a new MDI folder with these fixes? Maybe also better to put the pictures outside of the release note, not sure how they would display there, maybe someone knows.

BrieucF (commented):
Can you confirm that everything runs when e.g. you replace [these](https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml#L44-L45) lines with the new path (i.e. not using the standalone geometry)?

PR #534, opened 2025-10-30
Started MDI_o1_ShapeBased_v01 and add FFQuads to MDI_CADbased_o1_v01

aciarma:
BEGINRELEASENOTES
- [FCCee] MDI_o1_ShapeBased_v01 started from MDI_o1_v00 
- [FCCee] added antisolenoid field map to ShapeBased
- [FCCee] added final focus quadrupoles and their field to ShapeBased. Dimension and magnetic field defined via parameter xml based on different accelerator lattice versions and beam energy
- [FCCee] add quadrupoles also to MDI_CADbased_o1_v01"
ENDRELEASENOTES

### DIFF
--- a/FCCee/MDI/compact/MDI_o1_ShapeBased_v01/FFQuads_params_v572_Z.xml
+++ b/FCCee/MDI/compact/MDI_o1_ShapeBased_v01/FFQuads_params_v572_Z.xml
@@ -10,6 +10,7 @@
 
   
   <define>
+    <constant name="FFQ_ON" value="1"/> <!-- hardcoded only for now -->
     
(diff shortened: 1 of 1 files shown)


aciarma:
> When loading this from the detector side, it crashes because `FFQ_ON` is not defined, can this be fixed so that uncommenting:
> 
> ```
> <include ref="../../../MDI/compact/MDI_o1_CADBased_v01/MDI_o1_CADBased_v01_collection.xml" />
> ```
> 
> in the detector xml works out of the box?

it was intended to be like this but ok I hardcoded the flag into collection.xml.

PR #535, opened 2025-10-30
Update HCal segmentation classes to use vectors to store cell edge information.

scott-snyder:
Update FCCSWHCalPhiTheta_k4geo and FCCSWHCalPhiRow_k4geo to store cell edge information in vectors rather than hash maps.

Verified that simulation results are unchanged.

BEGINRELEASENOTES
- Updated HCal segmentation classes to more efficiently store cell edge information as vectors rather than as hash maps.
ENDRELEASENOTES

### DIFF
--- a/detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiRow_k4geo.h
+++ b/detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiRow_k4geo.h
@@ -329,6 +329,7 @@ namespace DDSegmentation {
           if (-idx < m_ibin)
             throw std::out_of_range("cellEdge");
           const auto& e = m_cellEdge[-idx - m_ibin];
+          // Swap the edges for the negative endcap side
           return Edge{e.high, e.low};
         }
       }


detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiTheta_k4geo.h
```diff
@@ -265,8 +265,42 @@ namespace DDSegmentation {
...
+      // which handle the lookup by bin number, throwing std::out_of_range
+      // for a nonexistent bin.
+      using Edges = std::pair<double, double>;
+      int m_ibin1 = 0;
+      int m_ibin2 = 9999999;
+
+      struct CellInfo {
+        CellInfo(double lo, double hi) : edges(lo, hi) {}
+        Edges edges{0, 0};
+      };
```
tmadlener: Is this type really necessary? Everything it does seems to be covered by `Edges` already, why not simply skip one of them?

detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiTheta_k4geo.h
```diff
@@ -265,8 +265,42 @@ namespace DDSegmentation {
...
+      std::vector<CellInfo> m_cellInfo1{};
+      std::vector<CellInfo> m_cellInfo2{};
+
+      const CellInfo& cellInfo(int ibin) const {
+        if (ibin < m_ibin1)
+          throw std::out_of_range("cellInfo");
+        if (ibin < m_ibin2)
+          return m_cellInfo1.at(ibin - m_ibin1);
+        return m_cellInfo2.at(ibin - m_ibin2);
```
tmadlener: I assume using `.at` here is desired over simply using `operator[]` because `ibin - m_ibinX` can still be out of range and we want the range check?

tmadlener (commented):
What is the technical reason for going from a map to a vector? Simply the fact that the vector can be structured such that a lookup can be done by index and is thus faster?

A few minor nits about naming and types below. (None of which I would really consider blocking)

detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiTheta_k4geo.h
```diff
@@ -265,8 +265,42 @@ namespace DDSegmentation {
...
+      // the first entry of the vector corresponds to bin number m_ibin1.
+      // For the endcap, we have two disjoint bin ranges, corresponding
+      // to the positive and negative endcaps.  In that case, the two
+      // ranges are stored in m_cellInfo1 and m_cellInfo2; m_ibin1
+      // gives the bin number of the first entry of m_cellInfo1 and
+      // m_ibin2 gives the bin number of the first entry of m_cellInfo2.
+      // To access the information, use the cellInfo() functions,
+      // which handle the lookup by bin number, throwing std::out_of_range
+      // for a nonexistent bin.
+      using Edges = std::pair<double, double>;
```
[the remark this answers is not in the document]
scott-snyder: Ok, i'll give that a try.

detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiRow_k4geo.h
```diff
@@ -310,8 +310,24 @@ namespace DDSegmentation {
...
+      std::pair<double, double> cellEdges(int idx) const {
+        if (idx > 0) {
+          if (idx < m_ibin)
+            throw std::out_of_range("cellEdges");
+          return m_cellEdges[idx - m_ibin];
+        } else {
+          if (-idx < m_ibin)
+            throw std::out_of_range("cellEdges");
+          const auto& e = m_cellEdges[-idx - m_ibin];
+          return std::make_pair(e.second, e.first);
```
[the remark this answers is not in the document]
scott-snyder: No, that change doesn't work.  This is for the negative side, so we need to swap the edges.

detectorSegmentations/include/detectorSegmentations/FCCSWHCalPhiTheta_k4geo.h
scott-snyder: Yeah, and i've been trying to train myself to write at() by default.

PR #547, opened 2025-11-20
[IDEA] Adopting Muon system for IDEA_o2

mahmoudali2:
- Automating Muon system endcap outer radius, Just like IDEA_o1.
- Changes on the builder have been already merged before in #526


jmcarcell:
I can confirm this constant is not being used anywhere else in k4geo, so this shouldn't change any behaviour.

PR #549, opened 2025-12-04
Introduce layer DetElements to Tracker and Vertex endcap constructors

### LINKED ISSUE #548: TrackerEndcap_o2_v06 does not create a layer DetElement
The `TrackerEndcap_o2_v06` (and almost certainly most of its predecessors) do not put the modules in each layer into a separate Assembly and a layer `DetElement`. The `TrackerBarrel_o1_v05` on the other hand does just this. Having a layer `DetElement` would make it much more straight forward to walk the created geometry tree and would greatly facilitate the conversion of the DD4hep geometry to Acts, as currently being done on https://github.com/key4hep/k4ActsTracking/pull/36

Is there a specific reason why the modules are placed directly into the Endcap `DetElement` rather than creating `DetElement`s for each layer.

tmadlener:
Without this all modules will be placed directly into the top level DetElement of the TrackerEndcap which makes it hard to walk this sub-detector on a layer by layer basis.

BEGINRELEASENOTES
- Introduce new versions of `TrackerEndcap_o2_v07` and `VertexEndcap_o1_v07` based on the corresponding `v06` versions.
  - Introduce a `layer_pos` (and `layer_neg` if `reflect == True`) `DetElement` and an assembly volume to make the resulting subdetector more hierarchical. This allows one to more easily walk the geometry tree as now the modules are placed into dedicated layers instead of being just placed into the top level sub detector volume.
- Use the new versions for the `MAIA_v0` geometry.
- **Other geometries that want this behavior have to actively opt-in!**

ENDRELEASENOTES

Fixes #548 

I haven't tested this extremely far yet, but I can walk the geometry tree a bit better now and it solves the problem that I originally set out to solve.

- [x] Check where this constructor is used
- [x] Validated things work as expected and unchanged after these changes

### DIFF
--- a/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/HCalBarrel_TileCal_v03.xml
+++ b/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/HCalBarrel_TileCal_v03.xml
@@ -92,6 +92,7 @@
            dRlayer - all different dRs/thicknesses used to form the layers.
            grid_size_row - takes 13 numbers which equals to the number of layers in the Barrel;
                            each value is the number of rows/scintillators grouped in a cell in each layer.
+           grouped_rows - this is not used for the Barrel.
            dz_row - size of the row consisting of 2 * Master plate + 1 * Spacer plate + 1 * Scintillator + Air.
            phi_bins - number of bins in phi
            offset_phi - center of the first phi bin.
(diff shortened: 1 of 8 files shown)


detector/tracker/TrackerEndcap_o2_v06_geo.cpp
```diff
@@ -155,6 +155,12 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
     double sumZ(0.), innerR(1e100), outerR(0.);
     double sensitiveThickness(0.0);
 
+    Assembly lay_vol_pos(_toString(x_layer.id(), "layer_pos%d"));
+    Assembly lay_vol_neg(_toString(x_layer.id(), "layer_neg%d"));
+
+    DetElement lay_elt_pos(sdet, _toString(x_layer.id(), "layer_pos%d"), l_id);
+    DetElement lay_elt_neg(sdet, _toString(x_layer.id(), "layer_neg%d"), l_id);
```
andresailer:
```suggestion
    Assembly lay_vol_pos(_toString(l_id, "layer_pos%d"));
    Assembly lay_vol_neg(_toString(l_id, "layer_neg%d"));

    DetElement lay_elt_pos(sdet, _toString(l_id, "layer_pos%d"), l_id);
    DetElement lay_elt_neg(sdet, _toString(l_id, "layer_neg%d"), l_id);
```

tmadlener:
Anyone against merging this? I would like to keep this topical, instead of piling on more and more detectors that need fixing in order to make the conversion to Acts easier to do.

Zehvogel (approved):
LGTM 👍

PR #551, opened 2025-12-23
Fix TPCAction

danieljeans:
- [n/a] the relevant README(s) were updated. See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md

BEGINRELEASENOTES
- Fix the logic of TPCAction: information of previous G4 steps was not correctly accessed.
ENDRELEASENOTES

### DIFF
--- a/detector/tracker/README.md
+++ b/detector/tracker/README.md
@@ -68,4 +68,9 @@ Initial design for a tracker based on thin-wall straw tubes.  This will likely b
  - 200MB memory usage for straw detector.
  - Stereo angle is implemented, is +/- 2 degrees in ALLEGRO_o2_v01
  - No consideration yet to endcap.
- 
+
+## TPC
+
+### TPC10
+Scalable TPC model, intended to be used with plugins/TPCSDAction to combine G4 steps into SimTrackerHits.
+TPCSDaction has some ingrained assumptions that G4 steps are passed to it in a logical way, i.e. in the order that they are produced.


plugins/TPCSDAction.cpp
```diff
@@ -55,9 +54,23 @@ namespace sim {
     G4ThreeVector CumulativeMeanMomentum{};
     G4int CumulativeNumSteps{};
 
-    G4Step const* previousStep{};
     std::map<int, G4double> padRowCentralRadii{};
 
+    struct stepInfo {
+      bool isValid{};
```
tmadlener:
```suggestion
      bool isValid{false};
```
I would assume that should be the default (and also is IIRC), but this way people would not have to guess and we are not at the mercy of the compiler never changing.

plugins/TPCSDAction.cpp
```diff
@@ -275,36 +290,38 @@ namespace sim {
...
+      stepAtEntranceToPadRingInfo.isValid = false;
     }
 
-    void DepositHiPtHit(G4Step const* step) // DJ extracted to separate fn
-    {
+    void DepositHiPtHit(stepInfo mainStepInfo) {
+
+      if (!mainStepInfo.isValid || !stepAtEntranceToPadRingInfo.isValid) {
+        std::cerr << "DepositHiPtHit ERROR, invalid stepInfo " << mainStepInfo.isValid << " "
+                  << stepAtEntranceToPadRingInfo.isValid << std::endl;
```
tmadlener: Can we go via some dd4hep logging facility for this? I think we should have rather simple access via the `sensitive`: https://dd4hep.web.cern.ch/dd4hep/reference/classdd4hep_1_1sim_1_1Geant4Action.html#a1abdae5d2d2af327f406697322b0d821

Otherwise, we can always go via `dd4hep::printout`

plugins/TPCSDAction.cpp
```diff
@@ -342,19 +357,18 @@ namespace sim {
...
     }
 
-    void DepositLowPtHit(G4Step const* step) {
-
-      Geant4Tracker::Hit* hit = new Geant4Tracker::Hit(
-          step->GetTrack()->GetTrackID(), step->GetTrack()->GetDefinition()->GetPDGEncoding(), CumulativeEnergyDeposit,
-          CumulativeMeanTime / CumulativeNumSteps); // DJ : simple average of all steps' time
+    void DepositLowPtHit(stepInfo stInf) {
+      if (!stInf.isValid) {
+        std::cerr << "DepositLowPtHit ERROR, invalid step info" << std::endl;
```
tmadlener: Also here, it would be nice to go through DD4hep logging facilities.

tmadlener (commented):
This looks like a fairly mechanical change once the `StepInfo` helper class is introduced. I have only minor comments here.

> I don't know if this has always been an issue, or if the the way of passing steps to the plugin action has changed.
I would like to think that it was working as intended when I wrote it in early 2023, but cannot guarantee it...

It could also be a change in Geant4, where they have started to re-use steps at some point rather than discarding them and creating new ones. I haven't checked the release notes of Geant4 nor DD4hep to be sure though. In any case this looks like a more robust solution.

plugins/TPCSDAction.cpp
```diff
@@ -112,7 +131,7 @@ namespace sim {
       fLowPtHitCollection = sensitive->collection(2);
 
       Geant4StepHandler h(step);
-      //	dumpStep( h , step ) ;
+      dumpStep(h, step);
```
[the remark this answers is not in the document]
danieljeans: yes, since this is silent unless we're in debug, I think it's OK to keep. It may be useful to dump step info without recompiling...

PR #552, opened 2025-12-24
SEcal05_ECRing: more consistent use of derived dimensions; remove overlap in ILC QD0 cryostat

danieljeans:

BEGINRELEASENOTES
- this code previously used a mixture of dimensions calculated in the xml description, and the more basic parameters from which they are calculated. Now use those calculated in the compact description, explicitly remove others. 
- remove duplicated material in QD0 cryostat (remove qd0_cryostat, duplicate of qd0_cryostat_wall).
ENDRELEASENOTES

### DIFF
--- a/ILD/compact/ILD_common_v02/SEcal06_siw_ECRing.xml
+++ b/ILD/compact/ILD_common_v02/SEcal06_siw_ECRing.xml
@@ -1,0 +1,83 @@
+<lccdd>
+
+  <detectors>
+    <detector name="EcalPlug" type="SEcal06_ECRing" id="ILDDetID_ECAL_PLUG" readout="EcalEndcapRingCollection" vis="ILD_BlueVis" >
+      <comment>EM Calorimeter Endcaps</comment>
+
+      <envelope vis="ILD_ECALVis">
+        <shape type="BooleanShape" operation="Subtraction" material="Air">
+          <shape type="BooleanShape" operation="Subtraction" material="Air">
+            <shape type="Box" dx="EcalEndcapRing_outer_radius + env_safety" dy="EcalEndcapRing_outer_radius + env_safety"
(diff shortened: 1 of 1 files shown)


danieljeans:
This is needed for FCCee models: at ILC the inner radius of the ECring is tied to the lumical, which sits within it. At FCCee this is no longer the case. Dependence on the Lumical dimensions is therefore no longer appropriate.

No effect on the ILD@ILC models.

danieljeans:
to make this SEcal05_ECRing driver cleaner and less confusing, I explicitly introduce the offset of the central hole as an input to the driver, rather than internally calculating from the crossing angle. The offset is now calculated in the xml detector description.

tmadlener:
I think technically this should constitute a bump in the version of the driver. However, looking into k4geo it looks like ILD is the only customer so far

```
FCCee/ILD_FCCee/compact/ILD_common_FCCee/SEcal06_siw_ECRing.xml
4:    <detector name="EcalPlug" type="SEcal05_ECRing" id="ILDDetID_ECAL_PLUG" readout="EcalEndcapRingCollection" vis="ILD_BlueVis" >

ILD/compact/ILD_common_v02/SEcal05_siw_ECRing.xml
4:    <detector name="EcalPlug" type="SEcal05_ECRing" id="ILDDetID_ECAL_PLUG" readout="EcalEndcapRingCollection" vis="ILD_BlueVis" >

ILD/compact/ILD_common_v01/SEcal05_siw_ECRing.xml
1:<detector name="EcalPlug" type="SEcal05_ECRing" id="ILDDetID_ECAL_PLUG" readout="EcalEndcapRingCollection" vis="BlueVis" >
```

(Curiously, the FCCee model already states `06` in the xml configuration)

PR #556, opened 2026-01-29
fix crash when cryo of noble liquid calorimeter  is made active

giovannimarchiori:
This MR fixes a crash when the cryostat of the noble liquid calorimeter is made active for special purposes (e.g. to study energy lost upstream), since the segmentation class try to calculate radial positions of "layers" into the sensitive volume of the cryogenic that do not exist.
The error message is
```
VolumeManager    ERROR lookupContext: Failed to search Volume context 0000000000000834 [Unknown identifier]
DDSim            ERROR Simulation failed!
Geant4Kernel     FATAL +++ Exception while simulating:VolumeManager: lookupContext: Failed to search Volume context 0000000000000834 [Unknown identifier]
```
and comes from 
https://github.com/AIDASoft/DD4hep/blob/e9165405b80fd044ee3c16880e2d62c595f47de3/DDCore/src/VolumeManager.cpp#L626

As a workaround, the part of the code that calculates those positions (that is used to calculate the cell position itself) is skipped if the cell is in the cryostat, and its position is set to 0,0,0 (only energies are used for these special cells AFAIK).

BEGINRELEASENOTES
- fix crash when cryo of noble liquid calorimeter  is made active
ENDRELEASENOTES

Tagging @BrieucF  and @scott-snyder (the crash was introduced by https://github.com/key4hep/k4geo/commit/c789de28596c635ef789822cc6c5567a909173d9)

### DIFF
--- a/detectorSegmentations/src/FCCSWGridModuleThetaMerged_k4geo.cpp
+++ b/detectorSegmentations/src/FCCSWGridModuleThetaMerged_k4geo.cpp
@@ -213,7 +213,8 @@ namespace DDSegmentation {
     double phi = (m_mergedModules[layer] - 1) * M_PI / m_nModules;
 
     // debug
-    // dd4hep::printout(dd4hep::VERBOSE, "FCCSWGridModuleThetaMerged_k4geo", "layer = %d, merged modules = %d, phi = %lf",
+    // dd4hep::printout(dd4hep::VERBOSE, "FCCSWGridModuleThetaMerged_k4geo", "layer = %d, merged modules = %d, phi =
+    // %lf",
     //                  layer, m_mergedModules[layer], phi);
 
     return phi;


scott-snyder:
hi -

Thanks for pointing this out.

This is ok for now, but ultimately we probably do want to give a sensible
position for these --- event displays would use that, if nothing else.

More specifically:

 - It would probably be better to put the printout changes into a separate
   commit/MR.  Further, i don't think we want to use dd4hep::printout
   in places that are called for each cell (position, and maybe phi/theta),
   as that gives an unavoidable out-of-line function call, in addition
   to any evaluation of the options.

 - I'd write your test like

```
    // skip cells in cryo
    if (decoder()->get(cID, m_cryoIndex) != 0)
      return Vector3D(0, 0, 0);
```

   to avoid additional indentation for the rest of the function.
   Also, the field index should be cached, rather than doing a string lookup
   each time.

 - For my information, can you provide instructions to reproduce?
   I'm in the middle of some changes to make cellid handling more
   efficient in the reconstruction, and would like to ensure that this
   is not going to result in any issues there.

giovannimarchiori:
Hi @scott-snyder 

> It would probably be better to put the printout changes into a separate
> commit/MR. Further, i don't think we want to use dd4hep::printout
> in places that are called for each cell (position, and maybe phi/theta),
> as that gives an unavoidable out-of-line function call, in addition
> to any evaluation of the options.

I commented the dd4hep::printout in places that are called for each cell 
I would leave the few remaining uncommented ones in this MR ..

> I'd write your test like
> 
>     // skip cells in cryo
>     if (decoder()->get(cID, m_cryoIndex) != 0)
>       return Vector3D(0, 0, 0);
> to avoid additional indentation for the rest of the function.

done

> Also, the field index should be cached, rather than doing a string lookup
> each time.

done

> For my information, can you provide instructions to reproduce?
> I'm in the middle of some changes to make cellid handling more
> efficient in the reconstruction, and would like to ensure that this
> is not going to result in any issues there.

Run the simulation after having set the cryo to sensitive in the compact file for the ECal e.g.
```
     <cryostat name="ECAL_Cryo">
        <material name="Aluminum"/>
        <dimensions rmin1="BarCryoECal_rmin" rmin2="BarCryoECal_rmin+CryoBarrelFront" rmax1="BarCryoECal_rmax-CryoBarrelBack" rmax2="BarCryoECal_rmax" dz="BarCryoECal_dz"/>
        <front sensitive="true"/> <!-- inner wall of the cryostat -->
        <side sensitive="true"/> <!-- both sides of the cryostat -->
        <back sensitive="true"/> <!-- outer wall of the cryostat -->
...
```

PR #557, opened 2026-02-05
Curved vertex features, more realistic vertex sensor description (sensitive thickness) and more efficient and detailed silicon wrapper

armin-ilg:
- [x] the code is covered by tests. See https://github.com/key4hep/k4geo/blob/main/test/CMakeLists.txt -> The tests for IDEA_o1_v03, IDEA_o2_v01, ALLEGRO_o1_v03, and ALLEGRO_o2_v01 don't work yet since in the nightlies the old silicon wrapper (whos name was changed) is referenced. I added a test for IDEA_o1_v04 however which works: 
[IDEA_o1_v04_test.log](https://github.com/user-attachments/files/25188564/IDEA_o1_v04_test.log)

BEGINRELEASENOTES
- Update of IDEA/ALLEGRO vertex detector to enable having part of the sensor insensitive (in thickness). The ultra-light, curved vertex detector can now be approximated by trapezoidal volumes, allowing track reconstruction with conformal tracking.
- Update of IDEA/ALLEGRO silicon wrapper: Only two layers and disks per side, leading to at least one hit almost everywhere. 4x4 cm^2 large sensors, both in barrel and disks. Simplification of disk design featuring ring-like module arrangement. Ferris-wheel geometry of silicon wrapper barrel supporting modules from and outside support shell, which is sitting on the drift chamber outer wall. 
ENDRELEASENOTES

Hi 

This is a large PR for a new version of the non-curved and ultra-light curved vertex detector and silicon wrapper for IDEA/ALLEGRO. 

Here's how the trapezoid-approximated ultra-light vertex looks like:
Curved_vertex_segments

The sensor is described by 40 µm of insensitive silicon, 10 µm of sensitive silicon, and 16 µm of insensitive silicon in this order inside out (properties of TPSCo 65 nm sensors used for ALICE ITS3). Here's how this looks like in the volume hierarchy:
Insensitive_thicknesses

The changes to the silicon wrapper are extensive, here are some pictures to show the changes:
SiWr envelope:
SiWr_envelope

SiWr showign only layer-level with supports:
Layer_volumes_with_supports

SiWr showign only layer-level without supports:
Layer_volumes_without_supports

The SiWr Barrel with the ferris-wheel layout
SiWrB_zoom-in

and showing the staves in detail:
SiWrB_zoom-in_detailed

One SiWr barrel stave from the front:
SiWrB_stave_front

and from the back:
SiWrB_stave_back

A SiWr disk ring:
SiWrD_ring

and in detail showing the modules/tiles from the front:
SiWrD_ring_detailed_front

and the back:
SiWrD_ring_detailed_back

And showing that there are no holes (except the 0.2 mm gap between neighbouring sensors) in the disk:
SiWrB_front-view_no_holes

Notes:
- I updated the naming of the silicon wrapper xml files so that the first version of the silicon wrapper is o1_v01 and not o1_v03 (as it was before). The new version of this PR is o1_v02. Please let me know if you don't want this change. This is also why the tests currently fail as the nightlies use the old naming scheme of the silicon wrapper, not finding o1_v03 in this PR.


tmadlener:
Surfacing this from a commit message for better visibility as it will probably require some thought (mid- to long term).

> SiliconWrapper must have different ReadoutID than vertex because it has many more modules

Currently all tracking detectors have the same readout id (globally). There are some differences between the different detector concepts, but inside a concept at least currently the assumption is that all of the tracking detectors have the same id.

I don't think that we should have the discussion in this PR, I am merely (ab)using it as an anchor for it for now.

armin-ilg:
> Surfacing this from a commit message for better visibility as it will probably require some thought (mid- to long term).
> 
> > SiliconWrapper must have different ReadoutID than vertex because it has many more modules
> 
> Currently all tracking detectors have the same readout id (globally). There are some differences between the different detector concepts, but inside a concept at least currently the assumption is that all of the tracking detectors have the same id.
> 
> I don't think that we should have the discussion in this PR, I am merely (ab)using it as an anchor for it for now.

Hey Thomas

Yes, at some point that discussion should be had. I don't think long-term this is feasible with increasing granularity of the various tracking detectors in full sim. Also the drift chamber actually has it's own definition (see https://github.com/key4hep/k4geo/blob/9df727cab1881ff342793969813edd692ebad982/FCCee/IDEA/compact/IDEA_o1_v03/DriftChamber_o1_v02.xml#L131). 

I, however, now (commit 10d1aad) changed the silicon wrapper to have more sensors per module (which is anyway benefitial for performance) and thus managed to keep the same GlobalTrackerReadoutID as before.

armin-ilg:
I rebased the PR branch now with:
```
git pull --ff-only k4geo main

git checkout <your-pr-branch>
git rebase k4geo/main

push --force-with-lease origin <your-pr-branch>
```
Let me know if I should have done something differently.

tmadlener:
Rebase procedure looks perfectly fine.

armin-ilg:
Ready for review now :-)

FCCee/IDEA/compact/IDEA_o1_v04/VertexComplete_IDEA_o1_v04.xml
```diff
@@ -0,0 +1,1974 @@
+<!-- Changes :
+   o1_v03 : IDEA/ALLEGRO vertex detector described in FCC Feasibility Study Final Report (2025). An ultra-light inner vertex detector design with four layers is included as well, enabled by setting "nInnerVertexLayers" to 4, disabling the layers below "Vertex inner barrel" and instead enabling the ultra-light layers below "Ultra-light inner vertex"
+   o1_v04 : Reduced sensitive thickness for ultra-light curved vertex. Removed 'auxiliary pads' in the ultra-light curved vertex to increase the area that is sensitive by about 2.5% (see ALICE IT3 TDR). Ultra-light curved vertex detector sensors are now approximated by trapezoidal volumes, enabling the use of conformal tracking that currently requires non-bent surfaces.
+-->
+
+<lccdd>
+    <info name="Vertex_IDEA_o1_v03"
```
s6anloes: This should be `v04`, no? Also in line 12 below.
A related question: why is the subdetector called Vertex_**IDEA**_o1_v03? The other subdetectors tend to not include the detector name in their name. This becomes somewhat relevant as now ALLEGRO uses the 'IDEA' vertex design.

FCCee/IDEA/compact/IDEA_o1_v04/IDEA_o1_v04.xml
```diff
@@ -0,0 +1,116 @@
...
+  <!-- Import accelerator components -->
+  <!-- shape based model of the beam pipe -->
+  <include ref="../../../MDI/compact/MDI_o1_v00/Beampipe_o4_v05.xml" />
+  <include ref="../../../MDI/compact/MDI_o1_v00/BeamInstrumentation_o1_v01.xml" />
+
+  <!-- engineered CAD model of the beam pipe -->
+  <!-- In order to use the CAD beampipe, build k4geo with the following CMake option:
+  cmake -D INSTALL_BEAMPIPE_STL_FILES=ON which will download the files needed -->
+  <!-- <include ref="../../../MDI/compact/MDI_o1_v01/Beampipe_CADimport_o1_v02.xml" /> -->
+  <!-- <include ref="../../../MDI/compact/MDI_o1_v01/BeamInstrumentation_o1_v01.xml"/> -->
```
s6anloes: These are no longer the correct files to point to. I think you must have copied an older verion of `IDEA_o1_v03.xml` for creating this file.
Check out the changes made in this commit:
https://github.com/key4hep/k4geo/commit/f19536dcd288556da4345959dd37f1b0a56906a8

```suggestion
<!-- <include ref="../../../MDI/compact/MDI_o1_CADBased_v01/MDI_o1_CADBased_v01_collection.xml" />
  <world material="Vacuum"/> -->
```

FCCee/IDEA/compact/IDEA_o1_v04/Vertex_curvedUltraLight_module.xml
s6anloes: Thanks for introducing these files, the reduction in the 'vertex complete' XML file is impressive!
You'll have to add symlinks to these files in the ALLEGRO directories, since some of the checks are failing, because they can't find the include file.

s6anloes (changes requested):
Hi, thanks for this PR, especiallyyour effort in reducing the breadth/complexity of the files.
Still, I think at some point, we could/should have another discussion about 'uncluttering' some of the calculations that are currently placed in the XML files, but that should be a separate PR.

I have opened a few discussions about detector versioning and naming, but maybe we'll need some input from the coordinators (@BrieucF).

Lastly you might have seen that the clang formatting test fails on some files, so please update.

Cheers!

FCCee/IDEA/compact/IDEA_o2_v01/IDEA_o2_v01.xml
```diff
@@ -51,7 +51,7 @@
   <include ref="../IDEA_o1_v03/DriftChamber_o1_v02.xml"/>
 
   <!-- Import silicon wrapper -->
-  <include ref="../IDEA_o1_v03/SiliconWrapper_o1_v03.xml"/>
+  <include ref="../IDEA_o1_v04/SiliconWrapper_o1_v02.xml"/>
```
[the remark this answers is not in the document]
armin-ilg: This was a mistake, thanks for spotting. Of course IDEA_o2_v01 should use (for the moment) absolutely the same vertex and silicon wrapper as o1

FCCee/IDEA/compact/IDEA_o1_v04/VertexComplete_IDEA_o1_v04.xml
armin-ilg: Ok, I'll do that. I propose 'Vertex_detailed_o1_v04.xml', what do you think? Shall I also change the names of the Vertex xml in IDEA o1_v02 and IDEA/ALLEGRO o1_v03?

FCCee/IDEA/compact/IDEA_o1_v04/SiliconWrapper_disk_fullTile_module.xml
```diff
@@ -0,0 +1,46 @@
+<!-- 
+    Sensor components of the silicon wrapper disk half-tile module (with three sensor modules in z). Multiple sensors are grouped into a 'module' to reduce the number of module IDs needed.
```
s6anloes:
```suggestion
    Sensor components of the silicon wrapper disk half-tile module (with six sensor modules in z). Multiple sensors are grouped into a 'module' to reduce the number of module IDs needed. 
```

This file contains six modules?

s6anloes (approved):
No further comments from my side.
Regarding the naming of the geometry driver, I would like to rename it to `TrackerBarrel_o3_v03` and `TrackerEndcap_o3_v03` (while also renaming the previous two versions retroactively). I have done some local testing and having an alias (two `DECLARE_DETELEMENT` calls) seems to work fine, so we can keep the `_detailed_` names for these versions to avoid breaking changes.
But I think this would probably be simpler in a separate PR.

We should investigate the failing checks before merging though

armin-ilg:
It doesn't look like the timeout is the culprit for the tests on IDEA/ALLEGRO for ubuntu24 nightly and alma9 release, maybe it has something to do with RAM usage, could the limits there be different for the different builds @s6anloes ? 

I'll have a try at reducing the silicon wrapper RAM usage, maybe that'll be enough to make the tests run.

jmcarcell:
High RAM leads to the jobs being killed rather than timing out. I've triggered a rerun. It was unclear to me what happened.

armin-ilg:
> High RAM leads to the jobs being killed rather than timing out. I've triggered a rerun. It was unclear to me what happened.

Thanks, nothing has changed sadly (since many commits the same issue occurred). The 'test_IDEA_withDRC_o1_v03' job is the first one that does not complete, I don't see anything in the logs though that could point towards what the problem is. Any idea?

tmadlener:
More or less random thought; Did any of the recent commits increase the verbosity of any of the tests. I have run into memory issues with CTest when it has to process a lot of (textual) output. So it might be that CTest is choking instead of the actual test using too much memory.

tmadlener:
In case changing order doesn't work, I have made a version of the `key4hep-build` action that runs the tests sequentially.

Change `@main` to `@no-parallel-tests` here for trying that.
https://github.com/key4hep/k4geo/blob/c027f0ea8ceff37447030ba44764e68812fbfa49/.github/workflows/key4hep-build.yaml#L33

armin-ilg:
@s6anloes @tmadlener : Changing the order of the tests was enough (I distributed the the IDEA and ALLEGRO tests throughout CMakeLists, I think these tests are the heaviest on RAM usage), the build tests all work now. The downstream-build fails still, but I guess that is not because of the PR?

I also simplified how the vertex and silicon wrapper (and the associated module.xml's) are called. By going up in the hierarchy, the module.xml's don't have to be symlinked:
```
../../../IDEA/compact/IDEA_o1_v04/SiliconWrapper_module.xml
```

Anything else before we can merge?

PR #566, opened 2026-02-22
Use subdetector-specific VolumeManager in FCCSWGridModuleThetaMerged_k4geo

scott-snyder:
Different subsystems use different bit-widths for the system field of identifiers.  This implies that volume IDs are not necessarily globally unique.  For example, volume IDs for the ALLEGRO HCAL barrel (detid=8, system bit width=4) have been observed to collide with those from the silicon wrapper (detid=24 [SiWr_disks], system bit width=5). This can be avoided by doing the lookup in the system-specific VolumeManager, rather than from the global instance.

Although i have not observed an issue with the ALLEGRO ECal, implement this change to avoid potential issues in the future.

Comments in VolumeManager imply that allowing different bit widths for the system field is intentional.  But it would probably be better to adopt a uniform width to avoid ambiguities.  However, this changes all cellids and invalidates existing data files.

BEGINRELEASENOTES
- detectorSegmentations/src/FCCSWGridModuleThetaMerged_k4geo.cpp: Avoid possible issue due to volume IDs not being globally unique.  
ENDRELEASENOTES

### DIFF
--- a/detectorSegmentations/src/FCCSWGridModuleThetaMerged_k4geo.cpp
+++ b/detectorSegmentations/src/FCCSWGridModuleThetaMerged_k4geo.cpp
@@ -76,7 +76,9 @@ namespace DDSegmentation {
 
     dd4hep::printout(dd4hep::INFO, "FCCSWGridModuleThetaMerged_k4geo", "Precalculating position info of radial layers");
     dd4hep::Detector* dd4hepgeo = &(dd4hep::Detector::getInstance());
-    VolumeManager vman = VolumeManager::getVolumeManager(*dd4hepgeo);
+    const DetElementObject& de = *dd4hepgeo->readout(this->name()).segmentation().detector();
+    VolumeManager vman_glob = VolumeManager::getVolumeManager(*dd4hepgeo);
+    VolumeManager vman = vman_glob.subdetector(de.id);
 
     std::vector<LayerInfo> out;
     out.reserve(m_nLayers);


giovannimarchiori:
Hi @scott-snyder @BrieucF the discussion on an overall harmonisation is still ongoing but I was wondering if from our side (ALLEGRO) we can move from 4 to 5 bits the encodings for all ECAL and HCAL barrels and endcaps? 
We have enough bits since so far

4        ECalBarrelModuleThetaMerged    system:0:4,cryo:4:1,type:5:3,subtype:8:3,layer:11:8,module:19:11,theta:30:10
5        ECalEndcapTurbine              system:0:4,cryo:4:1,type:5:3,subtype:8:3,side:11:-2,wheel:13:3,layer:16:12,module:28:11,rho:39:8,z:47:8
8        HCalBarrelReadout              system:0:4,layer:4:5,row:9:9,theta:18:9,phi:27:10
9        HCalEndcapReadout              system:0:4,type:4:3,layer:7:6,row:13:11,theta:24:11,phi:35:10

giovannimarchiori:
just for the record @BrieucF this is now on my to-do list but since changing cellID encoding (reserving more bits for the system field) requires updating maps of noise, neighbours and so on, I prefer to do it when I am less in a rush mode ;) and not before I have finished sorting out a few things with the latest code that calculates the noise for the ALLEGRO ecal barrel

scott-snyder:
This causes issues with resegmentation.  Closing for now.

PR #578, opened 2026-03-14
account for phi symmetry in idea calo to reduce # of volumes

davidlange6:
Take advantage of phi symmetry in calorimeter to reduce number of volumes created in SCEPCal (edit: now main calorimeter and timing layer).  Remove redundant calculations for trig functions, and reduce memory allocations. 

No physics changes are expected. RSS savings of ~9GB of RSS in IDEA simulation observed. With addition of same change to timing layer, another ~2GB are saved.

BEGINRELEASENOTES
- Given phi symmetry in calorimeter, each phi wedge is now created from same logical volume hierarchy. This reduction in number of independent physical volumes created in SCEPCal saves O(8GB) of RSS.
- organized SCEPCal main layer and timing layer codes to move barrel and endcap creation into separate functions. Cache expensive/repeated function called into local variables.

ENDRELEASENOTES


SanghyunKo (commented):
I went through the PR and it looks okay, except the two minor lines that I didn't understand (they were already there before this PR). I also tried the visualization and it looks fine (at least for the main layer).

Image

davidlange6:
The main changes in g4 results are gone with my new commit. I believe there is still something more to find

lopezzot:
Hello, the large memory footprint of current IDEA_o2 design is (again) due to the dd4hep volume manager having to deal with many placements, in this case crystals. The same problem has been fixed for the tube/fiber calorimeter bringing down the memory footprint from 20 Gb to 6 Gb (before the crystal inclusion). I believe we should apply the same fix to crystals by using `regexSD` instead of standard custom actions.

Of course any geometry improvement is also welcome.

SanghyunKo:
Testing the improvements from using regexSD can be done in parallel. Nevertheless, I think the memory footprint will be one of the main bottlenecks for IDEA_o2 when we move to production (so we should save as much as possible).

(on the point 1) IIUC, we currently never use optical surfaces for the crystal throughout the simulation. The scintillation and Cherenkov signals are calculated by smearing based on the user-given parameters (after checking which process generated the photon) within the SD action: https://github.com/key4hep/k4geo/blob/main/plugins/SCEPCal_MainSDAction.cpp#L31-L39

(FYI, the fiber DRC in IDEA_o1 also had a functionality to calculate optical transportation in the scintillation channel, but we made some compromises since we're moving to the phase where a broader community starts using the code, not just the detector R&D people)

(on the point 2) Regarding the way of retrieving the volume's position, there is a workaround; we can use Geant4 touchables, similar to what's already done in the fiber DRCs:
https://github.com/key4hep/k4geo/blob/main/plugins/FiberDRCaloSDAction.cpp#L173-L175
https://github.com/key4hep/k4geo/blob/main/plugins/DRTubesSglHpr.hh#L119-L140

lopezzot:
> Testing the improvements from using regexSD can be done in parallel. Nevertheless, I think the memory footprint will be one of the main bottlenecks for IDEA_o2 when we move to production (so we should save as much as possible).

It can be done in parallel, but it will inevitably change the SDAction to include the bit field manipulation in order to get the volume IDs. It is not completely independent or detached work.
IDEA_o2 was ~6 GB before the crystal inclusion. I am not sure if the memory footprint will be a bottleneck at all. Anyways we should try to see what happens.

jmcarcell:
Given that the consensus seems not to go ahead with these and there are already conflicts, shall I close this PR?

SanghyunKo:
I'd still prefer to have this merged, as I don't see any reason to replicate all the volumes along the phi direction. Regarding rebasing and checking cell positions, I will try to find some time afterward (but I would not close this PR).

davidlange6:
Closing this in favor of a rebased version on top of #611 to come

jmcarcell:
Keep this open and rebase later on this PR? Then one keeps also the original comments if it's merged from this PR.

davidlange6:
will try and hopefully i don't make a mess of the branch..

davidlange6:
ok, i believe now this pr is rebased on top of the current main+ pr 611.

s6anloes:
The code does not seem to compile at the moment:
```

[202/211] Building CXX object CMakeFiles/k4geoG4.dir/plugins/SCEPCal_MainSDAction.cpp.o
FAILED: [code=1] CMakeFiles/k4geoG4.dir/plugins/SCEPCal_MainSDAction.cpp.o 
ccache /usr/bin/g++ -DBOOST_SPIRIT_USE_PHOENIX_V3 -DDD4HEP_USE_XERCESC -DG4LIB_BUILD_DLL -DG4UI_USE_QT -DG4VIS_USE_OPENGLQT -DG4VIS_USE_RAYTRACER_QT -DG4VIS_USE_TOOLSSG_QT_GLES -DG4VIS_USE_TOOLSSG_QT_ZB -DPODIO_ENABLE_RNTUPLE=1 -DPODIO_JSON_OUTPUT -DPTL_BUILD_DLL -DQT_CORE_LIB -DQT_GUI_LIB -DQT_NO_DEBUG -DQT_WIDGETS_LIB -Dk4geoG4_EXPORTS -I/k4geo/detector/include -I/k4geo/detector/calorimeter/dual-readout/include -I/k4geo/detector/calorimeter/dual-readout-tubes/include -I/k4geo/detectorSegmentations/include -I/k4geo/detectorCommon/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-07-10/x86_64-ubuntu24.04-gcc13.3.0-opt/dd4hep/285e3402e58054f905ffae8702e90ad762555dd0_develop-gcsk6b/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/root/6.38.00-zuvmdm/include/root -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/boost/1.89.0-edvoke/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/xerces-c/3.3.0-jthpb2/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/geant4/11.4.0-zp37ps/include/Geant4 -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/clhep/2.4.7.2-aofkxr/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/qt-base/6.10.1-ptzudi/include/QtCore -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/qt-base/6.10.1-ptzudi/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/qt-base/6.10.1-ptzudi/mkspecs/linux-g++ -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/qt-base/6.10.1-ptzudi/include/QtGui -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/qt-base/6.10.1-ptzudi/include/QtWidgets -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/expat/2.7.3-svakk2/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/zlib-ng/2.2.4-v3vpc6/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-07-10/x86_64-ubuntu24.04-gcc13.3.0-opt/podio/dabf374a5d6b3f84b0cb3e48102c18896acd3ad8_develop-mrmjuu/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/vdt/0.4.6-pnslte/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-07-10/x86_64-ubuntu24.04-gcc13.3.0-opt/edm4hep/898a6e8e20393d94559523133c162bd2d3101288_develop-dqusgy/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-02-26/x86_64-ubuntu24.04-gcc13.3.0-opt/nlohmann-json/3.12.0-ykb3ot/include -isystem /cvmfs/sw-nightlies.hsf.org/key4hep/releases/2026-07-10/x86_64-ubuntu24.04-gcc13.3.0-opt/lcio/dbdf5338378bd23d77fd0ed71e45078594cd4296_develop-kkj5fb/include -fPIC -Wall -Wextra -Wpedantic -Wshadow -Wdeprecated -fdiagnostics-color=always -Werror -Wno-error=cpp -Wno-error=deprecated-declarations -O2 -g -DNDEBUG -std=c++20 -fPIC -W -Wall -pedantic -Wno-non-virtual-dtor -Wno-long-long -Wwrite-strings -Wpointer-arith -Woverloaded-virtual -Wno-variadic-macros -Wshadow -pipe -pthread -ftls-model=global-dynamic -Wodr -fipa-icf -flto-odr-type-merging -std=c++20 -fPIC -MD -MT CMakeFiles/k4geoG4.dir/plugins/SCEPCal_MainSDAction.cpp.o -MF CMakeFiles/k4geoG4.dir/plugins/SCEPCal_MainSDAction.cpp.o.d -o CMakeFiles/k4geoG4.dir/plugins/SCEPCal_MainSDAction.cpp.o -c /k4geo/plugins/SCEPCal_MainSDAction.cpp
/k4geo/plugins/SCEPCal_MainSDAction.cpp: In member function 'bool dd4hep::sim::Geant4SensitiveAction<T>::process(const G4Step*, G4TouchableHistory*) [with T = SCEPCal::SCEPCal_MainSDAction]':
/k4geo/plugins/SCEPCal_MainSDAction.cpp:78:44: error: 'cellID0' was not declared in this scope; did you mean 'cellID'?
   78 |     auto cellID_phi = segmentation->setPhi(cellID0, cellID2);
      |                                            ^~~~~~~
      |                                            cellID
/k4geo/plugins/SCEPCal_MainSDAction.cpp:81:14: error: redeclaration of 'G4Track* track'
   81 |     G4Track* track = step->GetTrack();
      |              ^~~~~
/k4geo/plugins/SCEPCal_MainSDAction.cpp:73:14: note: 'G4Track* track' previously declared here
   73 |     G4Track* track = step->GetTrack();
      |              ^~~~~
/k4geo/plugins/SCEPCal_MainSDAction.cpp:103:9: error: 'edep' was not declared in this scope
  103 |     if (edep > 0.) {
      |         ^~~~
```

Also there pre-commit is failing. Could you update the corresponding files?

davidlange6:
sigh - ok, means something went wrong with my rebase.. must mean i didn't test the last version. Will redo that.

davidlange6:
> sigh - ok, means something went wrong with my rebase.. must mean i didn't test the last version. Will redo that.

ok, I believe i have done it better this time.

davidlange6:
Clearly doing a rebase with #611 before it was final was the wrong approach.. I may not have gotten all the new conflicts properly. Will test and confirm.

davidlange6:
ok, indeed my branch is now totally out of sync since 611 was squashed and i have a partial (and wrong) history. Will see what can be fixed.

davidlange6:
Now this branch compiles and runs. i still need to check that the conflicts were all resolved without losing changes...

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -12,6 +12,9 @@
 #include "TGeoTrd2.h"
 #include "detectorSegmentations/SCEPCal_MainSegmentationHandle_k4geo.h"
 #include <bitset>
+#include <unordered_map>
+
+#include "XML/Utilities.h"
```
andresailer: double import (2 lines down)

plugins/SCEPCal_MainSDAction.cpp
```diff
@@ -69,24 +69,27 @@ namespace sim {
...
 
-      auto cIDwithChannel = segmentation->setCellID(vID, isCherenkov);
-
+      auto cellID_phi = segmentation->setPhi(vID, cellID2);
+      auto cIDwithChannel = segmentation->setCellID(cellID_phi, isCherenkov);
+      
       auto* hit = coll->findByKey<Geant4Calorimeter::Hit>(cIDwithChannel);
 
       if (!hit) {
         DDSegmentation::Vector3D pos = segmentation->position(vID); // always use scintillation channel for position
```
jbeirer: Shouldn't this use the phi-corrected `cellID_phi `rather than `vID`, as is already done in `SCEPCal_TimingSDAction`? vID corresponds to the shared crystal constructed with `phi=0`, while `cellID_phi` contains the actual phi of this placement. Since position() directly uses the supplied ID to look up `m_positionOf`, this seems to return the `phi=0` position for every replicated phi slice.

detector/calorimeter/SCEPCal_TimingLayer.cpp
```diff
@@ -275,205 +295,228 @@ static dd4hep::Ref_t create_detector_SCEPCal_TimingLayer(dd4hep::Detector& theDe
...
+  dd4hep::Volume tlendcapPhiAssemblyVolume("tlendcapPhiVol", tlendcapPhiAssemblyShape, theDetector.material("Vacuum"));
+  tlendcapPhiAssemblyVolume.setVisAttributes(theDetector, tlendcapAssemblyPhiVisXML.visStr());
 
-    dd4hep::Volume tlendcapPhiAssemblyVolume("tlendcapPhiVol", tlendcapPhiAssemblyShape,
+  dd4hep::Volume tlendcapPhiAssemblyVolume_1("tlendcapPhiVol_1", tlendcapPhiAssemblyShape_1,
                                              theDetector.material("Vacuum"));
-    tlendcapPhiAssemblyVolume.setVisAttributes(theDetector, tlendcapAssemblyPhiVisXML.visStr());
-    tlendcapGlobalAssemblyVol.placeVolume(tlendcapPhiAssemblyVolume, Transform3D(rotZphiGlobal));
+  tlendcapPhiAssemblyVolume_1.setVisAttributes(theDetector, tlendcapAssemblyPhiVisXML.visStr());
+  tlendcapGlobalAssemblyVol_1.placeVolume(tlendcapPhiAssemblyVolume_1);
```
jbeirer: Shouldn't `tlendcapPhiAssemblyVolume_1` also be placed in the final `iPhi` loop below, analogous to `tlendcapPhiAssemblyVolume`? As written, `_1` seems to be placed only once here without a phi rotation, while the other endcap is replicated over all phi slices. Wouldn't this leave the negative timing endcap with only one phi wedge?

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -577,234 +611,245 @@ static dd4hep::Ref_t create_detector_SCEPCal_MainLayer(dd4hep::Detector& theDete
...
+          XYZVector posGlobal_1 = (rotYthGlobal_1 * dispGlobal_1 + DISP_PROJ_R);
+
+          CreateEightPointShapeVolume_SetVolAttributes_Place_SetCellId(
+              "EndcapCrystalF", XTAL_LEN_F / 2, vsub, crystalFXML, trans_dispFsub, endcapThetaAssemblyVolume,
+              ENDCAP_SYSTEM_NO, iTheta, nGamma, nEpsilon, 0, posGlobal,  ENDCAP_PHI_START, ENDCAP_PHI_END, phi_endcap_rotations);
+          numCrystalsEndcap += 1;
+
+          CreateEightPointShapeVolume_SetVolAttributes_Place_SetCellId(
+              "EndcapCrystalF_1", XTAL_LEN_F / 2, vsub1, crystalFXML, trans_dispFsub_1, endcapThetaAssemblyVolume_1,
+              ENDCAP_SYSTEM_NO, 2 * N_THETA_ENDCAP + N_THETA_BARREL - iTheta, nGamma, nEpsilon, 0, posGlobal_1, ENDCAP_PHI_START, ENDCAP_PHI_END, phi_endcap_rotations);
```
jbeirer: Shouldn't this still be `2 * N_THETA_ENDCAP + N_THETA_BARREL - 1 - iTheta`? The corresponding rear crystal `EndcapCrystalR_1` below still uses the `-1`, and the code before this change also used `-1` here. Without it, the front and rear crystals of the same negative-endcap tower seem to get different theta IDs.

detector/calorimeter/SCEPCal_TimingLayer.cpp
```diff
@@ -275,205 +295,228 @@ static dd4hep::Ref_t create_detector_SCEPCal_TimingLayer(dd4hep::Detector& theDe
...
+      XYZVector posGlobal_1 = (dispGlobal_1);
+
+      CreateEightPointShapeVolume_SetVolAttributes_Place_SetCellId(
+          "TlEndcapCrystal", XTAL_DEPTH_T / 2, verticesT, crystalTXML, Transform3D(dispT), tlendcapThetaAssemblyVolume,
+          TLENDCAP_SYSTEM_NO, iTheta, nGamma, posGlobal, TLENDCAP_PHI_START, TLENDCAP_PHI_END, phi_endcap_rotations);
+      numCrystalsTlEndcap += 1;
+
+      CreateEightPointShapeVolume_SetVolAttributes_Place_SetCellId(
+          "TlEndcapCrystal_1", XTAL_DEPTH_T / 2, verticesT_1, crystalTXML, Transform3D(dispT),
+          tlendcapThetaAssemblyVolume_1, TLENDCAP_SYSTEM_NO, 2 * N_THETA_TLENDCAP + N_THETA_TLBARREL - iTheta, nGamma,
```
jbeirer: Shouldn't this still be `2 * N_THETA_TLENDCAP + N_THETA_TLBARREL - 1 - iTheta`? The code before this change used the `-1` here, and without it the negative timing-endcap theta numbering seems to be shifted by one, with the first ring getting an index one past the expected maximum.

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -385,190 +402,207 @@ static dd4hep::Ref_t create_detector_SCEPCal_MainLayer(dd4hep::Detector& theDete
...
+      if ((iPhi < PHI_LOAD_START) || (iPhi > PHI_LOAD_END))
+        continue;
+    } else {
+      if ((iPhi < PHI_LOAD_START) && (iPhi > PHI_LOAD_END))
+        continue;
+    }
+    nPhiSlicePlaced += 1;
+    double phiGlobal = iPhi * D_PHI_GLOBAL;
+    RotationZ rotZphiGlobal(phiGlobal);
+    auto pv = barrelGlobalAssemblyVol.placeVolume(barrelPhiAssemblyVolume, Transform3D(rotZphiGlobal));
```
jbeirer: Since the SD now gets the phi index from `GetCopyNumber(2)`, shouldn't we explicitly use `iPhi` as the copy number here, e.g. `placeVolume(barrelPhiAssemblyVolume, iPhi, Transform3D(rotZphiGlobal))`? The implicit DD4hep copy number is based on the number of already placed daughters, so it only matches `iPhi` if the phi slices are placed consecutively starting from zero. With `PHI_LOAD_START/END` selecting only part of the detector, the copy number and actual `iPhi` can differ.

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -300,80 +310,87 @@ static dd4hep::Ref_t create_detector_SCEPCal_MainLayer(dd4hep::Detector& theDete
...
         thePlacedVol.addPhysVolID("epsilon", nEpsilon);
         thePlacedVol.addPhysVolID("depth", nDepth);
 
-        segmentation->savePosition(volID_32, posGlobal);
+	//save the cell ids for each copy around phi
+	for ( int iPhi=phi_start; iPhi< phi_end; iPhi++ ) {
+	  XYZVector posGlobalPhi = phi_rotations[iPhi] * posGlobal;
+	  auto volID = segmentation->setVolumeID(nSystem, iPhi, nTheta, nGamma, nEpsilon, nDepth);
+	  int volID_32 = segmentation->getFirst32bits(volID);
+	  segmentation->savePosition(volID_32, posGlobalPhi);
```
jbeirer: Since the main goal here is to exploit the phi symmetry to reduce memory, do we still need to store one position entry for every phi copy? Since all positions are related by the known phi rotation, could we store only the `phi=0` position and rotate it on demand in `position()`? The current loop still makes `m_positionOf` scale with the full number of cells in phi, so this may retain a sizeable part of the memory usage that the geometry sharing is trying to remove.

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -385,190 +402,207 @@ static dd4hep::Ref_t create_detector_SCEPCal_MainLayer(dd4hep::Detector& theDete
...
+  dd4hep::Volume barrelPhiAssemblyVolume("barrelPhiAssembly", barrelPhiAssemblyShape, theDetector.material("Vacuum"));
+  barrelPhiAssemblyVolume.setVisAttributes(theDetector, barrelAssemblyPhiVisXML.visStr());
 
-      double r0e = (BARREL_INNER_R + PROJ_OFFSET_R) / sin(thC);
-      double r1e = r0e + XTAL_LEN_F;
-      double r2e = r1e + XTAL_LEN_R;
-      double y0e = r0e * tan(D_THETA_BARREL / 2.);
-      double y1e = r1e * tan(D_THETA_BARREL / 2.);
-      double y2e = r2e * tan(D_THETA_BARREL / 2.);
+  for (int iTheta = 0; iTheta < N_THETA_BARREL; iTheta++) {
```
jbeirer: With this refactor, doesn't `CONSTRUCT_BARREL=false` still construct the complete barrel phi wedge, including all theta/gamma/crystal volumes and position entries, and only skip placing it later? Previously the whole barrel construction was inside the phi loop starting from `CONSTRUCT_BARREL ? BARREL_PHI_START : BARREL_PHI_END`, so setting it to false skipped the construction entirely. Should the reusable wedge construction also be guarded by `CONSTRUCT_BARREL`?

detector/calorimeter/SCEPCal_MainLayer.cpp
jbeirer, line 17:
```suggestion
```
Remove duplicate include

plugins/SCEPCal_TimingSDAction.cpp
```diff
@@ -54,24 +54,35 @@ namespace sim {
...
+    //lower 32 bits of cellID but for phi=0 slidce
+    auto cellID_0 = thePreStepTouchable->GetCopyNumber(0);
+    //get phi index from grandparent
+    auto cellID_2 = thePreStepTouchable->GetCopyNumber(2);
 
     dd4hep::Segmentation* _geoSeg = &m_segmentation;
     auto segmentation =
         dynamic_cast<dd4hep::DDSegmentation::SCEPCal_TimingSegmentation_k4geo*>(_geoSeg->segmentation());
 
+    //int system = segmentation->System(cellID_0);
```
jbeirer: Can we remove this commented-out implementation now that `setPhi()` is used below?

plugins/SCEPCal_TimingSDAction.cpp
```diff
@@ -54,24 +54,35 @@ namespace sim {
     G4StepPoint* thePrePoint = step->GetPreStepPoint();
     G4TouchableHandle thePreStepTouchable = thePrePoint->GetTouchableHandle();
 
-    auto cellID = thePreStepTouchable->GetCopyNumber(0);
+    //lower 32 bits of cellID but for phi=0 slidce
```
jbeirer:
```suggestion
    //lower 32 bits of cellID but for phi=0 slice
```

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -300,80 +310,87 @@ static dd4hep::Ref_t create_detector_SCEPCal_MainLayer(dd4hep::Detector& theDete
...
+    phi_endcap_rotations[iPhi]=rotZPhiGlobal;
+  }
+  
   // Lambda for crystals
   auto CreateEightPointShapeVolume_SetVolAttributes_Place_SetCellId =
-      [&theDetector, &sens, &segmentation, &ScepcalDetElement, &USE_OPTICAL_SURFACES,
-       &PbWO4_to_ESR](std::string volName, double dz, const std::array<double, 16> vertices, xml_comp_t compXml,
-                      dd4hep::Transform3D transform, dd4hep::Volume assemblyVol, int nSystem, int nPhi, int nTheta,
-                      int nGamma, int nEpsilon, int nDepth, XYZVector posGlobal) {
+    [&theDetector, &sens, &segmentation, &ScepcalDetElement, &USE_OPTICAL_SURFACES, &PbWO4_to_ESR, &D_PHI_GLOBAL](
```
jbeirer: `D_PHI_GLOBAL` doesn't seem to be used in this lambda anymore. Can it be removed from the capture?

detector/calorimeter/SCEPCal_TimingLayer.cpp
```diff
@@ -239,34 +240,53 @@ static dd4hep::Ref_t create_detector_SCEPCal_TimingLayer(dd4hep::Detector& theDe
 
   ScepcalDetElement.setPlacement(tlbarrelAssemblyPlacedVol);
 
+  std::unordered_map<int, RotationZ> phi_barrel_rotations, phi_endcap_rotations;
```
jbeirer: Since the phi indices are dense and run from `0` to `PHI_SEGMENTS - 1`, would a `std::vector<RotationZ>` be simpler here? It would avoid the hashing/allocation overhead of `unordered_map` and still allow direct lookup with `phi_rotations[iPhi]`.

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -577,234 +611,245 @@ static dd4hep::Ref_t create_detector_SCEPCal_MainLayer(dd4hep::Detector& theDete
...
-        endcapGlobalAssemblyVol_1.placeVolume(endcapPhiAssemblyVolume_1, Transform3D(rotZphiGlobal));
-      }
-    }
+  dd4hep::Volume endcapPhiAssemblyVolume("endcapPhiVol", endcapPhiAssemblyShape, theDetector.material("Vacuum"));
+  endcapPhiAssemblyVolume.setVisAttributes(theDetector, endcapAssemblyPhiVisXML.visStr());
+  dd4hep::Volume endcapPhiAssemblyVolume_1("endcapPhiVol_1", endcapPhiAssemblyShape_1, theDetector.material("Vacuum"));
+  endcapPhiAssemblyVolume_1.setVisAttributes(theDetector, endcapAssemblyPhiVisXML.visStr());
 
-    for (int iTheta = ENDCAP_THETA_START; iTheta < N_THETA_ENDCAP; iTheta++) {
+  for (int iTheta = ENDCAP_THETA_START; iTheta < N_THETA_ENDCAP; iTheta++) {
```
jbeirer: Same question as for the barrel: with this refactor, doesn't `CONSTRUCT_ENDCAP=false` still construct the complete endcap wedges, including all crystal volumes and position entries, and only skip placing them later? Previously the whole construction was inside the `CONSTRUCT_ENDCAP`-controlled phi loop. Should the reusable endcap construction also be guarded?

detector/calorimeter/SCEPCal_TimingLayer.cpp
```diff
@@ -239,34 +240,53 @@ static dd4hep::Ref_t create_detector_SCEPCal_TimingLayer(dd4hep::Detector& theDe
...
-        thePlacedVol.addPhysVolID("phi", nPhi);
         thePlacedVol.addPhysVolID("theta", nTheta);
         thePlacedVol.addPhysVolID("gamma", nGamma);
 
-        segmentation->savePosition(volID_32, posGlobal);
+	for ( int iPhi=phi_start; iPhi< phi_end; iPhi++ ) {
+	  XYZVector posGlobalPhi = phi_rotations[iPhi] * posGlobal;
+	  auto volID = segmentation->setVolumeID(nSystem, iPhi, nTheta, nGamma);
+	  int volID_32 = segmentation->getFirst32bits(volID);
+	  //std::cout << posGlobalPhi << " " << volID_32 << std::endl;
```
jbeirer: Can we remove this leftover debug print?

jbeirer (changes requested):

davidlange6:
@jbeirer - cool - thanks for finding a number of places where I did not get the rebase properly.. Will fix these and the other comments.

detector/calorimeter/SCEPCal_TimingLayer.cpp
davidlange6: i'm not sure this is simpler as the code does not seem to assume 0 to PHI_SEGMENTS but rather a min to a max. But I have changed this to a vector

detector/calorimeter/SCEPCal_MainLayer.cpp
davidlange6: ok, that turned into a larger change. I have moved the barrel and endcap builders into separate functions and guarded each with the appropriate if statement (both for the main detector and the timing layer.

I will turn this back to a draft while i'm checking the damage I've done

detector/calorimeter/SCEPCal_MainLayer.cpp
davidlange6: i guess that would trigger a per-event cost to apply that rotation. I have not considered it (i'm not a big fan of having to store these positions..)

plugins/SCEPCal_MainSDAction.cpp
davidlange6: yes.

detector/calorimeter/SCEPCal_TimingLayer.cpp
davidlange6: yes, that is incorrect...

detector/calorimeter/SCEPCal_MainLayer.cpp
davidlange6: yes, i agree. i have changed this in all of the placements

davidlange6:
ok, I now see unchanged results comparing before/after this PR on a ddsim run and believe I addressed the review comments so far.

jbeirer:
Could you please apply the pre-commit changes? Otherwise this looks good to me now

davidlange6:
hopefully done now.

jbeirer (approved):

jmcarcell:
The release notes could be expanded, what has been done exactly?

davidlange6:
edited the release notes. perhaps its more clear now?

jbeirer:
Let's wait if anyone else has comments on this, otherwise I will merge tomorrow

detector/calorimeter/SCEPCal_TimingLayer.cpp
```diff
@@ -21,6 +22,322 @@ using ROOT::Math::RotationZ;
...
+  auto volID_0 = segmentation->setVolumeID(nSystem, 0, nTheta, nGamma);
+  int volID_0_32 = segmentation->getFirst32bits(volID_0);
+  dd4hep::PlacedVolume thePlacedVol = assemblyVol.placeVolume(theVolume, volID_0_32, transform);
+
+  thePlacedVol.addPhysVolID("system", nSystem);
+  thePlacedVol.addPhysVolID("theta", nTheta);
+  thePlacedVol.addPhysVolID("gamma", nGamma);
+
+  for (int iPhi = phi_start; iPhi < phi_end; iPhi++) {
+    XYZVector posGlobalPhi = phi_rotations[iPhi + phi0] * posGlobal;
```
jmcarcell: This is different from `Main` where a `std::unordered_map` is used (just pointing out). Shouldn't it be `iPhi - phi0` since that is the starting index? And this seems to work because when this is called `p.TIMING_PHI_START` and `p.TLENDCAP_PHI_START` are 0.

davidlange6: ha - yes. it certainly should be iphi-phi0

PR #579, opened 2026-03-16
Use the hierarchical drivers in MuSIC_v2 tracker as well

tmadlener:

BEGINRELEASENOTES
- Use the more hierarchical detector drivers for MuSIC_v2 introduced in #549 in more detector concepts to facilitate conversion to  ACTS

ENDRELEASENOTES

Technically the updated ACTS geometry conversion API could also deal with ungrouped / less hierarchical drivers but the conversion code on our side would be easier to maintain if we do not have to deal with this, resp. if we treat all of these the same.

### DIFF
--- a/MuColl/MuSIC/compact/MuSIC_v2/InnerTracker_o2_v07_01.xml
+++ b/MuColl/MuSIC/compact/MuSIC_v2/InnerTracker_o2_v07_01.xml
@@ -125,7 +125,7 @@
         </detector>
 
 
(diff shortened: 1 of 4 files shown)


Zehvogel:
I think for CLD we would need to bump the detector version (or only add it to v08) if we stick to our philosophy of treating almost everything as a potentially breaking change. Maybe @andresailer has an opinion?

madbaron (approved):

tmadlener:
I have removed the changes that touch CLD for now to decouple the discussion a bit. I will bring that back as a separate PR and continue with merging this.

PR #581, opened 2026-03-18
WireTracker_info helper struct, generalizing DCH_info from DD4HEP

sfranchel:

BEGINRELEASENOTES
- Implementation of `WireTracker_info` interface structure, generalizing the DD4HEP implementation of the `DCH_info` helper struct. The latter is now a child struct of `WireTracker_info`, together with the new `STT_info_struct`.
- Update naming convention in the Straw Tube Tracker (`multilayer` -> `superlayer`) for easier code sharing with Drift Chamber.

ENDRELEASENOTES

---
This PR takes the code from the DD4HEP [`DCH_info_struct`](https://github.com/AIDASoft/DD4hep/blob/cd6cbfdfc9e8ec942a7625b697848518ef55efb5/DDRec/include/DDRec/DCH_info.h#L32) and ports it to the k4geo repo, implementing the  `WireTracker_info` interface structure to share code between STT and DCH trackers.

The new interface contains a reduced set of elements, but keeps the explicit implementation of the functions that estimate the relative/global hit positions, used by the digitizers. 

Two derived class are defined: `DCH_info_struct` and `STT_info_struct`. The former should be an equivalent implementation of the DD4HEP version except for the methods of the `LayerInfo` sub-structure, which have been brought to the outer struct. This allows for more flexible definition of such methods (`IsStereoPositive`, `StereoSign` and `Pitch_z0`) without too many complications (or at least, I couldn't find a nicer way).

For easier code sharing, part of the naming scheme of the Drift Chamber is brought to the STT. In particular `multilayers` have been renamed as `superlayers`. Accordingly, some variables have been renamed in the cpp file of the geometry description.


sfranchel:
Pinging @s6anloes and Giovanni (by email) for a preliminary review.

The current implementation compiles and runs. But more thorough testing is probably needed.

The DCH digitizers would have to be updated. E.g. this code wouldn't work anymore:
https://github.com/key4hep/k4RecTracker/blob/main/DCHdigi/src/DCHdigi_v02.cpp#L134-L140
```
      // when building the twisted tube, the twist angle is defined as:
      //     cell_twistangle    = l.StereoSign() * DCH_i->twist_angle
      // which forces the stereoangle of the wire to have the oposite sign
      WireStereoAngle = (-1.) * l.StereoSign() * m_dch_info->stereoangle_z0(cell_rave_z0);
```
And since we're touching this part, I would try to solve the ambiguity of that -1. If we store directly the stereo angle for each layer, we could store it with the correct sign.

Cheers,
Stefano

detector/tracker/StrawTubeTracker_o1_v01_geo.cpp
```diff
@@ -66,61 +77,76 @@ static dd4hep::Ref_t create_straw_tracker(dd4hep::Detector& theDetector, xml_h e
...
+    // fill per-superlayer values of STT_info
+    STT_i->innermost_radius.push_back(layerRadius);
+    STT_i->nsectors.push_back(SLSectors);
+    STT_i->delta_phi.push_back(SLDelta_phi);
+    STT_i->stereo.push_back(Angle);
+
+    if (STT_i->nlayersPerSuperlayer != 0){
+      strawAssert(STT_i->nlayersPerSuperlayer == SLLayers,
+        "ERROR: specifying different number of layers per superlayer!\n"
+        "       This is currently not supported by STT_info.");
```
s6anloes: What would be needed to make this work?
We can replace the `nlayersPerSuperlayer` member with a vector maybe. To get the total number of layers you can take the sum of all entries. The most complicated would be the function to calculate which superlayer you're in based on the running layer number. But also implementing this should not be too complicated.
Might be worth pursuing, because conceptually, there is no reason for the STT (and also the DCH for that matter) to always have the same number of layers per superlayer. This could be an annoyance for future users

detector/tracker/StrawTubeTracker_o1_v01_geo.cpp
```diff
@@ -66,61 +77,76 @@ static dd4hep::Ref_t create_straw_tracker(dd4hep::Detector& theDetector, xml_h e
...
-    strawAssert(MLThickness > 0 || MLLayers > 0, "ERROR: Each <layer> in straw tube tracker must have either\n"
+    strawAssert(SLThickness > 0 || SLLayers > 0, "ERROR: Each <layer> in straw tube tracker must have either\n"
                                                  "       thickness or count attribute defined.");
 
-    double minThickness = (tubeThickness + tube_gap) * (1 + 0.866 * (MLLayers - 1));
-    if (MLThickness < 0) {
-      MLThickness = minThickness + MLphiGap; // add default gap to minimum possible thickness
+    // N.B. 0.866 = sqrt(3)/2 ==> minimal radial distance between  staggered tubes centers 
+    // in consecutive layers, in units of the diameter (i.e. tube thickness)
+    double minThickness = (tubeThickness + tube_gap) * (1 + 0.866 * (SLLayers - 1));
```
s6anloes:
```suggestion
    // N.B. sqrt(3)/2 ==> minimal radial distance between  staggered tubes centers 
    // in consecutive layers, in units of the diameter (i.e. tube thickness)
    double minThickness = (tubeThickness + tube_gap) * (1 + sqrt(3)/2 * (SLLayers - 1));
```
No real reason to use the approximation here, by the way.
Using sqrt(3) might make the connection to hexagonal stacking more obvious to someone who is familiar with the concept.

detector/tracker/WireTracker_info.h
s6anloes, line 1: Could you add section to the README to explain what `WireTracker_info.h` does?

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+        double height_z0 = {0};
+        /// cell parameter
+        double width_z0 = {0};
+        /// stereoangle at z=0
+        angle_t stereo_sw_z0 = {0};
+        /// radius (cylindrical coord) of sensitive wire
+        length_t radius_sw_z0 = {0};
+        /// radius (cylindrical coord) of 'down' field wires
+        length_t radius_fdw_z0 = {0};
+        /// radius (cylindrical coord) of 'up' field wires
```
s6anloes:
```suggestion
        /// placement radius (cylindrical coord) of sensitive wire
        length_t radius_sw_z0 = {0};
        /// placement radius (cylindrical coord) of 'down' field wires
        length_t radius_fdw_z0 = {0};
        /// placement radius (cylindrical coord) of 'up' field wires
```

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+    length_t Lhalf = {0};
+    /// Inner radius of the active volume
+    length_t rin = {0};
+    /// Outer radius of the active volume
+    length_t rout = {0};
+
+    /// input number of layers in each superlayer
+    layer_t nlayersPerSuperlayer = {0};
+    /// input number of superlayers
+    /// superlayer is an abstract level of grouping layers used to
```
s6anloes:
```suggestion
    /// superlayer is an (abstract) level of grouping layers used to
```
Not abstract for the straw tubes

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+    /// parametrize the increment of cells in each layer
+    layer_t nsuperlayers = {0};
+    /// Calculated as dch_nlayersPerSuperlayer * nsuperlayers
```
s6anloes:
```suggestion
    /// Calculated as nlayersPerSuperlayer * nsuperlayers
```
Also, this will no longer be the case if we do the change of having a variable number of layers per superlayer (`nlayersPerSuperlayer` becoming a vector)

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+    layer_t nlayers = {0};
+
+    /// global twist angle
+    /// alternating layers will change its sign
+    angle_t twist_angle = {0};
```
s6anloes: Should this perhaps be moved to `DCH_info` since it's not used in `STT_info`?

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+
+    // ------------------------------------------------------------------
+    //  Convenience setters
+
+    void Set_lhalf(length_t _Lhalf){Lhalf= _Lhalf;}
+    void Set_rin  (length_t _rin  ){rin  = _rin;  }
+    void Set_rout (length_t _rout ){rout = _rout; }
+
+    void Set_nlayersPerSuperlayer(int _nlayersPerSuperlayer){nlayersPerSuperlayer = _nlayersPerSuperlayer;}
```
s6anloes: Just tagging code lines with `#nlayerPerSuperlayer` that would need to be changed with variable `nlayersPerSuperlayer`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+    void Set_nsuperlayers        (int _nsuperlayers        ){nsuperlayers         = _nsuperlayers;        }
+
+    void Set_twist_angle (length_t _dch_twist_angle ){twist_angle = _dch_twist_angle;}
+
+    /// ------------------------------------------------------------------
+    // Wire-distance calculations  (used by digitisation / reconstruction)
+    // Notation: `ilayer` is the correlative number of the layer, `layer` is reserved to number within a superlayer
+    layer_t  CalculateILayerFromCellIDFields(int layer, int superlayer) const { layer_t ilayer = layer + (this->nlayersPerSuperlayer)*superlayer + 1; return ilayer;}
```
s6anloes: `#nlayerPerSuperlayer`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+///
+inline WireTracker_info_struct::Vector3D
+WireTracker_info_struct::Calculate_wire_z0_point(int superlayer, int ilayer, int sector, int nphi) const {
+  auto&    l   = this->database.at(ilayer);
+  double   rz0 = l.radius_sw_z0;
+  Vector3D p1(rz0, 0, 0);
+  double   phi_z0 = Get_cell_phi_angle(superlayer, ilayer, sector, nphi);
+  ROOT::Math::RotationZ z_rotation = ROOT::Math::RotationZ(phi_z0);
+  z_rotation(p1);
+  return p1;
```
s6anloes:
```suggestion
  Vector3D p1_rot = z_rotation(p1);
  return p1_rot;
```

I believe here is a bug.
The rotation does not occur in-place, you should use the return value.
At least that's how I understand the documentation: [https://root.cern/doc/v608/TransformPage.html](https://root.cern/doc/v608/TransformPage.html)
**Please double-check this!**

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+
+  // calculate phi rotation of whole twisted tube, ie, rotation at z=0
+  double phi_z0 = Get_cell_phi_angle(superlayer, ilayer, sector, nphi);
+  auto z_rotation = ROOT::Math::RotationZ(phi_z0);
+  z_rotation(p1);
+  z_rotation(p2);
+
+  //--- end calculating wire position
+
+  return (p2 -p1).Unit();
```
s6anloes:
```suggestion
  Vector3D p1_rot = z_rotation(p1);
  Vector3D p2_rot = z_rotation(p2);

  //--- end calculating wire position

  return (p2_rot -p1_rot).Unit();
```

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+    /// dphi between two sensing wires
+    angle_t Get_phi_width(int ilayer) const {
+        return (dd4hep::twopi/Get_ncells(ilayer))*dd4hep::rad;
+    };
+
+    //--------------------------------------------------------------
+    // DCH geo interface implementation
+
+    /// phi positioning, adding offset for odd ilayers
+    /// there is a staggering in phi for al {return database.at(ilayer).nwires/2;}ternating layers, 0.25*cell_phi_width*(ilayer%2);
```
s6anloes:
```suggestion
    /// there is a staggering in phi for alternating layers, 0.25*cell_phi_width*(ilayer%2);
```

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+    //--------------------------------------------------------------
+    // Per-layer geometry record
+
+    // Internal helper struct for defining the layer layout
+    struct LayerInfo
+    {
+        /// layer number
+        layer_t layer = {0};
+        /// 2x number of cells in that layer
+        int nwires = {0};
```
s6anloes:
```suggestion
        /// number of cells in that layer
        int ncells = {0};
```

Let's change this to `ncells` instead.
It bothers me to have `nwires` as member if it not sensible for `STT` and `ncells` is right there

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    /// tan(stereoangle) = R(z=L/2) / (L/2) * sin( twist_angle/2)
+    angle_t stereoangle_zLhalf(length_t r_zLhalf) const {
+        return atan( r_zLhalf/Lhalf*sin(twist_angle/2/dd4hep::rad));
+    }
+
+    /// separation between wires (along the circle)
+    length_t Pitch_z0(LayerInfo layer, length_t r_z0) const {
+        return dd4hep::twopi*r_z0/layer.nwires;
```
s6anloes:
```suggestion
        return dd4hep::twopi*r_z0/(layer.ncells*2);
```

Tagging all the changes with `#ncells`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+    };
+
+    /// calculate sign based on IsStereoPositive
+    int StereoSign(LayerInfo layer) const {
+        return (IsStereoPositive(layer)*2 - 1);
+    };
+
+    /// number of cells per layer, calculated as the total number of wires / 2
+    int Get_ncells(int ilayer) const {
+        return database.at(ilayer).nwires/2;
```
s6anloes:
```suggestion
        return database.at(ilayer).ncells;
```
`#ncells`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+        this->nlayers = this->nsuperlayers * this->nlayersPerSuperlayer;
+
+        ff_check_positive_parameter(this->first_width,"width of first layer cells" );
+        ff_check_positive_parameter(this->first_sense_r,"radius of first layer cells" );
+
+        // initialize layer 1 from input parameters
+        {
+            LayerInfo layer1_info;
+            layer1_info.layer         = 1;
+            layer1_info.nwires        = 2*this->ncell0;
```
s6anloes:
```suggestion
            layer1_info.ncells        = this->ncell0;
```
`#ncells`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+        // the rest are left as methods of WireTracker_info or WireTracker_info_layer class
+        // loop over all layers, skipping the first one
+        for(int ilayer = 2; ilayer<= this->nlayers; ++ilayer) {
+            // initialize empty object, parameters are set later
+            LayerInfo layer_info;
+
+            // the loop counter actually corresponds to the layer number
+            layer_info.layer = ilayer;
+            // nwires is twice the number of cells in this particular layer (ilayer)
+            layer_info.nwires = 2*(this->ncell0 + this->ncell_increment*Get_nsuperlayer_minus_1(ilayer) );
```
s6anloes:
```suggestion
            layer_info.ncells = this->ncell0 + this->ncell_increment*Get_nsuperlayer_minus_1(ilayer);
```
`#ncells`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+            {
+                double h  = previousLayer.height_z0;
+                double ru = previousLayer.radius_fuw_z0;
+                double rd = previousLayer.radius_fdw_z0;
+
+                if(0 == Get_nsuperlayer_minus_1(ilayer))
+                    layer_info.height_z0 = h*ru/rd;
+                else
+                    // we calculate the height (=width) as  h =  (ru + 1/2 * h) * (2pi / n_cells)
+                    layer_info.height_z0 = ru*dd4hep::twopi/(0.5*layer_info.nwires - dd4hep::pi);
```
s6anloes:
```suggestion
                    layer_info.height_z0 = ru*dd4hep::twopi/(layer_info.ncells - dd4hep::pi);
```
`#ncells`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+
+                layer_info.radius_sw_z0 = 0.5*layer_info.height_z0 + ru;
+                layer_info.stereo_sw_z0  = StereoSign(layer_info) * this->stereoangle_z0(layer_info.radius_sw_z0);
+            }
+
+            //calculate radius_fdw_z0, radius_fuw_z0, width_z0
+            layer_info.radius_fdw_z0 = previousLayer.radius_fuw_z0;
+            layer_info.radius_fuw_z0 = previousLayer.radius_fuw_z0 + layer_info.height_z0;
+            layer_info.width_z0 = dd4hep::twopi*layer_info.radius_sw_z0/(0.5*layer_info.nwires);
```
s6anloes:
```suggestion
            layer_info.width_z0 = dd4hep::twopi*layer_info.radius_sw_z0/(layer_info.ncells);
```
`#ncells`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+        oss << "\tN layers = " << nlayers << '\n';
+        oss << "\n";
+        oss << "\tN cells layer1 = " << ncell0 << '\n';
+        oss << "\tN cells increment per superlayer = " << ncell_increment << '\n';
+        oss << "\tN cells per sector = " << ncell_per_sector << '\n';
+        oss << "\n";
+        oss << "Layer parameters of DCH:\n";
+        oss
+                << "\t" << "layer"
+                << "\t" << "nwires"
```
s6anloes:
```suggestion
                << "\t" << "ncells"
```
`#ncells`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+        {
+            oss << "\nDatabase empty\n";
+            return;
+        }
+
+        for(const auto& [nlayer, l]  : database )
+        {
+            oss
+                    << "\t" << l.layer
+                    << "\t" << l.nwires
```
s6anloes:
```suggestion
                    << "\t" << l.ncells
```
`#ncells`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+    //  STT-specific global parameters
+
+    /// number of sectors
+    std::vector<int> nsectors {};
+    /// Half angular distance between two consecutive tubes within the same layer, for each superlayer.
+    /// This also corresponds to the delta_phi staggering of tubes across layers.
+    std::vector<angle_t> delta_phi {};
+    /// stereo angle of superlayers
+    std::vector<angle_t> stereo {};
+    // radius of sensitive in wire in innermost layers of each superlayer
```
s6anloes:
```suggestion
    // placement radius of sensitive wire in innermost layers of each superlayer
```

detector/tracker/StrawTubeTracker_o1_v01_geo.cpp
```diff
@@ -146,46 +172,70 @@ static dd4hep::Ref_t create_straw_tracker(dd4hep::Detector& theDetector, xml_h e
...
-          double phi = l * 2 * dd4hep::pi / MLSectors +
-                       (j + 2 * i) * delta_phi * pow(-1, MLNum); // direction of diagonal gap changes per ML
-
-          std::string placedTubeName = MLName + dd4hep::_toString(l, "sector%d") + dd4hep::_toString(j, "layer%d") +
+    for (int j = 0; j < SLLayers; ++j) {
+      for (int l = 0; l < SLSectors; ++l) {
+        for (int i = 0; i < SLphiRepeat; ++i) {
+          // place the envelope volume in the superlayer envelope
+          double phi = l * 2 * dd4hep::pi / SLSectors +
+                       (j + 2 * i) * SLDelta_phi * pow(-1, SLNum); // direction of diagonal gap changes per SL
```
s6anloes: It would be good to use the phi function from `STT_info` here, so that the computation is not duplicated and making sure it's always the same.
`STT_info->Get_cell_phi_angle(...)`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+
+            for(int ilayer = 0; ilayer < this->nlayersPerSuperlayer; ++ilayer) {
+                // initialize empty object
+                LayerInfo layer_info;
+
+                // set parameters
+                layer_info.layer = layer;
+                // in consecutive layers, minimal radial distance between staggered tubes centers
+                // is cell_diameter * sqrt(3)/2 (=0.866)
+                layer_info.radius_sw_z0 = R + (ilayer * cell_diameter * 0.866);
```
s6anloes:
```suggestion
                // is cell_diameter * sqrt(3)/2
                layer_info.radius_sw_z0 = R + (ilayer * cell_diameter * sqrt(3)/2);
```

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+                layer_info.stereo_sw_z0 = sl_stereo;
+
```
s6anloes: `#ncells`
We can fill ncells for this layer then as well:
`ncells = nsectors.at(layer) * ntubespersector`
where we need to also store the number of tubes per sector in `STT_info`

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+                this->database.emplace(layer, layer_info);
+                layer++;
+            }
+        }
+
+        std::cout << "\t+ Total size of DCH database = " << database.size() << std::endl;
```
s6anloes:
```suggestion
        std::cout << "\t+ Total size of STT database = " << database.size() << std::endl;
```

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+        return;
+    }
+
+    // STT database dumper
+    void ShowDatabase(std::ostream & oss) const override final {
+        oss << "\n";
+        oss << "Global parameters of DCH:\n";
```
s6anloes:
```suggestion
        oss << "Global parameters of STT:\n";
```

detector/tracker/WireTracker_info.h
```diff
@@ -0,0 +1,608 @@
...
+        oss << "\tGas, half length/mm = " << Lhalf/dd4hep::mm << '\n';
+        oss << "\tGas, radius in/mm  = " << rin/dd4hep::mm << '\n';
+        oss << "\tGas, radius out/mm = " << rout/dd4hep::mm<< '\n';
+        oss << "\n";
+        oss << "\tN superlayers = " << nsuperlayers << '\n';
+        oss << "\tN layers per superlayer = " << nlayersPerSuperlayer << '\n';
+        oss << "\tN layers = " << nlayers << '\n';
+        oss << "\n";
+        oss << "Layer parameters of DCH:\n";
```
s6anloes:
```suggestion
        oss << "Layer parameters of STT:\n";
```

s6anloes (commented):
Hi Stefano,
thanks for this PR.
I have made a few points I think are worth improving:

- Having a variable number of layers per superlayer (for the STT)
- Storing `ncells` instead of `nwires` in the `LayerInfo` (not strictly needed, but might make things easier in the future)

I have also tagged a potential bug in the rotation of vectors for the distance to wire calculation.

Lastly, it is important to note that this PR breaks the `DCHdigi_v02` in [k4RecTracker](https://github.com/key4hep/k4RecTracker/blob/main/DCHdigi/include/DCHdigi_v02.h).
I've done some investigation and I don't see a good way to prevent this. The problem is that with this PR, the extension added to the geometry [here](https://github.com/key4hep/k4geo/pull/581/changes#diff-07680e343491b2fb4a363f8d5d186812e261356b8a4ffec03a5f416e271b5803L481) will be the new `DCH_info` while the digitiser expects the old one. Due to the inheritance from the base `WireTracker_info` the memory signature of the `DCH_info` class changes and there is a segmentation fault when running the digitiser.
This means this PR will need to be accompanied by a sister PR in k4RecTracker to fix these issues. We can discuss in the next Common Digitiser meeting what will need to be done.
Once this PR is open (no longer marked as draft) please bring attention to that matter to the reviewer.

I believe the rest of my comments is straight-forward, but perhaps I'm forgetting something.
If you have questions, feel free to ask

detector/tracker/WireTracker_info.h
sfranchel: This function was directly removed, as it is not really needed at this point.

detector/tracker/WireTracker_info.h
sfranchel: The field wires radii were dropped form `layer_info`  as they were not really needed/used.

detector/tracker/StrawTubeTracker_o1_v01_geo.cpp
sfranchel: good point, we should use that since we're implementing it.

detector/tracker/StrawTubeTracker_o1_v01_geo.cpp
sfranchel: Following the discussions we had during the meeting last week, your suggestion was implemented.

detector/tracker/WireTracker_info.h
sfranchel: added a short paragraph. If more info is required, please reopen this thread and let me know which parts should be covered more in detail.

sfranchel:
Thank you @s6anloes for the thorough review and great suggestions. All your comments should be addressed now, I'll finally mark this PR as ready.

The related PR in k4RecTracker is:
https://github.com/key4hep/k4RecTracker/pull/83
That one will require changes to k4geo to be merged first.

Another note for the reviewers: to import the `WireTracker_info` new structure outside k4geo, I have added a soft link to the `WireTracker_info.h` file in the `detectorCommon` folder, as I noticed that one is build as public library. But please let me know if there's another recommended way to do that.

detector/tracker/DriftChamber_o1_v02.cpp
s6anloes:
```suggestion
  det.addExtension<dd4hep::rec::WireTracker_info_struct>(DCH_i);
```
This change is needed to fix the GenFit test not succeeding in https://github.com/key4hep/k4RecTracker/pull/83/

sfranchel: Thanks @s6anloes for finding the problem with that.  I apply the same change to the Straw Tube Tracker as well.

s6anloes:
pinging @SanghyunKo for a review :heart:

detectorCommon/include/detectorCommon/WireTracker_info.h
```diff
@@ -0,0 +1 @@
+../../../detector/tracker/WireTracker_info.h
```
SanghyunKo: I didn't understand the purpose of this file. I see the following comment:
> Another note for the reviewers: to import the WireTracker_info new structure outside k4geo, I have added a soft link to the WireTracker_info.h file in the detectorCommon folder, as I noticed that one is build as public library. But please let me know if there's another recommended way to do that.

If the issue is to have this in the include directory, we should just place the `WireTracker_info` class under either `detector/include` or `detectorCommon/include/detectorCommon`, not under `detector/tracker` (or do something like `detector/tracker/include` and add `target_include_directories(${PackageName}  PRIVATE ${PROJECT_SOURCE_DIR}/detector/tracker/include )` in the `CMakeLists.txt` like what calorimeters are already doing).

SanghyunKo (commented):
I don't have any comment on the geometry side. I have only one comment for the usage of the header file.

detectorCommon/include/detectorCommon/WireTracker_info.h
sfranchel: Thank you @SanghyunKo for your feedback. I wasn't sure if there was a proper way of doing it, but I removed the soft link and moved the header file over directly in the `dectorCommon` folder. Hopefully that's fine for you.

SanghyunKo (approved):

SanghyunKo:
Sorry, I was occupied by several conferences and am now back to the PR review. Now it looks good to me.

PR #586, opened 2026-04-07
update readout used by FCChh ECAL

giovannimarchiori:
BEGINRELEASENOTES
- Restore readout used for simulation of FCC-hh ECAL
ENDRELEASENOTES

Needed to make FCC-hh digitisation scripts work - and more in line with legacy FCC-hh code in https://github.com/HEP-FCC/FCCDetectors/tree/main/Detector


scott-snyder:
Go ahead and update the README, then we can merge this.

PR #587, opened 2026-04-10
Make xml parameter names related to ECal endcap more uniform

varnes:

BEGINRELEASENOTES

Change the names of xml parameters related to the ECal endcap so that all of them begin with the prefix "EMEC".  This required corresponding changes in the geometry and segmentation codes.  

An entry was also added to the materials.xml file, corresponding to sintered tungsten (10% copper mixture), which might be used for the absorbers.

ENDRELEASENOTES


giovannimarchiori:
@varnes: runtime_error: dd4hep: Evaluator::Object : unknown variable : EMEC_inner_radius : value=EMEC_inner_radius

You need to modify also FCCee/ALLEGRO/compact/ALLEGRO_o2_v01/DectDimensions.xml

varnes:
Hi @giovannimarchiori , Ah I forgot to check what happens when running from ALLEGRO_o2_v01.  The changes I just pushed to DectDimensions.xml in that directory should fix it.

FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/ECalEndcaps_Turbine_o1_v03.xml
```diff
@@ -3,99 +3,103 @@
...
     <!-- total amount of steel in one passive plate: it is divided for the outside layer on top and bottom -->
     <constant name="EMEC_steel_thickness" value="0.1*mm"/>
      <!-- total amount of glue in one passive plate: it is divided for the outside layer on top and bottom -->
      <constant name="EMEC_glue_thickness" value="0.1*mm"/> 
-     <constant name="nUnitCells1" value="384"/>
-     <constant name="nUnitCells2" value="720"/>
-     <constant name="nUnitCells3" value="1360"/>
+     <constant name="EMECnUnitCells1" value="448"/>
+     <constant name="EMECnUnitCells2" value="832"/>
+     <constant name="EMECnUnitCells3" value="1584"/>
```
giovannimarchiori, line 89: Hi @varnes I thought this PR was supposed to only change names of some variables but I notice that there is this change here... is the granularity of the turbine different? After merging your PR we're starting to have failures in CI tests in k4RecoCalorimeter and in ALLEGRO repositories due to the neighbour maps (e.g. cells missing in map) I suspect it's due to this change.... should we keep it (and update the maps) or revert it and keep the old maps?

PR #593, opened 2026-04-30
Fix compilation warnings.

scott-snyder:
Unused variables.

BEGINRELEASENOTES
- Remove some unused variables about which gcc16 was warning.
ENDRELEASENOTES


jmcarcell:
Either from recent copy-pasting or warnings were not there before because it's not the first time

PR #594, opened 2026-05-07
Fix position calculation for FCCSWHCalPhi[Theta,Row]_k4geo.

scott-snyder:
The position() method for a segmentation should return a position in the
local coordinate system of the associated dd4hep volume in order for ddsim
to set the position properly.  Do this for FCCSWHCalPhi[Theta,Row]_k4geo.
It's a little complicated because the theta cell binning doesn't
correspond in any simple way with the dd4hep volumes; add information
on the correspondence to the CellInfo structure.

Verified that this does not change the reconstructed positions, and with
this change, the positions of the hits output by simulation match
the reconstructed cell positions.

This is similar to what was already done for the Allegro ECal barrel.

BEGINRELEASENOTES
- Fix position calculations for Allegro HCal segmentations so that the positions are correct on the hits output by the simulation.
ENDRELEASENOTES

### DIFF
--- a/detectorSegmentations/src/FCCSWHCalPhiRow_k4geo.cpp
+++ b/detectorSegmentations/src/FCCSWHCalPhiRow_k4geo.cpp
@@ -15,7 +15,7 @@ namespace DDSegmentation {
   void FCCSWHCalPhiRow_k4geo::commonSetup() {
     // define type and description
     _type = "FCCSWHCalPhiRow_k4geo";
-    _description = "Phi-theta segmentation in the global coordinates";
+    _description = "Phi-row segmentation for HCal";
 
     // register all necessary parameters
     registerParameter("phi_bins", "Number of bins phi", m_phiBins, 1);
@@ -73,7 +73,7 @@ namespace DDSegmentation {
     // pseudo-layers. Need to adjust the cell position:
     if (m_detLayout == 1 && !m_groupedRows.empty()) {
       int aidx = std::abs(idx);
-      zpos += 0.5 * m_dz_row * (li.groupedRows[aidx - 1] - m_gridSizeRow.at(layer));
+      zpos += 0.5 * m_dz_row * li.groupedRows[aidx - 1];
     }
 
     if (idx < 0) {
@@ -127,7 +127,9 @@ namespace DDSegmentation {
       // i_section to be grouped in a pseudo-layer.
       auto groupedRows_start = m_groupedRows.begin();
       auto groupedRows_end = m_groupedRows.end();
+      bool grouped = false;
       if (m_detLayout == 1 && !m_groupedRows.empty()) {
+        grouped = true;
         size_t nGroupedRows = m_groupedRows.size() / m_offsetZ.size();
         groupedRows_start = m_groupedRows.begin() + i_section * nGroupedRows;
         groupedRows_end = (i_section == m_offsetZ.size() - 1) ? m_groupedRows.end() : groupedRows_start + nGroupedRows;
@@ -141,7 +143,13 @@ namespace DDSegmentation {
         for (int i_lay = 0; i_lay < m_numLayers.at(i_dR + i_section * N_dR); i_lay++) {
 
           double volOffset = (layerInSection % 2) ? m_oddVolOffset : m_evenVolOffset;
-          double zOffset = m_dz_row * m_gridSizeRow.at(out.size()) * 0.5 - volOffset;
+          double zOffset = - volOffset;
+          if (!grouped) {
+            // Offset to middle of cell for this layer.
+            // But if we're using groupedRows, then this is row-dependent,
+            // and we apply this instead in position().
+            zOffset += m_dz_row * m_gridSizeRow.at(out.size()) * 0.5;
+          }
 
           moduleDepth[i_section] += m_dRlayer.at(i_dR);
           out.push_back(LayerInfo{.type = i_section,
@@ -805,9 +813,10 @@ namespace DDSegmentation {
     // For the endcap, we need to fill in the type field.  For the negative
     // endcap, types are offset by three, and we also need to make the row
     // index positive.
+    const LayerInfo* li = nullptr;
     if (m_detLayout == 1) {
-      const LayerInfo& li = getLayerInfo(layer);
-      int type = li.type;
+      li = &getLayerInfo(layer);
+      int type = li->type;
       if (irow < 0) {
         irow = -irow;
         type += 3;
@@ -819,11 +828,10 @@ namespace DDSegmentation {
     // volume indices start with 0!
     int vrow = 0;
     if (m_detLayout == 1 && !m_groupedRows.empty()) {
-      const LayerInfo& li = getLayerInfo(layer);
-
-      // Rows grouped according to groupedRows rather than by grid_size_row
+      // Rows grouped according to groupedRows rather than by gridSizeRow.
+      // li was fetched in the stanza above.
       for (size_t i = 1; i < static_cast<size_t>(irow); i++)
-        vrow += li.groupedRows[i - 1];
+        vrow += li->groupedRows[i - 1];
     } else {
       // Normal case.
       vrow = (irow - 1) * m_gridSizeRow.at(layer);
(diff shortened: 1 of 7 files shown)


scott-snyder:
Failure is because some of the downstream tests look at the xml files from the nightly build rather than from this update.

detectorSegmentations/src/FCCSWHCalPhiRow_k4geo.cpp
```diff
@@ -56,26 +62,23 @@ namespace DDSegmentation {
...
+    // pseudo-layers. Need to adjust the cell position:
     if (m_detLayout == 1 && !m_groupedRows.empty()) {
-      int nrows = 0;
-      for (size_t i = 0; i < static_cast<size_t>(std::abs(idx)); i++)
-        nrows += li.groupedRows[i];
-      zpos = minLayerZ + nrows * m_dz_row - 0.5 * li.groupedRows[abs(idx) - 1] * m_dz_row;
-      if (idx < 0)
-        zpos = -zpos;
+      int aidx = std::abs(idx);
+      zpos += 0.5 * m_dz_row * (li.groupedRows[aidx - 1] - m_gridSizeRow.at(layer));
```
Archil-AD: why do you subtract m_gridSizeRow.at(layer) here?
The m_gridSizeRow parameter is not used if m_groupedRows is not empty.

detectorSegmentations/src/FCCSWHCalPhiRow_k4geo.cpp
```diff
@@ -130,17 +133,26 @@ namespace DDSegmentation {
...
       // Loop over groups of layers.
       for (uint i_dR = 0; i_dR < N_dR; i_dR++) {
         // Loop over individual layers.
-        for (int i_lay = 0; i_lay < m_numLayers[i_dR + i_section * N_dR]; i_lay++) {
-          moduleDepth[i_section] += m_dRlayer[i_dR];
-          out.push_back(LayerInfo{.radius = moduleDepth[i_section] - m_dRlayer[i_dR] * 0.5,
-                                  .halfDepth = m_dRlayer[i_dR] / 2,
+        for (int i_lay = 0; i_lay < m_numLayers.at(i_dR + i_section * N_dR); i_lay++) {
+
+          double volOffset = (layerInSection % 2) ? m_oddVolOffset : m_evenVolOffset;
```
Archil-AD, line 145: here you have hardcoded that the first layer always has "m_evenVolOffset" in all phi modules.

But we are considering the alternative geometry where the the same layer in the odd and even phi modules consists of different sequences (sequence_a or sequence_b). In this alternative geometry, scintillators from neighbouring phi modules are not placed side-by-sided which would avoid a possible effect of cross-talk between two neighbouring cells from two adjacent phi modules.

So, the current definition of volOffset would not work for the alternative geometry.

would it be possible to make it more generic?

detectorSegmentations/src/FCCSWHCalPhiRow_k4geo.cpp
```diff
@@ -779,5 +791,47 @@ namespace DDSegmentation {
...
+
+    // Get layer and row.
+    uint layer = decoder()->get(cID, m_layerIndex);
+    int irow = decoder()->get(vID, m_rowIndex);
+
+    // For the endcap, we need to fill in the type field.  For the negative
+    // endcap, types are offset by three, and we also need to make the row
+    // index positive.
+    if (m_detLayout == 1) {
+      const LayerInfo& li = getLayerInfo(layer);
```
Archil-AD: this could be moved outside if() to avoid retrieval of the LayerInfo second time after few lines later.

andresailer:
For the release notes:

> Fix position calculations for HCal segmentations so that the positions are correct on the hits output by the simulation.

Could you specify the specific hcals concerned in this PR, please?

scott-snyder:
Clarified the release notes to specify the Allegro HCal.

detectorSegmentations/src/FCCSWHCalPhiRow_k4geo.cpp
scott-snyder: That's there because the per-layer offset was earlier calculated as

```
          double volOffset = (layerInSection % 2) ? m_oddVolOffset : m_evenVolO
          double zOffset = m_dz_row * m_gridSizeRow.at(out.size()) * 0.5 - volOffset;
```

So in the grouped case, that gridsize term effectively cancels out the
gridsize term above.

Maybe it's clearer to make the initialization of zOffset conditional on whether
grouping is done, removing the gridsize term in that case.  Then we don't
need to subtract it off in position().  I'll make that change, thanks.

detectorSegmentations/src/FCCSWHCalPhiRow_k4geo.cpp
scott-snyder: Thanks; restructured to avoid that.

detectorSegmentations/src/FCCSWHCalPhiRow_k4geo.cpp
scott-snyder: We can't derive that information from the geometry, because createHCalEC
calls position().  So we'd have to pass it in as parameters to the segmentation.
But it seems premature to do that until it's known exactly how that's going
to be structured; otherwise, we risk implementing the wrong thing and
having to do it again.

I guess i could go ahead and change zOffset to be an array indexed
by phi, but there seems little point doing that before it's actually needed.

So i think it's best to wait until the geometry is actually changing.

giovannimarchiori: @Archil-AD OK with this proposal?

Archil-AD: Hi @giovannimarchiori and @scott-snyder ,  I'm ok with proposal.

However, with these changes  I am not able to run the reconstruction with phi-row segmentation for Endcap that I am using for the PandoraPFA.
The code crashes with:
VolumeManager    ERROR lookupContext: Failed to search Volume context 000004000BE00FD9 [Unknown identifier]

I am not sure if this crash is related, but clearly the pseudo-layers implementation for the endcap would not work properly with these changes because the segmentation class now returns the local position while here the global position is needed:
https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/HCalThreePartsEndcap_o1_v02_geo.cpp#L648

maybe I can try to implement calculation of the position without segmentation class, but first the crash for phi-row segmentation should be understood.

scott-snyder: Thanks.  Can you give me a recipe to reproduce the crash?

Archil-AD: Hi @scott-snyder ,  

apologies, I could not reproduce this error... probably I got it because I was trying to adjust the code for pandora...

so, these changes works for both phi-theta and phi-row segmentation with only one exception that it breaks the pseudo-layer definition for PandoraPFA.

I can try to fix it once this is merged.

Thanks a lot for the changes!

detectorSegmentations/src/FCCSWHCalPhiRow_k4geo.cpp
```diff
@@ -124,23 +127,40 @@ namespace DDSegmentation {
...
       for (uint i_dR = 0; i_dR < N_dR; i_dR++) {
         // Loop over individual layers.
-        for (int i_lay = 0; i_lay < m_numLayers[i_dR + i_section * N_dR]; i_lay++) {
-          moduleDepth[i_section] += m_dRlayer[i_dR];
-          out.push_back(LayerInfo{.radius = moduleDepth[i_section] - m_dRlayer[i_dR] * 0.5,
-                                  .halfDepth = m_dRlayer[i_dR] / 2,
+        for (int i_lay = 0; i_lay < m_numLayers.at(i_dR + i_section * N_dR); i_lay++) {
+
+          double volOffset = (layerInSection % 2) ? m_oddVolOffset : m_evenVolOffset;
+          double zOffset = - volOffset;
```
giovannimarchiori:
```suggestion
          double zOffset = -volOffset;
```

giovannimarchiori:
Since @Archil-AD has approved these changes I am going to merge this PR

PR #595, opened 2026-05-07
Fix warnings seen compiling with clang.

scott-snyder:
Unused variables / members.

BEGINRELEASENOTES
- Fix some warnings seen when compiling with clang.
ENDRELEASENOTES


jmcarcell:
Clang 22?

scott-snyder:
clang 20 actually.  clang finds more of such unused variables than gcc does (but by contract, gcc16 finds unused loop variables that clang does not).

PR #597, opened 2026-05-13
[ILD] Explain difference between sim and reco detector models in README

Victor-Schwan:
BEGINRELEASENOTES
- ILD Documentation: Add a paragraph on the difference between *simulation* and *reconstruction* detector models

ENDRELEASENOTES


Victor-Schwan:
As discussed in a meeting earlier this week with @gaede

ILD/compact/README.md
```diff
@@ -4,6 +4,15 @@
 The following ILD detector models are available in lcgeo ( current production models **ILD_ls5_v02** )
 
 
+In this table, *simulation* models are intended to be used with:
+```bash
+ddsim --compactFile $k4geo_DIR/ILD/compact/ILD_l5_v02/ILD_l5_v02.xml ....
```
jmcarcell:
```suggestion
ddsim --compactFile $K4GEO/ILD/compact/ILD_l5_v02/ILD_l5_v02.xml ....
```
I think we wanted to stop using `k4geo_DIR`.

ILD/compact/README.md
```diff
@@ -4,6 +4,15 @@
...
+```
+They can produce hit collections for multiple calorimeter technology options in a single run (to reduce computing cost). The *reconstruction* models are intended for:
+```bash
+k4run ILDReconstruction.py --compactFile $k4geo_DIR/ILD/compact/ILD_l5_v02/ILD_l5_o1_v02.xml ...
```
jmcarcell:
```suggestion
k4run ILDReconstruction.py --compactFile $K4GEO/ILD/compact/ILD_l5_v02/ILD_l5_o1_v02.xml ...
```
Same as above

ILD/compact/README.md
```diff
@@ -4,6 +4,15 @@
...
+```
+, where the reconstruction configuration must select which calorimeter option should be used to reconstruct the event from the corresponding hits. Simulation-only models must not be used for reconstruction.
```
jmcarcell:
```suggestion
where the reconstruction configuration must select which calorimeter option should be used to reconstruct the event from the corresponding hits. Simulation-only models must not be used for reconstruction.
```
this will be its own new line

ILD/compact/README.md
jmcarcell: Should the simulation and reconstruction models be different? They are different in your text.

Victor-Schwan: Yes! This is the whole point of the PR! I’m trying to explain that the models must be different, or rather that *simulation* models must not be used for the *reconstruction*. Is it worded in such a way that it’s easy to misunderstand?

jmcarcell: I think it's clear but not intuitive when for all other detectors we use the same model for simulation and reconstruction, or I think so. What if you use the one you are going to use for reconstruction for simulation?

Victor-Schwan: AFAIK, this does work but you'll only have the sim hits for its specific ECAL and HCAL choice. If you do use the sim model, you'll get sim hits for both ECAL and both HCAL options.

Victor-Schwan: With other "detectors" you refer to detector concepts as in IDEA, CLD, ... and not other ILD models, don't you? This is exactly why I added the explanation :wink:

PR #605, opened 2026-06-13
Optional geant4 plugin

madbaron:

BEGINRELEASENOTES
- Added a CMake flag (default behaviour is unchanged) to allow compiling the package without geant4 dependencies

ENDRELEASENOTES

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,6 +228,14 @@ INSTALL(EXPORT ${PROJECT_NAME}Targets
 
 include(CMakePackageConfigHelpers)
 
+# Only pull Geant4 into the exported config when k4geo was built against it,
+# otherwise downstream find_package(k4geo) fails in a Geant4-free environment.
+if(K4GEO_USE_GEANT4)
+  set(K4GEO_GEANT4_FIND_DEPENDENCY "find_dependency(Geant4 REQUIRED)")
+else()
+  set(K4GEO_GEANT4_FIND_DEPENDENCY "# Geant4 dependency omitted: k4geo built with K4GEO_USE_GEANT4=OFF")
+endif()
+
 configure_package_config_file(cmake/k4geoConfig.cmake.in ${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_CMAKEDIR}/k4geoConfig.cmake
   PATH_VARS CMAKE_INSTALL_CMAKEDIR
   INSTALL_DESTINATION ${CMAKE_INSTALL_CMAKEDIR} )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,7 +184,12 @@ else()
 endif()
 
 add_subdirectory(detectorSegmentations)
-add_subdirectory(detectorCommon)
+# detectorCommon provides Geant4-based digitization utilities (DetUtils_k4geo
+# uses DDG4/Geant4Mapping and G4Step) and links DD4hep::DDG4, so it is only
+# built when Geant4 support is enabled. Nothing else in k4geo links it.
+if(K4GEO_USE_GEANT4)
+  add_subdirectory(detectorCommon)
+endif()
 
 # Destination directories are hardcoded because GNUdirectories are not included
 set(k4geo_install_targets ${PackageName})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,15 +44,23 @@ ENDIF(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 option(BUILD_TESTING "Enable and build tests" ON)
 option(INSTALL_COMPACT_FILES "Copy compact files to install area" OFF)
 option(INSTALL_BEAMPIPE_STL_FILES "Download CAD files for building the detailed beampipe" OFF)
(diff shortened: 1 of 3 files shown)


jmcarcell:
I don't really like this. I don't think it compiles, there are more places where DDG4 is used. Tests also are not run. Then we do not have a way of testing this since all of our stacks have Geant4. Do you need to run without Geant4?

madbaron:
I'm using this in a lightweight container without geant. It compiles and works happily.

Sorry are there extra tests?
I thought the CI would run everything that is needed

jmcarcell:
What I mean is that it is very likely someone adds some code that depends on Geant4 but it's not guarded by your checks and then it breaks your workflow since our CI tests will be fine. Also, enabling that gives you a repository with much of the functionality reduced (I think you can not run any test). What is the purpose, saving space or time or both?

madbaron:
Thanks, that's clearer now!

We have a non-negligible fraction of the user base that does development (mostly of either analysis or algorithms downstream of the simulation) on mobile devices by pulling the docker images. 
The community asked for shrinking these down and possibly building more "application tailored" ones, and this PR is part of that effort. We are now test-shipping an "analysis" image (which essentially has only podio+edm4hep+python tools) a "sim" (which has most of the stack) and an "ml" image (which adds torch etc to the stack).

This change is aimed at building a digi/reco image (layered in between analysis and sim) without G4+data. 

I could add a required test (a-la https://github.com/key4hep/k4ActsTracking/blob/main/.github/workflows/mucoll-ci.yml) that pulls the digi/reco image and builds the package there if you like, so that any unguarded G4 dependency will be caught? Otherwise I'm also happy to take care of any fixes when they are needed

jmcarcell:
Ok so to go ahead with this I would like a few things:
- Not making it more complex. I don't have concrete suggestions right now but we should try to follow this or the CMake configuration can be very complicated.
- Can you check if the tests can be run without G4? I've never run `ddsim` without G4 so I don't know if it works. In case it doesn't, all the tests that can't be run should be skipped.
- We should also support building with and without LCIO, does that work? Maybe you want to build also without LCIO?

For testing that image may be the only option because a stack where DD4hep is not compiled against G4 is needed since otherwise DD4hep will try to find Geant4 and set the targets: https://github.com/AIDASoft/DD4hep/blob/master/cmake/DD4hepConfig.cmake.in#L51, so that could be an option.

I just realised that the split to the G4 dependencies only was done many years ago, @andresailer was it ever envisioned not to build with G4? Was the split done not to have everything link to G4?

andresailer:
> was it ever envisioned not to build with G4? Was the split done not to have everything link to G4?

Not building with Geant4 was not the goal. The split was done to not link reconstruction code against Geant4 libraries.

CMakeLists.txt
```diff
@@ -201,6 +228,14 @@ INSTALL(EXPORT ${PROJECT_NAME}Targets
 
 include(CMakePackageConfigHelpers)
 
+# Only pull Geant4 into the exported config when k4geo was built against it,
```
andresailer, line 231: It would be better to put this to where find_package(Geant4 REQUIRED) is called, because if we ever need to change that find_package we probably also have to change the find_dependency.

madbaron:
> * Can you check if the tests can be run without G4? I've never run `ddsim` without G4 so I don't know if it works. In case it doesn't, all the tests that can't be run should be skipped.

Without G4, ddhep is compiled with ~ddg4, which means there's no ddsim at all. I've now added some guards in the test CMakefile to skip the tests relying on ddsim (which essentially leaves us with only the name checks).

> * We should also support building with and without LCIO, does that work? Maybe you want to build also without LCIO?

I haven't tried this one yet, as we still want to support the workflows of the groups that are still partly on pure Marlin or the Marlin-wrapper. However I see there's a flag for this already, so it might be good to go?

> For testing that image may be the only option because a stack where DD4hep is not compiled against G4 is needed since otherwise DD4hep will try to find Geant4 and set the targets: https://github.com/AIDASoft/DD4hep/blob/master/cmake/DD4hepConfig.cmake.in#L51, so that could be an option.

I'll add a test based on the image soon.

For reference, I am using this image ghcr.io/muoncollidersoft/mucoll-sim-ubuntu24:experimental_reco_image in case anybody else would like to take a look.

jmcarcell:
Now `detectorCommon`, that depends on G4, is linked in the main plugin (from a PR that someone merged), which makes this approach not work anymore. This could (maybe) be circumvented by `target_include_directories` instead of linking, but chasing these issues is going to be annoying if not require significant refactors. I'm not sure how to continue with this long-term.

madbaron:
Thanks @jmcarcell, I'll close this PR.
We also realised that the reco-only stack would need to trigger some significant changes throughout the stack to be fully usable the way the community would like to and that's probably unsustainable.

jmcarcell:
I'm happy to add some options to disable stuff; it's also convenient for me sometimes because I build locally most of the stack for development, but Geant4 is hard to do. For building Geant4 specifically, you should build with `-DGEANT4_INSTALL_EXAMPLES=OFF` that I don't see in the Spack recipe and it is ON by default in Geant4; this will remove like 10k files. In addition, you can use `strip` on binaries to reduce their size a bit more (stack traces become completely useless). Also all the datasets are not needed if you are not running Geant4; I think only `G4ENSDFSTATEDATA` makes it crash when building downstream projects if not present.

PR #607, opened 2026-06-18
Resize ALLEGRO DCH

BrieucF:
BEGINRELEASENOTES
- [FCCee] Put envelope DCH parameters in DectDimensions and resize ALLEGRO DCH

ENDRELEASENOTES

In this version, I put the DCH sensitive length the same as the Straw Tube one, there is enough space to accommodate the DCH services. NB: a more thorough re-engineering would be needed to have a realistic detector layout but this will come later. The goal here is just to have a tracking system covering the available space inside the calorimeters.

Before the change:
Screenshot 2026-06-19 at 2 37 02 PM

After the change:
Screenshot 2026-06-19 at 2 35 54 PM

### DIFF
--- a/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/DectDimensions.xml
+++ b/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/DectDimensions.xml
@@ -110,7 +110,9 @@
     <constant name="DCH_outer_cyl_R_total"          value=" 2015   * mm "  />
     <constant name="DCH_gas_outer_cyl_R"          value=" 2000 * mm "  />
     <constant name="DCH_half_length_total"          value=" 3000   * mm "  />
-    <constant name="DCH_gas_Lhalf"                value=" 2750 * mm "  />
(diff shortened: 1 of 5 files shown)


giovannimarchiori:
Thanks @BrieucF ! For IDEA the changes have no impact, so just report about the results of the overlap checks for ALLEGRO when ready and then we merge

s6anloes:
Would you consider also updating the twist angle? With an increase of `z` length, this means that the stereo angle is reduced if the twist angle stays the same.

BrieucF:
> Would you consider also updating the twist angle? With an increase of `z` length, this means that the stereo angle is reduced if the twist angle stays the same.

Good point, I did it, even though a more thorough re-engineering would be needed for the new z extent.

giovannimarchiori (approved):
Looks good, thanks!

jbeirer:
Screenshot 2026-06-26 at 20 56 49


@BrieucF Just wanted to point out that I assume it is this PR that is causing the significant increase in RSS and wall time

BrieucF:
Hi Joshua, thanks for reporting this! This PR just resized some sub-detectors, I assume this translates in additional volume, probably from the Silicon wrapper.

jbeirer:
Sorry for the noise, this was just a test for the k4bench bot, which seems to do its job :)

PR #608, opened 2026-06-23
Fix calculation of sim hit positions in turbine ecal endcap

varnes:

BEGINRELEASENOTES

Implements the correct position calculation where the position() method of the segmentation returns the location of the segmented readout cell with respect to its parent G4 volume.  Since there are some calculation steps involved for one of the coordinates, the calculations are done on the first call and then cached to speed things up.  Ideally this would be done in the commonSetup() method, but some needed quantities (m_offsetRho, m_offsetZ) are not initialized before commonSetup() is run.

Needs to be syncronized with corresponding k4RecCalorimeter PR (https://github.com/HEP-FCC/k4RecCalorimeter/pull/264) so that the reconstructed cell positions are also correct.

ENDRELEASENOTES


scott-snyder:
Thanks for doing this Erich!

I presume that you've validated that after the two MRs:
  (a) Reconstructed cell positions are unchanged.
  (b) Hit positions match reconstructed cell positions.

If we're revamping this class, it would be a good idea to also make
sure that volumeID() works correctly (the default base-class implementation
generally does not).  Hopefully, something simple would work like in
FCCSWGridModuleThetaMerged_k4geo.cpp.
While this isn't strictly necessary if cellsSpanVolumes() returns false
(meaning that no cell encompasses multiple geant volumes), it would probably
be a good idea to prevent surprises.

I've noted some more minor issues below.  But in particular:
 - Prefer to use functions from the C++ standard library rather than TMath.
 - Any use of mutable needs to be explicitly thread-safe.

detectorSegmentations/include/detectorSegmentations/FCCSWEndcapTurbine_k4geo.h
```diff
@@ -200,6 +166,8 @@ namespace DDSegmentation {
   private:
     /// turbine blade angle in each wheel
     std::vector<double> m_bladeAngle;
+    /// sine of turbine blade angle in each wheel
+    std::vector<double> m_sinBladeAngle;
```
scott-snyder: You only divide by this.  So it would be better to here tabulate the reciprocal (csc) so we can multiply rather than dividing.

detectorSegmentations/include/detectorSegmentations/FCCSWEndcapTurbine_k4geo.h
```diff
@@ -223,6 +197,10 @@ namespace DDSegmentation {
     int m_rhoBins;
     ////grid size in rho
     std::vector<double> m_gridSizeRho;
+    /// "effective" grid size in rho for cells away from the center strip in z
+    mutable std::vector<std::vector<double>> m_effGridSizeRho;
```
scott-snyder: This doesn't appear to be used.

detectorSegmentations/include/detectorSegmentations/FCCSWEndcapTurbine_k4geo.h
```diff
@@ -223,6 +197,10 @@ namespace DDSegmentation {
...
+    /// vector of calculated local z positions.  Indices are [iWheel][iRho][iZ]
+    mutable std::vector<std::vector<std::vector<double>>> m_localZPositions;
```
scott-snyder: This should be handled in a way that's thread-safe.  You can look at what i did for the emb/hcal cases.

It would also be more efficient to flatten the last two indices.  I might introduce a small helper class, WheelInfo or something, that holds the position data for one wheel as a flat std::vector and has a method for doing lookup by rho/z.  You then have a vector of these WheelInfo objects.

detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
```diff
@@ -164,18 +164,33 @@ namespace DDSegmentation {
     m_zIndex = decoder()->index(m_zID);
     m_sideIndex = decoder()->index(m_sideID);
     m_layerIndex = decoder()->index(m_layerID);
+
+    // precalculate some quantities that will be used to determine positions
+    for (const auto& bladeAngle : m_bladeAngle) {
+      m_sinBladeAngle.push_back(TMath::Sin(bladeAngle));
```
scott-snyder: Should use std::sin rather than TMath.  Better to write the type explicitly rather than using auto.

detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
```diff
@@ -164,18 +164,33 @@ namespace DDSegmentation {
...
+    }
+    unsigned iWheel = 0;
+    for (const auto& readoutRhoLayers : m_numReadoutRhoLayers) {
```
scott-snyder: Better to have iWheel declared locally in the for loop.
```
for (unsigned iWheel = 0; const ...
```

Though actually, all three of these loops are iterating over wheels.  It would probably be clearer to collapse them to a single loop over iWheel and then just use indexing.

detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
```diff
@@ -220,46 +235,154 @@ namespace DDSegmentation {
...
+    CellID iSide = decoder()->get(cID, m_sideIndex);
+    
     double phiCent = twopi * (iModule + 0.5) / (m_nUnitCells[iWheel] / m_mergedModules[iWheel]);
-    double rhoLoc = rho(cID);
+    double rho = getGlobalRho(cID);
 
     double zdepth = m_numReadoutZLayers[iWheel] * m_gridSizeZ[iWheel];
 
-    double zLoc = TMath::Abs(z(cID)) - m_offsetZ[iWheel] - zdepth / 2;
+    double zFromCent = TMath::Abs(getGlobalZ(cID)) - m_offsetZ[iWheel] - zdepth / 2;
```
scott-snyder: Use std::abs (making sure that <cmath> is included).

detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
```diff
@@ -220,46 +235,154 @@ namespace DDSegmentation {
...
 
     // calculation position in frame with unit cell at phi = 0
-    double y = zLoc / TMath::Tan(m_bladeAngle[iWheel]);
-    double x = TMath::Sqrt(rhoLoc * rhoLoc - y * y);
+    double y = zFromCent / TMath::Tan(m_bladeAngle[iWheel]);
+    double x = TMath::Sqrt(rho * rho - y * y);
```
scott-snyder: Use std::tan/std::sqrt.
If you've already calculated sin(angle), you should be able to have something more efficient than tan here.

detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
```diff
@@ -220,46 +235,154 @@ namespace DDSegmentation {
...
     double locPhi = TMath::ATan2(y, x);
 
     // now rotate by phi position of the unit cell
     double phi = locPhi + phiCent;
     if (phi > TMath::Pi())
```
scott-snyder: Should use std::numbers::pi.

detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
```diff
@@ -220,46 +235,154 @@ namespace DDSegmentation {
...
+            int jCalibZ = jZ / m_gangedZLayers[jWheel];
+            float calibZPos[3];
+            calibZPos[0] = (jCalibZ - m_numCalibZLayers[jWheel] / 2) * m_gridSizeZ[jWheel] * m_gangedZLayers[jWheel] +
+                           (1 - m_numCalibZLayers[jWheel] % 2) * m_gridSizeZ[jWheel] * m_gangedZLayers[jWheel] / 2;
+            calibZPos[1] = calibZPos[0] + m_gridSizeZ[jWheel] * m_gangedZLayers[jWheel] / 2;
+            calibZPos[2] = calibZPos[1] - m_gridSizeZ[jWheel] * m_gangedZLayers[jWheel];
+
+            float calibZmaxArr[3], calibZminArr[3];
+            for (int jPos = 0; jPos < 3; jPos++) {
+              if (calibRhoMax > TMath::Abs(calibZPos[jPos])) {
```
scott-snyder: std::abs/std::sqrt

detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
```diff
@@ -220,46 +235,154 @@ namespace DDSegmentation {
...
+              } else {
+                calibZmaxArr[jPos] = 0;
+              }
+              if (calibRhoMin > TMath::Abs(calibZPos[jPos])) {
+                calibZminArr[jPos] = TMath::Sqrt(calibRhoMin * calibRhoMin - calibZPos[jPos] * calibZPos[jPos]);
+              } else {
+                calibZminArr[jPos] = 1000000;
+              }
+            }
+            float calibZmax = TMath::MaxElement(3, calibZmaxArr);
```
scott-snyder:
```
float calibZmax = std:::ranges::max_element(calibZmaxArr);
```

scott-snyder: And so forth for other TMath usages.

detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
```diff
@@ -220,46 +235,154 @@ namespace DDSegmentation {
...
+            readoutZPos[1] = readoutZPos[0] + m_gridSizeZ[jWheel] / 2;
+            readoutZPos[2] = readoutZPos[1] - m_gridSizeZ[jWheel];
+
+            float readoutZmaxArr[3], readoutZminArr[3];
+            for (int jPos = 0; jPos < 3; jPos++) {
+              if (readoutRhoMax > TMath::Abs(readoutZPos[jPos])) {
+                readoutZmaxArr[jPos] =
+                    TMath::Sqrt(readoutRhoMax * readoutRhoMax - readoutZPos[jPos] * readoutZPos[jPos]);
+              } else {
+                std::cout << "Odd: readoutRhoMax < readoutZPos " << readoutRhoMax << " " << readoutZPos[jPos]
```
scott-snyder: We shouldn't we writing to cout here.

varnes:
Hi @scott-snyder ,

  Thanks for the careful review.  Before going through the specific comments (all look reasonable to me at first glance) I want to address your first point below:

> I presume that you've validated that after the two MRs: (a) Reconstructed cell positions are unchanged. (b) Hit positions match reconstructed cell positions.

The answer is "almost".  The positions agree perfectly in z, but are slightly different in x and y.  This is likely related to the fact that the geometry is more complex in those directions.  The differences between the new and old x and y positions is ~1cm, which is less than the size of a cell.  I have not been able to track down the origin of these differences, but thought it would be good to make this PR in any case since it corrects the large (several meter) errors in the simulated hit positions.  

Please let me know if you'd prefer I put this in draft status to continue investigating the x and y position discrepancies.

detectorSegmentations/include/detectorSegmentations/FCCSWEndcapTurbine_k4geo.h
varnes: Yep, leftover from a dead end during development.  Now removed.

detectorSegmentations/include/detectorSegmentations/FCCSWEndcapTurbine_k4geo.h
varnes: Hi @scott-snyder Just for my understanding, in what you describe it sounds like there would need to be a calculation to find the correct index in the flat vector (idx = irho*iz or similar).  Is that still more efficient than implementing a multi-dimensional vector?

scott-snyder: Almost certainly, though exactly how much depends on the access pattern.  But in general, chasing pointers is expensive compared with a little integer arithmetic.

scott-snyder:
Thanks for the updates!

As for the position discrepancies, you're the one who best knows the EMEC geometry, so I leave it up to you as to the best way to proceed.

That said, a ~1cm discrepancy is large enough to probably indicate a real bug, not just a numeric issue, and going from something which is obviously wrong (and has thus been ignored) to something which may be subtly wrong may not always be the best tradeoff.

However, i don't think it's an immediate concern, so you should decide what will make it easier to chase this down.

varnes:
> Thanks for the updates!
> 
> As for the position discrepancies, you're the one who best knows the EMEC geometry, so I leave it up to you as to the best way to proceed.
> 
> That said, a ~1cm discrepancy is large enough to probably indicate a real bug, not just a numeric issue, and going from something which is obviously wrong (and has thus been ignored) to something which may be subtly wrong may not always be the best tradeoff.
> 
> However, i don't think it's an immediate concern, so you should decide what will make it easier to chase this down.

Hi @scott-snyder I'd prefer to go ahead with this PR and then continue to work on the ~1cm discrepancies. One quick question: from inside of the position() method of the segmentation, is there a way to tell the extent of the volume in all three coordinates?  It's not trivial since my volumes are oddly-shaped, with elliptical edges.

BrieucF:
Hi @scott-snyder @varnes shall we then rebase this, merge and fix the ~1cm discrepancy in another PR?

varnes:
> Hi @scott-snyder @varnes shall we then rebase this, merge and fix the ~1cm discrepancy in another PR?

Hi @BrieucF That would be my preference, yes.

BrieucF:
Ok, can you fix the conflicts?

varnes:
> Ok, can you fix the conflicts?

Yes, I will do this when I am back from vacation next week.

varnes:
Hi @BrieucF I have resolved the conflict.  Some of the checks failed afterwards, but from the log files it seems to me that this was a problem in the system, not in the k4geo code.

giovannimarchiori:
HI @varnes and @scott-snyder 
- I checked that Erich implemented Scott's suggestions and cleared the open conversations
- Erich, you have pre-commit errors, you need to format the code properly, as suggested in the output of the failing test, i.e. run 
`clang-format --style=file -i detectorSegmentations/include/detectorSegmentations/FCCSWEndcapTurbine_k4geo.h` 
and then commit the updated code
- I am rerunning the alma9 nightly tests hopefully the should run fine with today's nightly

After these steps  it should be OK to merge this PR and then we'll need the fix for this ~1cm discrepancy. 
Erich please let us know if https://github.com/HEP-FCC/k4RecCalorimeter/pull/264 should wait for the fix to the discrepancy or if the two PRs can proceed in parallel once this is merged

varnes:
Hi @giovannimarchiori The formatting problems are now fixed, and all of the checks pass.  This should go forward in parallel with https://github.com/HEP-FCC/k4RecCalorimeter/pull/264 while I continue to investigate the ~1 cm discrepancies.

giovannimarchiori:
OK, if @scott-snyder is satisfied we can merge

varnes:
I did have one question for @scott-snyder that might help me in tracking down the ~1cm issues:  from inside of the position() method of the segmentation, is there a way to tell the extent of the volume in all three coordinates? It's not trivial since my volumes are oddly-shaped, with elliptical edges.

varnes:
Hi @giovannimarchiori @scott-snyder Just FYI that the latest change to FCCSWEndcapTurbine_k4geo fixes the ~1 cm problem in the radial (theta) direction.

giovannimarchiori (approved):
comments by @scott-snyder have been implemented and positions are identical to old code after latest commits - approving

jmcarcell:
This broke k4RecCalorimeter (https://github.com/HEP-FCC/k4RecCalorimeter/pull/279). Ideally the dowstream-build workflow would have told us but I need to fix it...
Edit: AH I see it's there in the notes

giovannimarchiori:
There is this PR for that purpose… https://github.com/HEP-FCC/k4RecCalorimeter/pull/264

jmcarcell:
Yes but if one is merged without the other the nightlies won't build and can't be published; CI will never pass in the other repository.

PR #609, opened 2026-06-25
add method to return number of wheels for Turbine endcap segmentation

giovannimarchiori:
BEGINRELEASENOTES
- add method to return number of wheels for Turbine endcap segmentation

ENDRELEASENOTES

### DIFF
--- a/detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
+++ b/detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
@@ -40,6 +40,12 @@ namespace DDSegmentation {
     m_bladeAngle.clear();
 
     try {
+      m_numWheels.push_back(dd4hepgeo->constant<double>("EMECnWheels"));
+    } catch (...) {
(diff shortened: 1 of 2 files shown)


detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp
```diff
@@ -39,6 +39,12 @@ namespace DDSegmentation {
 
     m_bladeAngle.clear();
 
+    try {
+      m_numWheels = dd4hepgeo->constant<int>("EMECnWheels");
+    } catch (...) {
+      std::cout << "EMECnWheels not found in detector metadata, exiting..." << std::endl;
+      exit(1);
```
BrieucF, line 46: Use throw instead?

giovannimarchiori: I used the structure of all other try ... blocks in the following lines - should I change all of them?

BrieucF: Indeed, since it is already used everywhere in this file let's leave it like that

PR #611, opened 2026-06-30
Prerequisites for IDEA option2 PFlow study

SanghyunKo:

BEGINRELEASENOTES
- Correct the `DetType` flag of the IDEA o2 fiber HCAL from ECAL to HCAL, with the correct barrel and endcap flag
- Add the `DetType` flag for the IDEA o2 crystal ECAL
- Add the `dd4hep::rec::LayeredCalorimeterData` for the IDEA o2 ECAL & HCAL
- Fixing the bug that `nGamma` in the IDEA o2 crystal ECAL wasn't set properly in the endcap
- Add the truth Edep collection for the IDEA o2 fiber HCAL (needed for the truth linking)
- Properly populate the calo hit contributions for the IDEA o2 crystal ECAL (previously empty, needed for the truth linking)
- Add a CI test for IDEA o2 with a partial geometry of (Δθ, Δφ) = (30 deg, 30 deg) of ECAL & HCAL

ENDRELEASENOTES

A bunch of code work needed for the IDEA o2 PFlow reconstruction downstream. I decided to pack all the changes into a single PR because they highly interfere with each other. This is one of many PRs that I'll create for the IDEA PFlow, and it's the most upstream change. No changes to the detector itself.


SanghyunKo:
Added CI test with the partial geometry (~6 GB memory, ~3 min execution time). The geometry loads roughly (Δθ, Δφ) = (30 deg, 30 deg) of ECAL & HCAL in the center of the barrel, omitting the endcap and the timing layer. 

screenshot_partialGeometry

detector/calorimeter/dual-readout-tubes/src/DRBarrelTubes_o1_v01.cpp
```diff
@@ -44,6 +45,17 @@ static Ref_t create_detector(Detector& description, xml_h entities, SensitiveDet
   barrel_placed.addPhysVolID("system", det_id);
   s_detElement.setPlacement(barrel_placed);
 
+  // apply <type_flags> and expose r-z extent for downstream reco
+  dd4hep::xml::setDetectorTypeFlag(entities, s_detElement);
+
+  dd4hep::rec::LayeredCalorimeterData* caloData = new dd4hep::rec::LayeredCalorimeterData;
```
jbeirer, line 52:
```suggestion
// DD4hep DetElement takes ownership and deletes the extension on destruction
dd4hep::rec::LayeredCalorimeterData* caloData = new dd4hep::rec::LayeredCalorimeterData;
```

This ownership transfer is not obvious from the raw new. Could we document that the DetElement extension mechanism takes ownership?

detector/calorimeter/dual-readout-tubes/src/DREndcapTubes_o1_v01.cpp
```diff
@@ -499,6 +501,21 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
...
   sdet.setPlacement(AssemblyEndcapPV);
 
+  // apply <type_flags> and expose r-z extent for downstream reco
+  dd4hep::xml::setDetectorTypeFlag(e, sdet);
+
+  const double dtheta = thetaB / NbOfEndcap;                    // per-tower theta
+  const double theta_min = thetaB - NbOfEndcapReduced * dtheta; // innermost built tower edge
+  const double zmin = innerR * tan(thetaB);                     // front-face plane
+
+  dd4hep::rec::LayeredCalorimeterData* caloData = new dd4hep::rec::LayeredCalorimeterData;
```
jbeirer, line 512:
```suggestion
// DD4hep DetElement takes ownership and deletes the extension on destruction
dd4hep::rec::LayeredCalorimeterData* caloData = new dd4hep::rec::LayeredCalorimeterData;
```

This ownership transfer is not obvious from the raw new. Could we document that the DetElement extension mechanism takes ownership?

detector/calorimeter/dual-readout-tubes/src/DREndcapTubes_o1_v01.cpp
```diff
@@ -499,6 +501,21 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
...
+  caloData->layoutType = dd4hep::rec::LayeredCalorimeterData::EndcapLayout;
```
jbeirer, line 513: Should the symmetry metadata be populated here? `outer_symmetry` appears to be `NbOfZRot`, since the endcap is assembled from 72 phi wedges. For `inner_symmetry`, should the central opening be represented as the same 72-fold polygon or intentionally approximated as circular with `0`?

FCCee/IDEA/compact/README.md
```diff
@@ -56,3 +56,7 @@ April 2025: Added the dual-readout segmented crystal ECAL
 
 February 2026: Added the neighborhood finding algorithm for the crystal dual-readout calorimeter.
 April 2026: Switch to vertex detector and silicon wrapper from IDEA_o1_v04.
+
+July 2026: Added `LayeredCalorimeterData` extensions (r-z extent and DetType flags) to the dual-readout barrel and endcap tubes calorimeters, so downstream reconstruction can retrieve their inner radius / z directly from the geometry.
```
jbeirer:
```suggestion
July 2026: Added reconstruction metadata to the dual-readout barrel and endcap tubes calorimeters: `LayeredCalorimeterData` extensions provide their r-z extents, while the XML `DetType` flags are applied to their `DetElement`s. This allows downstream reconstruction to retrieve their inner radius / z directly from the geometry.
```

test/compact/IDEA_o2_v01_CI/DectDimensions_IDEA_o2_v01_CI.xml
jbeirer, line 1: Could we avoid maintaining a complete copy of the production dimensions file here? This file appears to differ from `DectDimensions_IDEA_o2_v01.xml` only in the following values:

* `scepcal_phi_load_end`: `127` → `10`
* `scepcal_barrel_theta_load_start`: `0` → `183`
* `scepcal_barrel_theta_load_end`: `489` → `307`
* `DRBTEndCaloPhi`: `360*deg` → `30*deg`
* `DRBTTowerNumberEnd`: `45` → `12`

Perhaps these replacements could be stored in a small JSON or YAML map under a dedicated CI directory and applied by an XML-aware script to a temporary copy in the build directory. The script could require every selector to match exactly once and fail if the production file changes unexpectedly. This would keep the production geometry untouched while avoiding roughly 370 duplicated lines that may drift.

test/compact/IDEA_o2_v01_CI/IDEA_o2_v01_CI.xml
jbeirer, line 1: Could this CI master file also be generated from the production `IDEA_o2_v01.xml` rather than maintained as a separate complete copy? The CI-specific transformations here seem limited to:

* replacing the dimensions include with the generated CI dimensions file;
* replacing the SCEPCal include with the generated CI SCEPCal file;
* removing the `DREndcapTubes_o1_v01.xml` include.

These operations could be recorded alongside the dimension replacements in the same CI transformation specification and applied to a temporary copy during test setup. This would reduce the risk that future changes to the production master file are not reflected in the CI geometry.

test/compact/IDEA_o2_v01_CI/SCEPCal_o1_v01_CI.xml
jbeirer, line 1: Could this file be generated from the production `SCEPCal_o1_v01.xml` using the same CI transformation mechanism rather than maintained as another large copy? The CI-specific changes appear to be:

* setting the main-layer endcap and both timing-layer `construct` attributes to `false`;
* removing the optical-surface definitions;
* removing the unused optical-property matrices;
* removing the PbWO scintillation properties while retaining the properties needed for the CI simulation.

These attribute changes and element removals could be expressed as XML selectors in a small JSON or YAML file and applied to a temporary build-directory copy. The generation script should verify that each expected element is found exactly once, so changes to the production XML fail clearly rather than silently producing an outdated CI geometry.

FCCee/IDEA/compact/IDEA_o2_v01/IDEA_o2_v01_CI.xml
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lccdd xmlns:compact="http://www.lcsim.org/schemas/compact/1.0"
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  xs:noNamespaceSchemaLocation="http://www.lcsim.org/schemas/compact/1.0/compact.xsd">
+
+  <info name="IDEA_o2_v01"
```
jbeirer: Since this is a reduced geometry rather than the complete IDEA_o2_v01 detector, would it be useful to give the <info> block a distinct name/title, for example IDEA_o2_v01_CI? Keeping the production name may make CI logs or generated geometry metadata look as though the full detector was loaded.

example/SteeringFile_IDEA_o2_v01.py
jbeirer, line 1: I am not sure we should add CI-specific logic to this steering file. The authoritative IDEA option-2 steering file is maintained in [HEP-FCC/FCC-config](https://github.com/HEP-FCC/FCC-config/blob/main/FCCee/FullSim/IDEA/IDEA_o2_v01/SteeringFile_IDEA_o2_v01.py), so keeping another full copy in k4geo already creates a risk of the two versions drifting, and this change makes the k4geo copy diverge further.

Could the test instead use the official FCC-config steering file and reconcile it with the reduced geometry at test setup? This is similar to the approach implemented in [`key4hep/k4Bench#110`](https://github.com/key4hep/k4Bench/pull/110): create a run-local copy of the steering file and remove only entries from `SIM.geometry.regexSensitiveDetector` whose corresponding detector is absent from the compact geometry. For this CI geometry, that would remove only `DREndcapTubes` while retaining the `DRBarrelTubes` configuration.

[`HEP-FCC/FCC-config#373` ](https://github.com/HEP-FCC/FCC-config/pull/373) considered exposing a steering-file option for this, but its proposed `--no-simulateDRCalo` switch disabled both tube calorimeters and was closed after k4Bench handled the reconciliation itself. We could either apply the same targeted reconciliation in this test or revisit that PR with a more granular/general mechanism, but I would prefer not to introduce an `IDEA_O2_CI` convention into a duplicated steering file.

test/CMakeLists.txt
jbeirer, line 1: This currently only checks that one barrel event finishes successfully. `add_ddsim_gun_test` does not inspect the output file, so the test could still pass if an expected SCEPCal or DRBT hit collection is missing or empty. Do we want to keep this as a smoke test only, or add a lightweight sanity check that the expected output collections exist and contain hits?

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -205,6 +207,45 @@ static dd4hep::Ref_t create_detector_SCEPCal_MainLayer(dd4hep::Detector& theDete
...
+  extensionData->extent[5] = BARREL_INNER_R + XTAL_LEN_F + XTAL_LEN_R; // endcap rmax
+  extensionData->layers.resize(2);
+  extensionData->layers.at(0).distance = BARREL_INNER_R;              // z-axis to the front layer
+  extensionData->layers.at(1).distance = BARREL_INNER_R + XTAL_LEN_F; // z-axis to the rear layer
+  extensionData->layers.at(0).sensitive_thickness = XTAL_LEN_F;       // thickness of the front layer
+  extensionData->layers.at(1).sensitive_thickness = XTAL_LEN_R;       // thickness of the rear layer
+
+  // TODO separate barrel & endcap
+  // type is barrel for the moment
+  extensionData->layoutType = dd4hep::rec::LayeredCalorimeterData::BarrelLayout;
```
jbeirer, line 255: I see the `TODO` about separating the barrel and endcap. Just to clarify the current intended usage: since this extension is marked as `BarrelLayout` and the layer distances are barrel-specific, should downstream reconstruction currently treat only the barrel layer information as valid and use the extent fields for the endcap?

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -205,6 +207,45 @@ static dd4hep::Ref_t create_detector_SCEPCal_MainLayer(dd4hep::Detector& theDete
...
+  // create DDRec extension to fill dimensions needed downstream
+  auto extensionData = new dd4hep::rec::LayeredCalorimeterData;
+  // rmin, rmax, zmin, zmax, rmin2, rmax2
+  // inner r & z are avg values (for track extrapolation)
+  // outer r & z are the (cylinder's) envelope values
+  extensionData->extent[0] = BARREL_INNER_R;                           // barrel rmin
+  extensionData->extent[1] = BARREL_INNER_R + XTAL_LEN_F + XTAL_LEN_R; // barrel rmax
+  extensionData->extent[2] = BARREL_HALF_Z;                            // endcap zmin
+  extensionData->extent[3] = BARREL_HALF_Z + XTAL_LEN_F + XTAL_LEN_R;  // endcap zmax
+  extensionData->extent[4] = BARREL_HALF_Z * std::tan(thC_ec_beg);     // endcap rmin
```
jbeirer: Is this intended to point to the centre of the first endcap crystal ring rather than the physical inner edge? The geometry uses the lower edge of the ring and applies the projective offset when calculating its minimum radius.

detectorSegmentations/include/detectorSegmentations/SCEPCal_MainSegmentation_k4geo.h
```diff
@@ -159,12 +170,13 @@ namespace DDSegmentation {
...
-    int m_iTheta_barrel_end_ = -1;   // theta index end for barrel
-    int m_nPhi_ = -1;                // number of phi segments
-    int m_nGamma_ = -1;              // number of gamma segments
+    int m_detId_barrel_ = 0;              // system id for barrel
+    int m_detId_endcap_ = 0;              // system id for endcap
+    int m_iTheta_barrel_start_ = -1;      // theta index start for barrel
+    int m_iTheta_barrel_end_ = -1;        // theta index end for barrel
+    int m_nPhi_ = -1;                     // number of phi segments
+    int m_nGamma_ = -1;                   // number of gamma segments (barrel; fallback)
+    std::map<int, int> m_nGammaPerTheta_; // gamma count per global theta field (endcap is projective)
```
jbeirer: In the current implementation, the map is populated only for the endcap, while barrel theta values use the constant `m_nGamma_` fallback. Could the comment be adjusted to reflect that?

detectorSegmentations/src/SCEPCal_MainSegmentation_k4geo.cpp
```diff
@@ -83,10 +85,10 @@ namespace DDSegmentation {
           setCellID(System(cID), Phi(cID), Theta(cID), igamma + 1, Epsilon(cID), Depth(cID), isCherenkov(cID));
       int iphi_low = modulo(iphi - 1, m_nPhi_);
       CellID cID_low =
-          setCellID(System(cID), iphi_low, Theta(cID), m_nGamma_ - 1, Epsilon(cID), Depth(cID), isCherenkov(cID));
+          setCellID(System(cID), iphi_low, Theta(cID), nGammaHere - 1, Epsilon(cID), Depth(cID), isCherenkov(cID));
```
jbeirer, line 88: Adjacent endcap theta rings can have different numbers of gamma cells, but this keeps the same gamma index when moving to the next ring. Could this create a neighbour cell that does not exist? Should the gamma index be adjusted to match the angular position in the neighbouring ring?

detectorSegmentations/src/SCEPCal_MainSegmentation_k4geo.cpp
```diff
@@ -68,8 +68,10 @@ namespace DDSegmentation {
     // get gamma
     int igamma = Gamma(cID);
     int iphi = Phi(cID);
+    // gamma count for THIS cell's theta ring (barrel: constant; endcap: projective/per-ring)
+    const int nGammaHere = nGammaAtTheta(Theta(cID));
```
jbeirer, line 72: The geometry seems to start the negative endcap at m_iTheta_barrel_end_ + 2, while the neighbour logic assumes it starts at m_iTheta_barrel_end_ + 1. Could you check whether this creates a neighbour ID for which no crystal exists?

plugins/DRTubesSDAction.cpp
```diff
@@ -78,6 +80,10 @@ namespace sim {
     m_userData.collection_cher_left = defineCollection<Geant4Calorimeter::Hit>("DRETCherLeft");
     m_userData.collection_drbt_cher = defineCollection<Geant4Calorimeter::Hit>("DRBTCher");
     m_userData.collection_drbt_scin = defineCollection<Geant4Calorimeter::Hit>("DRBTScin");
+    // Combined MC-truth energy-deposit collection for both scintillating and
+    // Cherenkov fibers. The two are kept separable downstream by the cellID
+    // (cherenkov bit: scint=0, Cherenkov=1; system field: endcap=25, barrel=28).
+    m_userData.collection_edep = defineCollection<Geant4Calorimeter::Hit>("DRTubeEdep");
```
jbeirer, line 86: The barrel and endcap use different cell-ID encodings: the [endcap encoding](https://github.com/key4hep/k4geo/blob/ab53f1bf792eb800570dd886d5f98a1839db8fd4/plugins/DRTubesSDAction.cpp#L187) uses `tower:6,air:1,col:16`, while the [barrel encoding](https://github.com/key4hep/k4geo/blob/ab53f1bf792eb800570dd886d5f98a1839db8fd4/plugins/DRTubesSDAction.cpp#L235) uses `tower:-8,air:6,col:-16`.

However, both sensitive-detector instances create a collection named [`DRTubeEdep`](https://github.com/key4hep/k4geo/blob/ab53f1bf792eb800570dd886d5f98a1839db8fd4/plugins/DRTubesSDAction.cpp#L86).

Since an EDM4hep collection stores only one cell-ID encoding in its metadata, won’t either the barrel or endcap IDs be decoded incorrectly?

plugins/SCEPCal_MainSDAction.cpp
```diff
@@ -118,17 +102,44 @@ namespace sim {
...
+      } else // keep it until it passes the reflection
         return true;
-    }
+    } // if optical photon
+
+    if (edep <= 0.)
+      return true; // do not save hits with no energy deposit
+
+    // Do not fill edep and S hits if there is no ionizing energy deposit
+    auto Scount = m_userData.SmearSsignal(edep);
```
jbeirer, line 127: `SmearSsignal(edep)` can randomly return zero even when the step has a real energy deposit. Since this returns before creating the `SCEPCal_MainEdep` hit, doesn’t that make the truth-energy collection depend on the random scintillation response and cause some real deposits to be lost?

Could the truth hit and contribution be recorded before this check, with `Scount <= 0` skipping only the scintillation-count hit?

jbeirer (changes requested):

detector/calorimeter/dual-readout-tubes/src/DRBarrelTubes_o1_v01.cpp
SanghyunKo: Added the comment

detector/calorimeter/dual-readout-tubes/src/DREndcapTubes_o1_v01.cpp
SanghyunKo: Added the comment (same as above)

detector/calorimeter/dual-readout-tubes/src/DREndcapTubes_o1_v01.cpp
SanghyunKo: I don't think there is any use-case of this, but this is free so I just added it

FCCee/IDEA/compact/README.md
SanghyunKo: Updated as suggested

test/compact/IDEA_o2_v01_CI/DectDimensions_IDEA_o2_v01_CI.xml
SanghyunKo: I think this suggestion is a completely different mechanism for handling XML files (the same logic applies across the whole k4geo repo, not just this PR), which is outside the scope of this PR...

test/compact/IDEA_o2_v01_CI/IDEA_o2_v01_CI.xml
SanghyunKo: I prefer to have the partial geometry XML file in the repo rather than generating it on the fly during the CI test. Geometry update is rare and I think duplicating a file is affordable overhead.

(I found this useful for fast debugging and/or instructing other users/students without waiting for the full geometry construction)

test/compact/IDEA_o2_v01_CI/SCEPCal_o1_v01_CI.xml
SanghyunKo: Same as above

FCCee/IDEA/compact/IDEA_o2_v01/IDEA_o2_v01_CI.xml
SanghyunKo: Fixed to `IDEA_o2_v01_CI`

example/SteeringFile_IDEA_o2_v01.py
SanghyunKo: I somehow share the point raised here; My take is to strip ALL XML and steering files from k4geo and put everything in the config repo. But again, I think this is beyond the scope of this PR. We should first discuss this in the key4hep meeting and decide whether we want to do so.

test/CMakeLists.txt
SanghyunKo: I think @s6anloes and his summer student are working on this "lightweight" test as an independent effort.

detector/calorimeter/SCEPCal_MainLayer.cpp
SanghyunKo: Yes, the ECAL barrel and endcap are stitched into a single geometry; so this is the intended use case.

detector/calorimeter/SCEPCal_MainLayer.cpp
SanghyunKo: This is intended; I don't think there is a use case for the inner ring AFAIK.

detectorSegmentations/include/detectorSegmentations/SCEPCal_MainSegmentation_k4geo.h
SanghyunKo: Expanded the comment as requested

detectorSegmentations/src/SCEPCal_MainSegmentation_k4geo.cpp
SanghyunKo: The `neighbour` function is designed as a bitwise operator rather than a lookup for existing physical cells. This is an intended trade-off for computing performance. The function can return a void cellID, but this is not harmful because no calo hit will be matched to the given cellID downstream.

detectorSegmentations/src/SCEPCal_MainSegmentation_k4geo.cpp
SanghyunKo: Same as above

plugins/DRTubesSDAction.cpp
SanghyunKo: I think this is a very good point and a real issue. My take is simply to synchronize the cellID encoding scheme between the HCAL barrel and endcap and consolidate the collections. AFAIK, the HCAL barrel/endcap cellID encoding schemes just "happen to be" different from the historical reason (@s6anloes may confirm this).

We already produce 10 different sim calo hit collections (3 ECAL + 7 HCAL collections; ECAL scint, ECAL cheren, ECAL truth, HCAL barrel scint, HCAL barrel cheren, HCAL endcap left scint, HCAL endcap left cheren, HCAL endcap right scint, HCAL endcap right cheren, HCAL truth); this is already very annoying downstream, and I really don't want to emit another calo hit collection.

However, in practice, I have never experienced an issue with the different cellID encoding scheme in this collection because it is accessed only via a link, not directly. The sole purpose of this collection is to establish MC particle links via calo-hit contributions. Physically meaningful sim calo hits live in different collections, where the cellID metadata is intact.

plugins/SCEPCal_MainSDAction.cpp
SanghyunKo: Indeed, that was my first shot, but the file size exploded (~O(10 MB/evt), since it records every tiny energy deposit), and I had to give up storing truth hits that do not generate the physical calo hit anyway.

test/compact/IDEA_o2_v01_CI/DectDimensions_IDEA_o2_v01_CI.xml
jbeirer: My concern was mainly that these _CI files are reduced copies of the production IDEA o2 geometry, rather than separate detector versions, so they could easily drift when the production files are updated.

I do not think this needs to block the PR, but if more reduced CI geometries are added later, it may be worth revisiting whether generating them would be cleaner.

test/compact/IDEA_o2_v01_CI/IDEA_o2_v01_CI.xml
jbeirer: In that case, could we at least group the three derived files separately rather than mixing them with the production IDEA o2 files?

Perhaps test/compact/IDEA_o2_v01_CI/ would fit, since k4geo already keeps test-specific compact geometries under test/compact.

test/compact/IDEA_o2_v01_CI/SCEPCal_o1_v01_CI.xml
jbeirer: Same point here, I am fine with keeping the reduced files checked into the repository, especially if they are useful for local debugging.

In that case, I think the three CI-specific files should at least live together under something like `test/compact/IDEA_o2_v01_CI/`, rather than next to the production geometry components. That would make it much clearer that they form one derived test geometry and need to be kept in sync.

example/SteeringFile_IDEA_o2_v01.py
jbeirer: As far as I can see, this repository only contains the two IDEA steering files under `example/`. Are these currently used for anything other than the tests here?

This is separate from the broader discussion about where the XML files should live, I am not suggesting moving XML into or out of k4geo here. My point is only about avoiding duplicated steering configuration.

If these example steering files are not otherwise used, I think it would be cleaner to remove them and let the test use a small test-specific steering file containing only the configuration it actually needs.

test/CMakeLists.txt
jbeirer: Sounds good

detector/calorimeter/SCEPCal_MainLayer.cpp
jbeirer: Just to clarify the intended use: does the downstream reconstruction specifically need the centre of the first endcap ring?

`extent[4]` would normally be understood as the endcap’s inner radial boundary, so other users could easily misinterpret what is stored here. If the first-ring centre is required instead, would it be better to provide it separately or at least document this convention here?

detectorSegmentations/src/SCEPCal_MainSegmentation_k4geo.cpp
jbeirer: What happens when the same gamma index exists in the neighbouring ring but represents a different angular position because that ring has a different `nGamma`? Could this return a real but geometrically wrong neighbour?

detectorSegmentations/src/SCEPCal_MainSegmentation_k4geo.cpp
jbeirer: On the negative side, isn’t the first actual endcap ring at `m_iTheta_barrel_end_ + 2`, while this connects the barrel to `+1`?

If so, this would not only return a void ID but also omit the real barrel–endcap neighbour. Is the gap intentional?

plugins/DRTubesSDAction.cpp
jbeirer: Agreed, synchronising the barrel and endcap encodings sounds cleaner than adding another collection.

plugins/SCEPCal_MainSDAction.cpp
jbeirer: Ok, is `SCEPCal_MainEdep` intended to be a complete energy-deposit collection, or only to provide truth links for deposits that produced a scintillation signal? If it is the latter, could this behaviour be documented explicitly?

plugins/DRTubesSDAction.cpp
s6anloes: Yes, it's something I can tackle if people call for it!

detector/calorimeter/dual-readout-tubes/src/DREndcapTubes_o1_v01.cpp
```diff
@@ -499,6 +501,24 @@ static Ref_t create_detector(Detector& description, xml_h e, SensitiveDetector s
...
+  const double theta_min = thetaB - NbOfEndcapReduced * dtheta; // innermost built tower edge
+  const double zmin = innerR * tan(thetaB);                     // front-face plane
+
+  // DD4hep DetElement takes ownership and deletes the extension on destruction
+  dd4hep::rec::LayeredCalorimeterData* caloData = new dd4hep::rec::LayeredCalorimeterData;
+  caloData->layoutType = dd4hep::rec::LayeredCalorimeterData::EndcapLayout;
+  caloData->inner_symmetry = NbOfZRot;
+  caloData->outer_symmetry = NbOfZRot;
+  caloData->extent[0] = zmin * tan(theta_min); // rmin (innermost instrumented tower)
+  caloData->extent[1] = innerR + tower_height; // rmax
```
s6anloes:
```suggestion
  caloData->extent[1] = innerR + tower_height*sin(thetaB); // rmax
```
Tower is inclined, so maximal radius is less than `innerR+tower_height`

test/compact/IDEA_o2_v01_CI/DectDimensions_IDEA_o2_v01_CI.xml
SanghyunKo: I don't foresee these `_CI` files being persistent. It's there simply because we cannot afford CI with the full geometry for the moment. Given that the detector geometry version up does not happen often, I think this is fine as a temporary stopgap. Once we improve the geometry construction to a manageable level, we should use full geometry for the CI and strip this off.

test/compact/IDEA_o2_v01_CI/IDEA_o2_v01_CI.xml
SanghyunKo: Okay, i's now moved to the suggested path.

test/compact/IDEA_o2_v01_CI/SCEPCal_o1_v01_CI.xml
SanghyunKo: Moved

example/SteeringFile_IDEA_o2_v01.py
SanghyunKo: Ideally yes, the config should be stored and managed uniquely, but there comes a chicken-and-egg problem. Suppose we replace CIs by checking out steering files from FCC-config. Imagine someone updates the k4geo and the corresponding ddsim steering file in FCC-config (as in this case). Then we're forced to merge the PR in either of the two repos with broken CI.

This can be fixed by adding functionality to run the CI by pulling another PR's branch in the different repo, but until then we need to endure the overhead of two config files, which I consider not too expensive.

detector/calorimeter/SCEPCal_MainLayer.cpp
SanghyunKo: I think no one should really use this if one wants a precise geometry description (`LayeredCalorimeterData` is not really suitable for describing the projective geometry anyway), but I updated the line.

detectorSegmentations/src/SCEPCal_MainSegmentation_k4geo.cpp
SanghyunKo: When `nGamma` changes, the definition of neighbor becomes not trivial. We can try the same `gamma` index as before, or giving the cellID of all touching/adjacent cells. I updated the algorithm following the latter.

detectorSegmentations/src/SCEPCal_MainSegmentation_k4geo.cpp
SanghyunKo: Okay, I got what you mean. In fact, this is not an artifact of the neighbor-finding algorithm, rather rooted in the original geometry construction code. And indeed, "jumping" over the 1 theta index when crossing the barrel-endcap border seems a bit weird. I shifted the `iTheta` of the negative endcap modules by -1 to make `iTheta` continuous.

plugins/SCEPCal_MainSDAction.cpp
SanghyunKo: Added comments

detector/calorimeter/dual-readout-tubes/src/DREndcapTubes_o1_v01.cpp
SanghyunKo: Fixed

plugins/DRTubesSDAction.cpp
jbeirer: Since DRTubeEdep is introduced in this PR, could we either unify the encodings here or link a follow-up issue that will address it?

jbeirer:
Hi @SanghyunKo,

thanks for addressing all the comments. I am fine with leaving the `DRTubeEdep` encoding issue out of this PR, but it would be good to open a follow-up issue and link it here so it does not get lost.

Unless anyone has further concerns, I am happy to merge this at the end of the day.

plugins/DRTubesSDAction.cpp
s6anloes: A follow-up issue is the way to do I'd say.
I'm not sure changing the encodings is something we'd want to do just like that.
It will have an impact on any user analysis scripts which may or may not exist at the moment.
I would consider making this change with a new `IDEA_o2_v02`, along with a few other changes that I expect will be coming soon, like a change in the encoding of the drift chamber

jbeirer: @s6anloes Sounds good, feel free to open one and link it here. Thanks!

jbeirer (approved):

example/SteeringFile_IDEA_o2_v01.py
```diff
@@ -113,13 +116,18 @@
...
-SIM.action.mapActions["DREndcapTubes"] = "DRTubesSDAction"
-## Configure the regexSD for DREndcapTubes subdetector
-SIM.geometry.regexSensitiveDetector["DREndcapTubes"] = {
-    "Match": ["DRETS"],
-    "OutputLevel": 4,
-}
+## DREndcapTubes only exists in the full geometry; the CI wedge drops it, so skip its SDAction/regexSD
+## here -- otherwise they reference a missing detector and ddsim exits.  Triggered by IDEA_O2_CI=1 or,
+## as a fail-safe, by using a "*_CI.xml" compact on the command line.
+_idea_o2_ci = bool(os.environ.get("IDEA_O2_CI")) or any("_CI.xml" in a for a in sys.argv)
```
jmcarcell: I think an argument would be better than an environment variable; they tend to be a bit harder to track. And it makes it easier to copy and paste a command to run it.

detector/calorimeter/SCEPCal_MainLayer.cpp
```diff
@@ -205,6 +207,50 @@ static dd4hep::Ref_t create_detector_SCEPCal_MainLayer(dd4hep::Detector& theDete
...
+  // outer r & z are the (cylinder's) envelope values
+  extensionData->extent[0] = BARREL_INNER_R;                           // barrel rmin
+  extensionData->extent[1] = BARREL_INNER_R + XTAL_LEN_F + XTAL_LEN_R; // barrel rmax
+  extensionData->extent[2] = BARREL_HALF_Z;                            // endcap zmin
+  extensionData->extent[3] = BARREL_HALF_Z + XTAL_LEN_F + XTAL_LEN_R;  // endcap zmax
+  // endcap rmin: inner edge of the first constructed ring (thC_ec_beg is its centre), converted
+  // from the projective frame to a physical radius (r = z*tan(theta) - PROJ_OFFSET_R), the same
+  // convention used for the phi-slice envelope and the beampipe-clearance test above
+  extensionData->extent[4] = BARREL_HALF_Z * std::tan(thC_ec_beg - D_THETA_ENDCAP / 2) - PROJ_OFFSET_R;
+  extensionData->extent[5] = BARREL_INNER_R + XTAL_LEN_F + XTAL_LEN_R; // endcap rmax
```
s6anloes:
```suggestion
  extensionData->extent[5] = BARREL_INNER_R + (XTAL_LEN_F + XTAL_LEN_R)*sin(whatever_angle_is_correct); // endcap rmax
```
Same comment I gave in the fibre endcap: since the towers are tilted, the max extent will be less than just adding the lengths.
Not sure which one is the correct variable to use thoug (angle should be 45*deg), and it's minor so it's fine if you ignore this comment.

detectorSegmentations/include/detectorSegmentations/SCEPCal_MainSegmentation_k4geo.h
```diff
@@ -137,6 +138,41 @@ namespace DDSegmentation {
...
+      const int nTo = nGammaAtTheta(thetaTo);
+
+      if (nFrom == nTo)
+        return {gamma};
+
+      // the cell covers [gamma, gamma + 1] / nFrom of the phi slice; collect every cell of the
+      // target ring that this interval overlaps (cells merely touching at a boundary are excluded)
+      const int first = (gamma * nTo) / nFrom;                      // floor
+      const int last = ((gamma + 1) * nTo + nFrom - 1) / nFrom - 1; // ceil - 1
+
```
s6anloes: The intent behind the calculations is not clear to me, and I'm not sure they are 100% correct all the time.
I calculated one example and found that the returned gamma values were incorrect.
A more human-readable logic will be easier to validate

example/SteeringFile_IDEA_o2_v01.py
SanghyunKo: Removed the env variable part

detector/calorimeter/SCEPCal_MainLayer.cpp
SanghyunKo: Done

detectorSegmentations/include/detectorSegmentations/SCEPCal_MainSegmentation_k4geo.h
SanghyunKo: Fixed

test/CMakeLists.txt
```diff
@@ -114,12 +114,18 @@ SET_TESTS_PROPERTIES( ${test_name} PROPERTIES TIMEOUT 900)
...
+# The full IDEA_o2_v01 geometry needs ~18 GB and cannot run on CI, so we load a barrel phi/theta
+# wedge (IDEA_o2_v01_CI.xml) and fire a pi- into it to exercise the SCEPCal + DRBT sensitive
+# detectors.  The steering skips the DREndcapTubes SD when the compact file is a "*_CI.xml".
+if(DCH_INFO_H_EXIST)
+add_ddsim_gun_test(${CMAKE_CURRENT_SOURCE_DIR}/compact/IDEA_o2_v01_CI/IDEA_o2_v01_CI.xml
+  N_EVENTS 1
+  EXTRA_ARGS --steeringFile=${CMAKE_CURRENT_SOURCE_DIR}/../example/SteeringFile_IDEA_o2_v01.py
+             --gun.particle pi- --gun.energy 10*GeV --gun.direction 0.966,0.259,0.01
+             --random.seed 1988301045
+  PROPERTIES TIMEOUT 900 RUN_SERIAL TRUE)
```
jmcarcell, line 127: Does it not run without `RUN_SERIAL`? Otherwise it adds directly to the time it takes to run the tests.

detectorSegmentations/include/detectorSegmentations/SCEPCal_MainSegmentation_k4geo.h
```diff
@@ -160,10 +160,14 @@ namespace DDSegmentation {
...
-      const int first = (gamma * nTo) / nFrom;                      // floor
-      const int last = ((gamma + 1) * nTo + nFrom - 1) / nFrom - 1; // ceil - 1
+      // Express the cell's azimuthal span in units of target-ring cells, then take every target
+      // cell it covers.  The -1 on the ceiling drops a cell that the span only touches at its edge,
+      // and keeps the last cell of the ring in range (highEdge is exactly nTo there).
+      const double lowEdge = static_cast<double>(gamma) * nTo / nFrom;
+      const double highEdge = static_cast<double>(gamma + 1) * nTo / nFrom;
+
+      const int first = static_cast<int>(std::floor(lowEdge));
+      const int last = static_cast<int>(std::ceil(highEdge)) - 1;
```
s6anloes, line 170: Thanks, that's much better and it seems to work fine, assuming that $\gamma$ is zero-indexed.
The only concern that remains is maybe floating point precision when applying the `ceil` function.
Say you're interested in the last crystal in the 'from' row. That crystal will have $\gamma=nFrom-1$.
In the `highEdge` calculation this should cancel then perfectly with the `nFrom` in the denominator, leaving `highEdge = nTo`, which should be perfect integer making the `ceil` call safe. 
However, since you're casting to double, there could be unexpected behaviour in this scenario.

Maybe we can round before the `ceil` call with some tolerance. Generally speaking I don't expect any of the calculations to 'coincidentally' be very close to an integer, so the tolerance can be generous.

test/CMakeLists.txt
SanghyunKo: Yes, I intentionally added 'RUN_SERIAL' (you can see a commit dedicated to this in the commit history); otherwise, the CI competes with other workflows over the memory and often killed due to the OOM. We should see whether this is still the case after #578 though.

detectorSegmentations/include/detectorSegmentations/SCEPCal_MainSegmentation_k4geo.h
SanghyunKo: I'm not sure I understand your concern:
> However, since you're casting to double, there could be unexpected behaviour in this scenario.
> Maybe we can round before the ceil call with some tolerance. Generally speaking I don't expect any of the calculations to 'coincidentally' be very close to an integer, so the tolerance can be generous.

The `std::floor` and `std::ceil` ARE the ones that guarantee safe casting to integer. There is no unexpected behavior here (check the behavior of `std::floor` and `std::ceil` under IEEE 754 if you want). In fact, the tolerance will be the one that breaks the logic and creates the artifact that you concern. Suppose the following: 
```
 nFrom=11 nTo=10 gamma=9:  span [8.1818, 9.0909]
     no tolerance  -> cells [8, 9]   correct
     tolerance 0.1 -> cells [8]      loses the real overlap with cell 9
```
Of course 0.1 is a pretty big number here, but you can always find some `nFrom` and `nTo` that breaks the logic for an arbitrary tolerance.

s6anloes:
No more comments, from my side. 
Only @jmcarcell left to resolve and then we can merge @jbeirer ?

example/SteeringFile_IDEA_o2_v01.py
jmcarcell: Hm, okay, I'm not sure if it's impossible to use a parser since we do that in k4FWCore where it is possible to add arguments from steering files.

test/CMakeLists.txt
jmcarcell: This is approx +100 s of testing. It's also possible to coordinate better resource usage between tests so that maybe it can run in parallel with other tests that don't need a lot of memory but that's probably outside of the scope of this PR.

test/compact/IDEA_o2_v01_CI/SCEPCal_o1_v01_CI.xml
```diff
@@ -0,0 +1,314 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lccdd>
+
+  <properties>
```
jmcarcell: Maybe this could be exctracted from these and the non `_CI` file not to be copied in several places?

SanghyunKo: Now the shared `<properties>` block is isolated into a separated file `SCEPCal_properties_o1_v01.xml‎`.

jbeirer:
Hi @jmcarcell,

are you ok with the changes now? If yes, I would propose to merge today.

jbeirer:
Merging this so we can proceed with https://github.com/key4hep/k4geo/pull/578

jbeirer:
Hi @SanghyunKo, 

I guess these performance gains are expected due to hit creation now sitting behind the detection/threshold checks in `SCEPCal_MainSDAction` ([L105-L111](https://github.com/key4hep/k4geo/blob/6ea15a270e79a3e336e8db8f5584f8fdcf555f5c/plugins/SCEPCal_MainSDAction.cpp#L105-L111), and the early returns at [L119-L129](https://github.com/key4hep/k4geo/blob/6ea15a270e79a3e336e8db8f5584f8fdcf555f5c/plugins/SCEPCal_MainSDAction.cpp#L119-L129)), our nightlies see IDEA_o2_v01 drop ~30%, all of it in `SCEPCal_MainLayer` (4.55 → 3.04 s/event)?

SanghyunKo:
Indeed, previously it was returning calo hits with zero photon count, and it seems that the removing them (and their lookup operation) has this nice side effect.

SanghyunKo:
I realized that moving the CI `xml` files into the test folder has caused another side effect (see https://github.com/key4hep/k4geo/pull/611#discussion_r3691235385). Now I cannot access them on the stack because k4geo does not install the test folder. Hence, the original idea of using partial geometry for the downstream tests (e.g. in HEP-FCC/k4RecCalorimeter#271) is now effectively nullified.

I need either of the followings: 1) put CI `xml` files back to the `FCCee/IDEA/compact` subfolder, or 2) install `test/compact` folder so that we can access those `xml` files in the stack. Say me the word and I will create the follow-up PR.

jbeirer:
> I realized that moving the CI `xml` files into the test folder has caused another side effect (see [#611 (comment)](https://github.com/key4hep/k4geo/pull/611#discussion_r3691235385)). Now I cannot access them on the stack because k4geo does not install the test folder. Hence, the original idea of using partial geometry for the downstream tests (e.g. in [HEP-FCC/k4RecCalorimeter#271](https://github.com/HEP-FCC/k4RecCalorimeter/pull/271)) is now effectively nullified.
> 
> I need either of the followings: 1) put CI `xml` files back to the `FCCee/IDEA/compact` subfolder, or 2) install `test/compact` folder so that we can access those `xml` files in the stack. Say me the word and I will create the follow-up PR.

I would personally be in favour of going with 2, but let’s see if others have other preferences

tmadlener:
TL;DR: I think shipping the files is fine, but I would try to come up with a folder structure that doesn't include `test`.

I, personally, don't like shipping things that are mainly (only) there for testing. If this is an ongoing development still in flux, what speaks against installing them along the "normal" compact files (even if they are generated on the fly)? I guess the main questions are:
- Who (how many) need these files to be available from the nightlies?
- What is the expected frequency for changes for these files?
- What is the long term goal here?

jmcarcell:
Also most packages are built with `BUILD_TESTING=OFF`, which skips the test folder

jbeirer:
Maybe the main question is whether we need to install these geometries at all.

For k4RecCalorimeter#271 the test currently picks it up from $K4GEO, but I guess that’s not a hard requirement.

Do we expect other downstream packages to use the reduced geometry as well? If yes, I’d be in favour of installing it somewhere like

`FCCee/IDEA/compact/reduced/IDEA_o2_v01/`

If this is only needed for that one downstream test, I’m not sure it’s worth making it part of the installed geometry.

SanghyunKo:
Currently we'd need the reduced geometry if we want to test the IDEA o2 workflow in downstream - k4RecCalorimeter, k4GaudiPandora, and FCC-config (and possibly more, i.e. k4RecTracker). Otherwise it would certainly hit the OOM issue. Even if #578 manages to reduce the memory successfully, I doubt that we can use the full geometry in CI workflows out of the box because it still takes 20mins for the geometry initialization.

If we acknowledge that we need the reduced geometry, I think the consensus is to put the reduced geometry `xml` files back to where the other `xml` files are located (with some substructure). If there is no imminent confrontation, I'll create the PR in such a way.

PR #615, opened 2026-07-13
IDEA: Adding non-uniform muon-system builder and renaming muon system files

mahmoudali2:
BEGINRELEASENOTES
- Adding a new version of muon system builder (sequence-based layer configuration), where we can build a non-uniform detector/yoke sequence of layers.
- Renaming the builders to be independent of uRWELL(which is the case), and only xml is uRWELL dependent.
- Adding another version of the muon system xml for the 4L-50 cm configuration, and its optional, so user can choose either the baseline or this optimised version.

ENDRELEASENOTES

### DIFF
--- a/FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml
+++ b/FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml
@@ -61,14 +61,14 @@
   <include ref="EndPlateAbsorber_o1_v01.xml"/>
 
   <!-- Import pre-shower -->
-  <include ref="Preshower_o1_v01.xml"/>
+  <include ref="PreshowerMuRWELL_o1_v01.xml"/>
 
   <!-- Import fiber-based dual-readout calorimeter -->
   <!-- (uncomment the following line to effectively include it) -->
   <include ref="FiberDualReadoutCalo_o1_v01.xml"/>
 
   <!-- Import muon system -->
-  <include ref="MuonSystem_o1_v01.xml"/>
+  <include ref="MuonSystemMuRWELL_o1_v01.xml"/>
 
   <plugins>
     <plugin name="DD4hepVolumeManager"/>
--- a/FCCee/IDEA/compact/IDEA_o1_v04/IDEA_o1_v04.xml
+++ b/FCCee/IDEA/compact/IDEA_o1_v04/IDEA_o1_v04.xml
@@ -61,14 +61,14 @@
   <include ref="../IDEA_o1_v03/EndPlateAbsorber_o1_v01.xml"/>
 
   <!-- Import pre-shower -->
-  <include ref="../IDEA_o1_v03/Preshower_o1_v01.xml"/>
+  <include ref="../IDEA_o1_v03/PreshowerMuRWELL_o1_v01.xml"/>
 
   <!-- Import fiber-based dual-readout calorimeter -->
   <!-- (uncomment the following line to effectively include it) -->
   <include ref="../IDEA_o1_v03/FiberDualReadoutCalo_o1_v01.xml"/>
 
   <!-- Import muon system -->
-  <include ref="../IDEA_o1_v03/MuonSystem_o1_v01.xml"/>
+  <include ref="../IDEA_o1_v03/MuonSystemMuRWELL_o1_v01.xml"/>
 
   <plugins>
     <plugin name="DD4hepVolumeManager"/>
--- a/FCCee/IDEA/compact/IDEA_o1_v04/MuonSystemMuRWELL_o1_v02.xml
+++ b/FCCee/IDEA/compact/IDEA_o1_v04/MuonSystemMuRWELL_o1_v02.xml
@@ -1,0 +1,85 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lccdd>
+
+  <info name="MuonSystem"
+	title="IDEA muon system: 4-layer layout with 500 mm iron yoke for FCC-ee"
+	author="Mahmoud Althakeel, mahmoud.althakeel@cern.ch"
+	url="no"
+	status="development"
+	version="o1_v02">
+    <comment>
+      Version o1_v02: 4 active layers (3 yokes), 500 mm iron yoke thickness.
+      Increases sensitive decay-length coverage to ~6 m relative to o1_v01
+      (3 layers, 300 mm yoke).  Factory: muonSystem_o1_v01.
+    </comment>
+  </info>
+
+  <define>
+     <!---                            Muon System Parameters                       -->
+     <!-- %%%%%%                      microRWELL chamber different layers thicknesses                %%%%%% -->
+       <constant name = "G10_FR4Thick"                          value = "1.6*mm"/>
+       <constant name = "CuThick"                               value = "0.035*mm"/>
+       <constant name = "GasLayerThick"                         value = "6*mm"/>
+       <constant name = "Cu2Thick"                              value = "0.005*mm"/>
+       <constant name = "KaptonThick"                           value = "0.05*mm"/>
+       <constant name = "CarbonFiberThick"                      value = "0.0001*mm"/>
+       <constant name = "CarbonFiber2Thick"                     value = "0.1*mm"/>
+       <constant name = "SiThick"                               value = "1.6*mm"/>
+       <constant name = "mRWELLTotalThickness"                  value = "G10_FR4Thick+CuThick+GasLayerThick+Cu2Thick+KaptonThick+CarbonFiberThick+CuThick+CarbonFiber2Thick+CuThick+SiThick"/> <!-- This sequense is the current order of mRWELL (total 10 slices) -->
+
+       <constant name = "mRWELLYLength"                         value = "250*mm"/>
+       <constant name = "mRWELLZLength"                         value = "250*mm"/>
+
+       <constant name = "overlapingY"                           value = "5*mm"/> <!-- the common distance between mRWELL chambers in Y direction -->
+       <constant name = "overlapingZ"                           value = "5*mm"/> <!-- the common distance between mRWELL chambers in Z direction -->
+       <constant name = "clearance"                             value = "1*mm"/> <!-- it's a small distance to be used to avoid overlapping between the different volumes ~ 1 mm -->
+     <!-- %%%%%%                      Iron yoke thickness (500 mm)          %%%%%% -->
+      <constant name = "YokeThickness"                          value = "500*mm"/>
+      <!-- %%%%%%                      Readout description                  %%%%%% -->
+      <constant name = "MuonSystemReadoutID"            type="string"  value = "system:5,type:-2,layer:2,chamber:13,slice:1,y:-10,z:-10"/>  <!-- The bit field is divided into 2^5 systems(IDEA sub-detectors), 2^2 types(0 = Barrel Muon System, +1 = positive endcap, and -1 = negative endcap"), 2^2 layers(number of layers in barrel for example) ,2^13 chambers(the number of muRWELL chambers in every layer), 2^1 slice(number of sensitive layers inside every chambers), and 2^10 y&z strips in every sensitive layer-->
+  </define>
+
+  <readouts>
+    <readout name="MuonSystemCollection">
+     <segmentation type="CartesianGridYZ" grid_size_y="1.2*mm" grid_size_z="1.2*mm"/>   <!-- Depending on strip pitch 1.4 mm  -->
+     <id>${MuonSystemReadoutID}</id>
+  </readout>
+  </readouts>
+
+  <detectors>
+    <!-- mRWELL envelope -->
+    <detector name="Muon-System" type="muonSystem_o1_v01" id="DetID_muonSystem" readout="MuonSystemCollection">
+      <dimensions x="mRWELLTotalThickness/2.0" y="mRWELLYLength" z="mRWELLZLength" x_offset="0*mm" y_offset="0*mm" z_offset="0*mm" material="Air"/>
+      <sensitive type="tracker"/>
+
+      <!-- Specify the detector parameters and the overlap /// if you want exclude any component, e.g: endcap, just put endcapDetectorLayerss=0 in the Endcap Entry  // radius is put in the middle, so its not the inner neither the outer -->
+      <generalParameters numSides="numberOfSides" overlapY="overlapingY" overlapZ="overlapingZ" clearance="clearance"/>
+      <Barrel numDetectorLayers ="4"  rmin="BarrelFirstLayerRadius" length="BarrelLength" numYokes="3" yoke_Thickness="YokeThickness" yoke_Material="G4_Fe"/>
+      <Endcap numDetectorLayers="4"  rmin="EndcapLayersInnerRadius"  z_offset="EndcapFirstLayerZOffset" numYokes="3" yoke_Thickness="YokeThickness" yoke_Material="G4_Fe" />
+
+      <!-- mRWELL chamber -->
+      <!-- note: all thicknesses are half-lengths -->
+      <slice x="G10_FR4Thick/2.0" material="G10_FR4" vis="G10_FR4_vis" />
+      <slice x="CuThick/2.0" material="G4_Cu" vis="Cu_vis" />
+      <slice x="GasLayerThick/2.0" material="ARCO2CF4" sensitive="true" vis="Sensitive_vis" />
+      <slice x="Cu2Thick/2.0" material="G4_Cu" vis="Cu_vis" />
+      <slice x="KaptonThick/2.0" material="Kapton" vis="Kapton_vis" />
+      <slice x="CarbonFiberThick/2.0" material="CarbonFiber" vis="CarbonFiber_vis" />
+      <slice x="CuThick/2.0" material="G4_Cu" vis="Cu_vis" />
+      <slice x="CarbonFiber2Thick/2.0" material="CarbonFiber" vis="CarbonFiber_vis" />
+      <slice x="CuThick/2.0" material="G4_Cu" vis="Cu_vis" />
+      <slice x="SiThick/2.0" material="G4_Si" vis="Si_vis" />
+    </detector>
+  </detectors>
+
+  <plugins>
+    <plugin name="DD4hep_GenericSurfaceInstallerPlugin"> <!-- Plugin needed for the surfaces-->
+      <argument value="Muon-System" />
+      <argument value="dimension=2" />
+      <argument value="u_y=1." />
+      <argument value="v_z=1." />
+      <argument value="n_x=1." />
+    </plugin>
+  </plugins>
+
+</lccdd>
--- a/FCCee/IDEA/compact/IDEA_o2_v01/IDEA_o2_v01.xml
+++ b/FCCee/IDEA/compact/IDEA_o2_v01/IDEA_o2_v01.xml
@@ -69,7 +69,7 @@
   <include ref="DRBarrelTubes_o1_v01.xml"/>
 
   <!-- Import muon system -->
-  <include ref="../IDEA_o1_v03/MuonSystem_o1_v01.xml"/>
+  <include ref="../IDEA_o1_v03/MuonSystemMuRWELL_o1_v01.xml"/>
 
   <plugins>
     <plugin name="DD4hepVolumeManager"/>
--- a/detector/muonSystem/README.md
+++ b/detector/muonSystem/README.md
@@ -2,9 +2,12 @@
 
 The IDEA detector concept includes a muon detection system and pre-shower designed using µRWELL technology. Each station consists of a large mosaic of 50 × 50 cm² µRWELL detectors. For more technical details about the new detector technology, see the [µRWELL](https://iopscience.iop.org/article/10.1088/1748-0221/10/02/P02008).
 
-## muonSystemMuRWELL_o1_v01.cpp
+## muonSystem_o1_v01.cpp
 The first version of the detailed muon system driver, it can be used to describe both barrel and endcap, if you want to eleminate one of them, just set the number of layers= 0.
 The code has been designed to be very flexible, where the user can choose the number of sides in the R-Phi plane, `numSides` (hexagon, octagon, etc), and the detector builder will automatically calculate the number and places of the copied chambers. Some of the code advantages: 
  * If the side length do not fit with an integer number of 50 × 50 cm² , the builder will make a chamber with unusual dimensions, which can fit the excess area at the end of the side.
  * The availability to make multiple layers with different inner radius and barrel length.
  * The code is very general, it can be used to describe any detector system made from repeated tiles (e.g. pre-shower) and has the capability to fill the gaps with unusual dimensions tiles.
+
+## muonSystem_o1_v02.cpp
+The second version of the muon system driver. Instead of fixed `numDetectorLayers`/`numYokes` counts, `<Barrel>` and `<Endcap>` take an ordered sequence of `<layer type="detector"/>` and `<layer type="radiator" thickness="..." material="..."/>` child elements, allowing arbitrary detector/radiator arrangements (e.g. two detector layers, then a radiator, then a detector, then a radiator). The endcap sequence is mirrored for the negative side so that detector layer 0 is always closest to the IP in both endcaps.
--- a/detector/muonSystem/muonSystem_o1_v01.cpp
+++ b/detector/muonSystem/muonSystem_o1_v01.cpp
@@ -4,7 +4,7 @@ mahmoud.ali@cern.ch
 
 Factory for IDEA muon system
 Expected xml structure (the 'sensitive' keyword is optional and defaults to false):
-<detector type="muonSystemMuRWELL_o1_v01" ...>
+<detector type="muonSystem_o1_v01" ...>
   <dimensions x="..." y="..." z="..." z_offset="..." x_offset="..." y_offset="...">  <!--  dimension of the local
 chamber envelope. x: the half length of the thickness of the chamber, y&z: the half length of the 2D plane dimensions of
 the chambers--> <sensitive type="tracker"/>
@@ -41,7 +41,7 @@ using dd4hep::rec::VolCylinder;
 using dd4hep::rec::VolPlane;
 using dd4hep::rec::volSurfaceList;
 
-static dd4hep::Ref_t createmuonSystemMuRWELL_o1_v01(dd4hep::Detector& lcdd, dd4hep::xml::Handle_t xmlElement,
+static dd4hep::Ref_t createmuonSystem_o1_v01(dd4hep::Detector& lcdd, dd4hep::xml::Handle_t xmlElement,
                                                     dd4hep::SensitiveDetector sensDet) {
   dd4hep::xml::DetElement xmlDet = static_cast<dd4hep::xml::DetElement>(xmlElement);
   std::string name = xmlDet.nameStr();
@@ -1272,4 +1272,4 @@ static dd4hep::Ref_t createmuonSystemMuRWELL_o1_v01(dd4hep::Detector& lcdd, dd4h
   return detElement;
 }
 
-DECLARE_DETELEMENT(muonSystemMuRWELL_o1_v01, createmuonSystemMuRWELL_o1_v01)
+DECLARE_DETELEMENT(muonSystem_o1_v01, createmuonSystem_o1_v01)
--- a/detector/muonSystem/muonSystem_o1_v02.cpp
+++ b/detector/muonSystem/muonSystem_o1_v02.cpp
@@ -1,0 +1,903 @@
+/*
+ @author: Mahmoud Althakeel
+mahmoud.althakeel@cern.ch
+
+Factory for IDEA muon system — o1_v02 (sequence-based layer configuration)
+
+New in this version vs o1_v01:
+  <Barrel> and <Endcap> now accept an explicit ordered sequence of <layer>
+  child elements instead of fixed numDetectorLayers/numYokes counts.
+  This allows arbitrary detector/radiator arrangements, e.g.:
+    two detector layers → one radiator → one detector → one radiator
+
+Expected XML structure:
+<detector type="muonSystem_o1_v02" ...>
+  <dimensions x="chamberHalfThickness" y="chamberHalfY" z="chamberHalfZ" .../>
+  <sensitive type="tracker"/>
+  <generalParameters numSides="..." overlapY="..." overlapZ="..." clearance="..."/>
+
+  <Barrel rmin="..." length="...">
+    <layer type="detector"/>
+    <layer type="detector"/>
+    <layer type="radiator" thickness="300*mm" material="G4_Fe"/>
+    <layer type="detector"/>
+    <layer type="radiator" thickness="300*mm" material="G4_Fe"/>
+  </Barrel>
+
+  <Endcap rmin="..." z_offset="...">
+    <layer type="detector"/>
+    <layer type="detector"/>
+    <layer type="radiator" thickness="300*mm" material="G4_Fe"/>
+    <layer type="detector"/>
+    <layer type="radiator" thickness="300*mm" material="G4_Fe"/>
+  </Endcap>
+
+  <!-- mRWELL chamber slices (same as o1) -->
+  <slice x="..." material="..." [sensitive="true"] [vis="..."]/>
+  ...
+</detector>
+
+Sequence rules:
+  - Layers are placed in the order listed (innermost = first).
+  - Each <layer type="detector"> has the same chamber thickness derived from
+    the <dimensions> element (same as o1).
+  - Each <layer type="radiator"> requires thickness and material attributes.
+  - Endcap sequence is mirrored (reversed) for the negative endcap so that
+    detector layer 0 is always closest to the IP in both endcaps.
+*/
+
+#include "DD4hep/DetFactoryHelper.h"
+#include "DDRec/DetectorData.h"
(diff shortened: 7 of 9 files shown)


detector/muonSystem/muonSystem_o1_v01.cpp
```diff
@@ -1272,4 +1272,4 @@ static dd4hep::Ref_t createmuonSystemMuRWELL_o1_v01(dd4hep::Detector& lcdd, dd4h
   return detElement;
 }
 
-DECLARE_DETELEMENT(muonSystemMuRWELL_o1_v01, createmuonSystemMuRWELL_o1_v01)
\ No newline at end of file
+DECLARE_DETELEMENT(muonSystem_o1_v01, createmuonSystem_o1_v01)
```
s6anloes, line 1275:
```suggestion
DECLARE_DETELEMENT(muonSystem_o1_v01, createmuonSystem_o1_v01)
DECLARE_DETELEMENT(muonSystemMuRWELL_o1_v01, createmuonSystem_o1_v01)
```

Keep the old name as an alias to ensure backwards-compatibility. Otherwise all existing xml files will break!
And don't forget to add a emtpy new line at the end of the file

detector/muonSystem/README.md
```diff
@@ -2,9 +2,12 @@
 
 The IDEA detector concept includes a muon detection system and pre-shower designed using µRWELL technology. Each station consists of a large mosaic of 50 × 50 cm² µRWELL detectors. For more technical details about the new detector technology, see the [µRWELL](https://iopscience.iop.org/article/10.1088/1748-0221/10/02/P02008).
 
-## muonSystemMuRWELL_o1_v01.cpp
+## muonSystem_o1_v01.cpp
```
s6anloes, line 5:
```suggestion
## muonSystem_o1_v01.cpp (Formely named muonSystemMuRWELL_o1_v01.cpp)
```

detector/muonSystem/README.md
```diff
@@ -2,9 +2,12 @@
...
 The first version of the detailed muon system driver, it can be used to describe both barrel and endcap, if you want to eleminate one of them, just set the number of layers= 0.
 The code has been designed to be very flexible, where the user can choose the number of sides in the R-Phi plane, `numSides` (hexagon, octagon, etc), and the detector builder will automatically calculate the number and places of the copied chambers. Some of the code advantages: 
  * If the side length do not fit with an integer number of 50 × 50 cm² , the builder will make a chamber with unusual dimensions, which can fit the excess area at the end of the side.
  * The availability to make multiple layers with different inner radius and barrel length.
  * The code is very general, it can be used to describe any detector system made from repeated tiles (e.g. pre-shower) and has the capability to fill the gaps with unusual dimensions tiles.
```
s6anloes, line 10:
```suggestion
 * The code is very general, it can be used to describe any detector system made from repeated tiles (e.g. pre-shower) and has the capability to fill the gaps with unusual dimensions tiles.
 * 
 
 The old name `muonSystemMuRWELL_o1_v01` is retained as an alias for backward compatibility.
```

FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml
s6anloes, line 1: To be honest, I don't like having the renaming of the preshower and muon system files in `IDEA_o1_v03`.
It is not easy to predict the impact it will have on existing user workflows.
And there is no particular need or urgency to renaming the files in past versions. The current names have been alright for a few years now, so I'd say they can stay like this.
This renaming of the xml files should be a change for a new detector version.

FCCee/IDEA/compact/IDEA_o1_v04/IDEA_o1_v04.xml
s6anloes, line 1: Here the existing names would stay as well, based on my comment on `FCCee/IDEA/compact/IDEA_o1_v03/IDEA_o1_v03.xml`

FCCee/IDEA/compact/IDEA_o1_v04/MuonSystemMuRWELL_o1_v02.xml
```diff
@@ -0,0 +1,85 @@
...
+       <constant name = "GasLayerThick"                         value = "6*mm"/>
+       <constant name = "Cu2Thick"                              value = "0.005*mm"/>
+       <constant name = "KaptonThick"                           value = "0.05*mm"/>
+       <constant name = "CarbonFiberThick"                      value = "0.0001*mm"/>
+       <constant name = "CarbonFiber2Thick"                     value = "0.1*mm"/>
+       <constant name = "SiThick"                               value = "1.6*mm"/>
+       <constant name = "mRWELLTotalThickness"                  value = "G10_FR4Thick+CuThick+GasLayerThick+Cu2Thick+KaptonThick+CarbonFiberThick+CuThick+CarbonFiber2Thick+CuThick+SiThick"/> <!-- This sequense is the current order of mRWELL (total 10 slices) -->
+
+       <constant name = "mRWELLYLength"                         value = "250*mm"/>
+       <constant name = "mRWELLZLength"                         value = "250*mm"/>
```
s6anloes, line 31: From the `detector/muonSystem/README.md` it seems this is the half length of the modules?
Please clarify with a comment.
Maybe also in the `v01` of the file

FCCee/IDEA/compact/IDEA_o1_v04/MuonSystemMuRWELL_o1_v02.xml
```diff
@@ -0,0 +1,85 @@
...
+
+       <constant name = "overlapingY"                           value = "5*mm"/> <!-- the common distance between mRWELL chambers in Y direction -->
+       <constant name = "overlapingZ"                           value = "5*mm"/> <!-- the common distance between mRWELL chambers in Z direction -->
```
s6anloes, line 34: With `common distance` you mean overlap? The comment is not super clear.
Also, there is a typo in the variable name: it's overlapping (with two p's)

FCCee/IDEA/compact/IDEA_o1_v04/MuonSystemMuRWELL_o1_v02.xml
```diff
@@ -0,0 +1,85 @@
...
+       <constant name = "clearance"                             value = "1*mm"/> <!-- it's a small distance to be used to avoid overlapping between the different volumes ~ 1 mm -->
+     <!-- %%%%%%                      Iron yoke thickness (500 mm)          %%%%%% -->
+      <constant name = "YokeThickness"                          value = "500*mm"/>
+      <!-- %%%%%%                      Readout description                  %%%%%% -->
+      <constant name = "MuonSystemReadoutID"            type="string"  value = "system:5,type:-2,layer:2,chamber:13,slice:1,y:-10,z:-10"/>  <!-- The bit field is divided into 2^5 systems(IDEA sub-detectors), 2^2 types(0 = Barrel Muon System, +1 = positive endcap, and -1 = negative endcap"), 2^2 layers(number of layers in barrel for example) ,2^13 chambers(the number of muRWELL chambers in every layer), 2^1 slice(number of sensitive layers inside every chambers), and 2^10 y&z strips in every sensitive layer-->
+  </define>
+
+  <readouts>
+    <readout name="MuonSystemCollection">
+     <segmentation type="CartesianGridYZ" grid_size_y="1.2*mm" grid_size_z="1.2*mm"/>   <!-- Depending on strip pitch 1.4 mm  -->
```
s6anloes, line 44: In the comment it says 1.4 mm pitch instead of 1.2?
Also if these are strips, should one dimension be significantly larger than the other?
So if the pitch is `1.2 mm` in one direction, I'm expecting the strips to be O(cm) long, no?

FCCee/IDEA/compact/IDEA_o2_v01/IDEA_o2_v01.xml
```diff
@@ -69,7 +69,7 @@
   <include ref="DRBarrelTubes_o1_v01.xml"/>
 
   <!-- Import muon system -->
-  <include ref="../IDEA_o1_v03/MuonSystem_o1_v01.xml"/>
+  <include ref="../IDEA_o1_v03/MuonSystemMuRWELL_o1_v01.xml"/>
```
s6anloes, line 72: This would also be the old name.
And I agree that we need to wait for the okay from the IDEA spokespeople before we can introduce a new version with 4 layers

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+  //  General parameters
+  // -------------------------------------------------------------------------
+  auto generalParameters = xmlElement.child(_Unicode(generalParameters));
+  int    numSides   = generalParameters.attr<int>("numSides");
+  double overlapY   = generalParameters.attr<double>("overlapY");
+  double overlapZ   = generalParameters.attr<double>("overlapZ");
+  double clearance  = generalParameters.attr<double>("clearance");
+
+  double shapeAngle         = 360.0 / numSides;
+  double shapeAngle_radians = M_PI / numSides;
```
s6anloes, line 100: I'm guessing it's intentional to have the radian value to be half of the degree angle? Maybe you can rename the variable to make that clear?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+  }
+
+  // -------------------------------------------------------------------------
+  //  Chamber / detector-layer geometry 
+  // -------------------------------------------------------------------------
+  double sideLengthEstimate = 2.0 * radius * std::tan(shapeAngle_radians);
+  double chamberAngle_rad   = std::atan(dimensions.x() / dimensions.y());
+  double rectangleThickness = (2.0 * dimensions.y()) * std::sin(chamberAngle_rad)
+                            + (2.0 * dimensions.x()) * std::cos(chamberAngle_rad)
+                            + 1.2 * clearance;
```
s6anloes, line 126: What is this 1.2 factor? It's the only time it appears in all the file. Also why do you need a factor in front of the clearance at all (except some geometrical factors)? Couldn't you just make the clearance larger if you need more?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+  double rectangleAngle_rad = std::atan(rectangleThickness / dimensions.z());
+  double detectorVolumeThickness, endcapDetectorEnvZ;
+  if (sideLengthEstimate <= (2.0 * dimensions.y()) && numBarrelDetectorLayers == 1) {
+    detectorVolumeThickness = rectangleThickness * 1.33;
```
s6anloes, line 130: The 1.33 seems like a geometrical factor.
Can you explain in a comment where it comes from?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+                                      endcapDetectorLayerOuterRadius, EndcaptotalLength);
+
+  dd4hep::Position    unionPos (0.0, 0.0,  endcapOffset);
+  dd4hep::Position    unionPos2(0.0, 0.0, -endcapOffset);
+  dd4hep::Rotation3D  unionRot(dd4hep::RotationY(0.0 * dd4hep::degree));
+  dd4hep::Transform3D unionTransform (unionRot, unionPos);
+  dd4hep::Transform3D unionTransform2(unionRot, unionPos2);
+
+  dd4hep::UnionSolid barrelAndPositiveEndcap(BarrelEnv, EndcapEnv, unionTransform);
+  dd4hep::UnionSolid systemEnvelope(barrelAndPositiveEndcap, EndcapEnv, unionTransform2);
```
s6anloes, line 258: I don't think you really need this gymanstics with the `UnionSolid` for the system Envelope.
You could place the barrel and endcaps simply as sister volumes into the world volume, getting rid of the boolean operation.

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+  // -------------------------------------------------------------------------
+  //  BARREL volume
+  // -------------------------------------------------------------------------
+  dd4hep::PolyhedraRegular BarrelEnvBase(numSides, radius, barrelRMax, barrelLength);
+  dd4hep::PolyhedraRegular BarrelLastDetExt(numSides, lastDetLayerRMin, lastDetLayerRMax, barreltotalLength);
+
+  std::string barrelName = name + "-Barrel" + std::to_string(0);
+  dd4hep::Position    barrelUnionPos(0., 0., 0.);
+  dd4hep::Transform3D barrelUnionTransform(dd4hep::Rotation3D(dd4hep::RotationY(0.)), barrelUnionPos);
+  dd4hep::UnionSolid  barrelUnion(BarrelEnvBase, BarrelLastDetExt, barrelUnionTransform);
```
s6anloes, line 277:
```suggestion
  dd4hep::UnionSolid  barrelUnion(BarrelEnvBase, BarrelLastDetExt);
```
You don't _need_ to pass a transform. Per default identity rotation and no translation are assumed

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+  dd4hep::Volume      BarrelVolume(barrelName, barrelUnion, mat);
+
+  dd4hep::PlacedVolume barrelPhys = detectorVolume.placeVolume(
+      BarrelVolume, dd4hep::Transform3D(dd4hep::RotationZ(0.), dd4hep::Position(0., 0., 0.)));
```
s6anloes, line 281:
```suggestion
      BarrelVolume);
```
No transformation needed

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+  //  BARREL volume
+  // -------------------------------------------------------------------------
+  dd4hep::PolyhedraRegular BarrelEnvBase(numSides, radius, barrelRMax, barrelLength);
+  dd4hep::PolyhedraRegular BarrelLastDetExt(numSides, lastDetLayerRMin, lastDetLayerRMax, barreltotalLength);
+
+  std::string barrelName = name + "-Barrel" + std::to_string(0);
+  dd4hep::Position    barrelUnionPos(0., 0., 0.);
+  dd4hep::Transform3D barrelUnionTransform(dd4hep::Rotation3D(dd4hep::RotationY(0.)), barrelUnionPos);
+  dd4hep::UnionSolid  barrelUnion(BarrelEnvBase, BarrelLastDetExt, barrelUnionTransform);
+  dd4hep::Volume      BarrelVolume(barrelName, barrelUnion, mat);
```
s6anloes, line 278: But ideally we'd get rid of the boolean oprations, especially since they are at the top-level volume, meaning you'll pay the navigation overhead for every track.
Maybe it's possible to make the last layer and the rest of the barrel adjacent sister volumes instead?
By the way, why is it only the last layer that extents to `barreltoalLength`? Is that just an assumption of the builder? Maybe we can add documentation on this, or maybe I have missed it

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+
+      // ---- Sides --------------------------------------------------------
+      for (int side = 0; side < numSides; ++side) {
+        int sideID = (seqIdx + 1) * 100 + (side + 1);
+
+        dd4hep::Box sideEnvelope (sideEnvX, sideEnvY,  sideEnvZ);
+        dd4hep::Box sideEnvelope2(sideEnvX, sideEnvY2, sideEnvZ);
+        std::string sideName = dd4hep::xml::_toString(sideID, "side%d");
+        dd4hep::Volume sideVol (sideName, sideEnvelope,  mat);
+        dd4hep::Volume sideVol2(sideName, sideEnvelope2, mat);
```
s6anloes, line 335: Not really necessary to have two declarations of the shape and volume if you only use one based on $side % 2 == 0$.
Better (in the sense clearer to read) would be to make this once for the Box shape, and the the rest is almost identical between `sideVol` and `sideVol2` so you no longer need the distinction between the two (get rid of `sideVol2`)

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+
+        double angle_degrees = shapeAngle * side;
+        double angle_radians = angle_degrees * M_PI / 180.0;
+        double sideXOffset  = (barrelLayerRMid - detectorVolumeThickness / 2.0) * std::cos(angle_radians + shapeAngle_radians);
+        double sideYOffset  = (barrelLayerRMid - detectorVolumeThickness / 2.0) * std::sin(angle_radians + shapeAngle_radians);
+        double sideXOffset2 = (barrelLayerRMid + detectorVolumeThickness / 2.0) * std::cos(angle_radians + shapeAngle_radians);
+        double sideYOffset2 = (barrelLayerRMid + detectorVolumeThickness / 2.0) * std::sin(angle_radians + shapeAngle_radians);
```
s6anloes, line 342:
```suggestion
double placement_radius = barrelLayerRMid
              + (side % 2 == 0 ? -1.0 : 1.0) * detectorVolumeThickness / 2.0;

double sideXOffset = radius * std::cos(angle_radians + shapeAngle_radians);
double sideYOffset = radius * std::sin(angle_radians + shapeAngle_radians);
```

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+
+        dd4hep::Position     sideTrans;
+        dd4hep::PlacedVolume sidePhys;
+        if (side % 2 == 0) {
+          sideTrans = dd4hep::Position(sideXOffset,  sideYOffset,  0.0);
+          sidePhys  = BarrelDetectorLayerVolume.placeVolume(sideVol,  dd4hep::Transform3D(sideRotation, sideTrans));
+        } else {
+          sideTrans = dd4hep::Position(sideXOffset2, sideYOffset2, 0.0);
+          sidePhys  = BarrelDetectorLayerVolume.placeVolume(sideVol2, dd4hep::Transform3D(sideRotation, sideTrans));
+        }
```
s6anloes, line 355: Then you can get rid if this if statement , removing the need for these duplicated variables

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+        sideDE.setPlacement(sidePhys);
+        sideVol.setVisAttributes(lcdd,  xmlDet.visStr());
+        sideVol2.setVisAttributes(lcdd, xmlDet.visStr());
+
+        // ---- Rectangles -------------------------------------------------
+        int numRectangles = (int)(barrelLayerLength / (2.0 * dimensions.z() - overlapZ));
+        for (int rectangle = 0; rectangle <= numRectangles; ++rectangle) {
+          double rectangleEnvY = (side % 2 == 0) ? sideEnvY : sideEnvY2;
+          double rectangleEnvZ = (rectangle == numRectangles)
+                                     ? (remainderZ * dimensions.z() + clearance / 4.0)
```
s6anloes, line 366: Where is this `4.0` coming from? Can you comment?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+        } else {
+          sideTrans = dd4hep::Position(sideXOffset2, sideYOffset2, 0.0);
+          sidePhys  = BarrelDetectorLayerVolume.placeVolume(sideVol2, dd4hep::Transform3D(sideRotation, sideTrans));
+        }
+        dd4hep::DetElement sideDE(BarrelDetectorLayerDE, sideName + "DE", sideID);
+        sideDE.setPlacement(sidePhys);
+        sideVol.setVisAttributes(lcdd,  xmlDet.visStr());
+        sideVol2.setVisAttributes(lcdd, xmlDet.visStr());
+
+        // ---- Rectangles -------------------------------------------------
```
s6anloes, line 361: Can you add a comment explaning what the `Rectangles` represent in the detector?
`Sides`, `chambers` and so on are a bit descriptive, but just 'Rectangle` is not really clear.

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+        int numRectangles = (int)(barrelLayerLength / (2.0 * dimensions.z() - overlapZ));
+        for (int rectangle = 0; rectangle <= numRectangles; ++rectangle) {
+          double rectangleEnvY = (side % 2 == 0) ? sideEnvY : sideEnvY2;
+          double rectangleEnvZ = (rectangle == numRectangles)
+                                     ? (remainderZ * dimensions.z() + clearance / 4.0)
+                                     : (dimensions.z() + clearance / 2.0);
```
s6anloes, line 367: Also this division by `2.0`. I see this a lot in the code that clearance appears with different factors.
This is not good practice. It seems like maybe you need to introduce different clearances in the xml file, or check if you can remove these factors

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+          double rectangleEnvX = (rectangleEnvY <= dimensions.y())
+                                     ? dimensions.x()
+                                     : detectorVolumeThickness / 4.5;
```
s6anloes, line 370: Is the `4.5` a geometric factor? If so, explain in a comment. If not, then we need to find a better solution than this magic number

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+          } else {
+            rectangleRemainderEnvXPos = 0.0;
+            rectangleEnvZPos = barrelLayerLength / 2.0 - dimensions.z()
+                               - (rectangle * (2.0 * dimensions.z() - overlapZ)) - clearance;
+          }
+
+          double yRotation = std::atan(rectangleEnvX / (rectangleEnvZ - 2.0 * overlapZ));
+          double yRemainderRotation;
+          dd4hep::RotationY rotationY(yRotation);
+          if (rectangleEnvZ <= (dimensions.z() + clearance / 2.0) / 10.0)
```
s6anloes, line 392: `10.0`? These numbers, Mahmoud, what do they mean?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+            bns << "-MuRWELL_Barrel_" << barrelIdCounter++;
+            std::string BarrelChamberName = name + bns.str();
+
+            double rectangleRemainderY, rectangleRemainderREnvYPos;
+            if (numChambersInRectangle == 0) {
+              rectangleRemainderY = std::abs(std::fmod((2.0 * rectangleEnvY - clearance),
+                                                        (2.0 * dimensions.y() - clearance))) / (2.0 * dimensions.y());
+              rectangleRemainderREnvYPos = (chamberIndex * 2.0 * dimensions.y()) - (overlapY * chamberIndex)
+                                         + dimensions.y_offset() - rectangleEnvY
+                                         + rectangleRemainderY * dimensions.y() + clearance / 20.0;
```
s6anloes, line 439: `20.0` ...

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+            } else {
+              rectangleRemainderY = std::fmod(2.0 * (rectangleEnvY - clearance),
+                                               (2.0 * dimensions.y() - overlapY)) / (2.0 * dimensions.y());
+              rectangleRemainderREnvYPos = (chamberIndex * 2.0 * dimensions.y()) - (overlapY * chamberIndex)
+                                         + dimensions.y_offset() - rectangleEnvY
+                                         + rectangleRemainderY * dimensions.y() + 1.5 * clearance;
```
s6anloes, line 445: Same issue `1.5`

s6anloes, line 445: Also `rectangleRemainderREnvYPos` is almost identical between the two cases, except for the `clearance` factor. Do we really need this different factor?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+                lcdd.material(dimensions.materialStr()));
+
+            double envYPos = (chamberIndex * 2.0 * dimensions.y()) - (overlapY * chamberIndex)
+                           + dimensions.y_offset() - rectangleEnvY + dimensions.y() + clearance / 20.0;
+
+            double zRotation = 0.0, rectangleRemainderZRotation = 0.0;
+            if (numChambersInRectangle != 0) {
+              zRotation = std::atan(dimensions.x() / (dimensions.y() - 2.0 * overlapY));
+              rectangleRemainderZRotation = std::atan(dimensions.x() /
+                  (rectangleRemainderY * dimensions.y() - overlapY * 1.5));
```
s6anloes, line 464: Why does the overlap have different factors here?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+        dd4hep::PlacedVolume barrelRadiatorPhys = barrelRadiatorEnvVol.placeVolume(
+            barrelRadiatorVol, dd4hep::Transform3D(dd4hep::RotationY(0.), dd4hep::Position(0, 0, 0)));
+        dd4hep::DetElement barrelRadiatorDE(barrelRadiatorEnvelopeDE, barrelRadiatorName + "DE", sideID);
+        barrelRadiatorDE.setPlacement(barrelRadiatorPhys);
+        barrelRadiatorVol.setVisAttributes(lcdd.visAttributes("yoke_vis"));
+      }
+
+      dd4hep::PlacedVolume radiatorLayerPhys = BarrelVolume.placeVolume(
+          BarrelRadiatorLayerVolume,
+          dd4hep::Transform3D(dd4hep::RotationZ(0.), dd4hep::Position(0., 0., 0.)));
```
s6anloes, line 594: Same here

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+      detectorLayerPhys.addPhysVolID("layer", numBarrelLayer);
+      dd4hep::DetElement BarrelDetectorLayerDE(
+          BarrelDE, name + "-Barrel_DetectorLayerDE_" + std::to_string(numBarrelLayer + 1), numBarrelLayer + 1);
+      BarrelDetectorLayerDE.setPlacement(detectorLayerPhys);
+      BarrelDetectorLayerVolume.setVisAttributes(lcdd.visAttributes("no_vis"));
+
+      // ---- Sides --------------------------------------------------------
+      for (int side = 0; side < numSides; ++side) {
+        int sideID = (seqIdx + 1) * 100 + (side + 1);
```
s6anloes, line 329: Add a comment on the logic of the `sideID` here and the other places it is calculated

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+        for (int side = 0; side < numSides; ++side) {
+          int sideID = (numEndcapLayer + 1) * 100 + (side + 1);
+
+          dd4hep::Trapezoid endcapDetectorSideTrap(
+              detectorVolumeThickness / 2.0, detectorVolumeThickness / 2.0,
+              endcapDetectorSideLength / 2.0, endcapDetectorSideTrapLength / 2.0,
+              endcapDetectorSideTrapYLength / 2.0);
+          dd4hep::Box endcapDetectorSideBox(
+              detectorVolumeThickness / 2.0, endcapDetectorSideBoxLength / 2.0, endcapDetectorSideBoxYLength / 2.0);
+          dd4hep::UnionSolid endcapDetectorSideEnvelope(
```
s6anloes, line 662: Another `UnionSolid` I'd like to avoid, though for this one I'm not sure I understand the geometry exactly. Thus I can't tell if this can be avoided or not.
Can you explain the need for this boolean shape?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+            double z3 =  endcapDetectorEnvZ / 4.0;
+            double z4 =  endcapDetectorEnvZ * 3.0 / 4.0;
+            if      (side % 4 == 0) endcapDetectorSideEnvTrans = dd4hep::Position(ecXOffset, ecYOffset, z1);
+            else if (side % 4 == 1) endcapDetectorSideEnvTrans = dd4hep::Position(ecXOffset, ecYOffset, z2);
+            else if (side % 4 == 2) endcapDetectorSideEnvTrans = dd4hep::Position(ecXOffset, ecYOffset, z3);
+            else                    endcapDetectorSideEnvTrans = dd4hep::Position(ecXOffset, ecYOffset, z4);
+          } else {
+            double zPos = (side % 2 == 0) ? -endcapDetectorEnvZ / 2.0 : endcapDetectorEnvZ / 2.0;
+            endcapDetectorSideEnvTrans = dd4hep::Position(ecXOffset, ecYOffset, zPos);
+          }
```
s6anloes, line 693: Can you explain the siginifance/meaning of this block in a comment?
Why is `side%4` an important quantity to decide the translation?
Seems quite hard-coded...
Unless this is supposed to represent the four quadrants?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+          for (int rectangle = 0; rectangle <= numRectangles; ++rectangle) {
+            double rectangleEnvY =
+                (endcapDetectorLayerOuterRadius - rectangle * (2.0 * dimensions.y() - overlapY))
+                * std::tan(shapeAngle_radians);
+            double rectangleEnvZ;
+            if (rectangle == numRectangles)
+              rectangleEnvZ = endcapRemainderZ * dimensions.z() + clearance / 4.0;
+            else
+              rectangleEnvZ = dimensions.z() + clearance / 2.0;
+            double rectangleEnvX = (rectangleEnvY <= dimensions.y()) ? dimensions.x() : detectorVolumeThickness / 4.5;
```
s6anloes, line 714: Again different factors in front of `clearance` and `detectorVolumeThickness`, but they seem consisten with before, so maybe add a comment explaning where they come from?

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+
+              dd4hep::Box    rrmYEnv(dimensions.x(), rectangleRemainderY * dimensions.y(), chamberHalfZ);
+              dd4hep::Volume rectangleRemainderYEnvVolume(
+                  EndcapChamberName + "rectangleRemainderY", rrmYEnv, lcdd.material(dimensions.materialStr()));
+
+              double envYPos = (chamberIndex * 2.0 * dimensions.y()) - (overlapY * chamberIndex)
+                             + dimensions.y_offset() - rectangleEnvY + dimensions.y() + 0.005;
+              double rectangleRemainderREnvYPos = (chamberIndex * 2.0 * dimensions.y()) - (overlapY * chamberIndex)
+                             + dimensions.y_offset() - rectangleEnvY
+                             + rectangleRemainderY * dimensions.y() + 0.005;
```
s6anloes, line 766: These `0.005` are not good. Do they have a physical meaning, or were they only added to avoid overlaps?
If it's due to overlaps, please find a waay to resolve them 'analytically'

detector/muonSystem/muonSystem_o1_v02.cpp
```diff
@@ -0,0 +1,903 @@
...
+        barrelRadiatorEnvVol.setVisAttributes(lcdd, xmlDet.visStr());
+
+        dd4hep::Trapezoid barrelRadiator(
+            barrelLength / 2.0, barrelLength / 2.0,
+            barrelRadiatorSideLength / 2.0, barrelRadiatorSideLength2 / 2.0,
+            spec.thickness / 2.0);
+        std::string barrelRadiatorName = dd4hep::xml::_toString(side, "barrelRadiator%d");
+        dd4hep::Volume barrelRadiatorVol(barrelRadiatorName, barrelRadiator, spec.matObj);
+        dd4hep::PlacedVolume barrelRadiatorPhys = barrelRadiatorEnvVol.placeVolume(
+            barrelRadiatorVol, dd4hep::Transform3D(dd4hep::RotationY(0.), dd4hep::Position(0, 0, 0)));
```
s6anloes, line 586: What is the point of `barrelRadiatorEnvVol`? It has the exact same dimensions as `barrelRadiatorVol` and is placed inside `barrelRadiatorVol` without any transformation, so the two are exactly the same?
Can't you just remove the envelope volume?

s6anloes (changes requested):
Hi Mahmoud,
thanks for the PR, I have left some comments. If you have questions, feel free to ask.

One thing I noticed when benchmarking the asymmetric constructor with the XML file you sent me privately:
the memory consumption is increased by a factor of 3 (1 GB for the Muon system from `IDEA_o1_v04` and 3 GB for the asym file you sent me).
Now this is of course not an apples-to-apples comparison since the number of layers is very different.
Could I ask you to create an XML file for the asymmetric `muonSystem_o1_v02.cpp` driver, which has the exact same geometry as the current `IDEA_o1_v04` muon system (which uses `muonSystem_o1_v01.cpp`)?
I'd like to understand how big the difference in memory is for two identical geometries.
Then we know if we need some drastic improvements in the code.

Thanks!

PR #619, opened 2026-07-21
Remove a few unused functions in detectorCommon

jmcarcell:
BEGINRELEASENOTES
- Remove a few unused functions in detectorCommon

ENDRELEASENOTES

I was able to build k4RecCalorimeter with these changes.

### DIFF
--- a/detectorCommon/include/detectorCommon/DetUtils_k4geo.h
+++ b/detectorCommon/include/detectorCommon/DetUtils_k4geo.h
@@ -16,32 +16,18 @@
 #include "DDSegmentation/CartesianGridXYZ.h"
 #include "DDSegmentation/PolarGridRPhi.h"
 
-#include "G4Step.hh"
-
 // CLHEP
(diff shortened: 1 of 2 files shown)


giovannimarchiori:
Hi @jmcarcell 

the functions are unused so it does not hurt to remove them, but maybe the utility method that returns min/max value of a bitfield is something that could still be useful some time? I would let @andresailer or @tmadlener say the final word on this

Cheers,
  Giovanni

jmcarcell:
If they are useful then we can write them again, given that now the cost to do so is very little. I think it's you who has the last word since you wrote them in 2023.

giovannimarchiori:
not really, I just moved to k4geo code that was in FCCDetectors and was there since at least 2020... https://github.com/HEP-FCC/FCCDetectors/blob/320954b289b0b8a4a19163549db4db8494459113/Detector/DetCommon/src/DetUtils.cpp

;)

jmcarcell:
Then even worse :), let's remove them

PR #620, opened 2026-07-27
fix(sim): use keyed calorimeter hit lookups

jbeirer:
BEGINRELEASENOTES
- Fixed severe simulation slowdowns in DRcalo and pre-shower sensitive detectors by using keyed calorimeter hit lookups.

ENDRELEASENOTES

## Summary

The `IDEA_o1_v03` simulation of the `p8_ee_Zbb_ecm91` sample did not complete even a single event within the configured timeout, as reported in key4hep/k4Bench#83.

Profiling showed that the calorimeter sensitive detectors performed a linear search through the full hit collection for every simulation step. As the collection grows during large hadronic showers, this leads to approximately quadratic runtime scaling.

This PR:

* replaces linear `Geant4HitCollection::find()` searches with keyed `findByKey()` lookups
* uses the existing cell ID as the collection key
* updates `CaloPreShowerSDAction` to use matching keyed insertion
* reduces individual hit lookups from O(N) to O(log N)
* preserves the existing hit creation and accumulation behaviour

With these changes, the previously stuck `Z → bb` event simulation completes successfully.

fyi @BrieucF

### DIFF
--- a/plugins/CaloPreShowerSDAction.cpp
+++ b/plugins/CaloPreShowerSDAction.cpp
@@ -81,7 +81,7 @@ namespace sim {
     Geant4HitCollection* coll = (layer == m_userData._firstLayerNumber ? collection(m_userData._preShowerCollectionID)
                                                                        : collection(m_collectionID));
 
-    Hit* hit = coll->find<Hit>(CellIDCompare<Hit>(cell));
+    Hit* hit = coll->findByKey<Hit>(cell);
     if (h.totalEnergy() < std::numeric_limits<double>::epsilon()) {
       return true;
     } else if (!hit) {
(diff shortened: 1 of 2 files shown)


SanghyunKo:
Looks good to me

jbeirer:
Thanks @SanghyunKo for the review, will take the liberty to merge to have this fixed in the next nightly

jmcarcell:
Did you check you get exactly the same results as before? If it's only a search that is faster that should be the case

jbeirer:
> Did you check you get exactly the same results as before? If it's only a search that is faster that should be the case

Hi @jmcarcell, I couldn't compare the Z→bb sample, since unpatched it doesn't finish even a single event and therefore never writes an output file. But for `IDEA_o1_v03` I compared 10 × 20 GeV π⁻ events and all relevant branches are bit-identical.

PR #621, opened 2026-07-27
Fix the thickness of the sense wires in the drift chamber

jmcarcell:

BEGINRELEASENOTES
- Fix thickness in the sense wires in the drift chamber, where the radius of the tubes is set to the thickness causing them to have a diameter of 2*thickness

ENDRELEASENOTES

All the other tubes seem to have their radius correctly set to thickness / 2, see, for example, https://github.com/key4hep/k4geo/blob/e26121080e1a5f0745c3e690ee0fb357ed808269/detector/tracker/DriftChamber_o1_v02.cpp#L364


s6anloes:
Thanks Juan Miguel!
For anyone interested, this is the impact on the material budget:
Before
image

After:
image

Should be a factor of 4 in the sense wire contribution due to the area scaling with $r^2$, which seems about right from the plots

PR #623, opened 2026-07-30
Fix z index decoding in turbine endcap getLocalZ

jbeirer:
BEGINRELEASENOTES
Fix z index decoding in turbine endcap getLocalZ
ENDRELEASENOTES

Introduced in #608


giovannimarchiori (approved):
clearly a bug, thanks for the fix

PR #624, opened 2026-07-30
Fix flat index in turbine endcap local z lookup

jbeirer:
BEGINRELEASENOTES
Fix flat index in turbine endcap local z lookup

ENDRELEASENOTES

The local `z` values are stored in a one-dimensional vector, but each value is identified by two indices: the radial layer (`iRho`) and the `z` layer (`iZ`).

The previous calculation, `iRho * iZ`, does not uniquely identify a cell. Different index pairs can produce the same result, for example, `(1, 2)` and `(2, 1)` both map to index `2`. Some values are therefore written to the wrong position or overwrite another cell’s value.

The correct flat index is `iRho * numReadoutLayersZ + iZ`, which assigns every `(iRho, iZ)` pair its own position in the vector.

### DIFF
--- a/detectorSegmentations/include/detectorSegmentations/FCCSWEndcapTurbine_k4geo.h
+++ b/detectorSegmentations/include/detectorSegmentations/FCCSWEndcapTurbine_k4geo.h
@@ -18,15 +18,17 @@ namespace DDSegmentation {
   class EndcapTurbineWheelLocalZ {
   public:
     // constructor with number of rho and z cells in wheel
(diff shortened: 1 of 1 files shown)


giovannimarchiori:
@varnes can you please have a look at this PR? it seems OK for me

varnes:
HI @giovannimarchiori Yes, I agree with this change.  Thanks @jbeirer !

giovannimarchiori (approved):

PR #629, opened 2026-08-04
Fixup small errors in cell positioning, and re-syncronize calibration and upstream xmls

varnes:
- [x ] the code is sufficiently well documented (e.g. inline comments)
- [ NA] the relevant README(s) were updated. See e.g. https://github.com/key4hep/k4geo/blob/main/detector/calorimeter/README.md or https://github.com/key4hep/k4geo/blob/main/FCCee/IDEA/compact/README.md
- [ x] the code is covered by tests. See https://github.com/key4hep/k4geo/blob/main/test/CMakeLists.txt
- [ x] the PR source branch has been rebased (`--ff-only`) to `k4geo/main`
- [x ] the PR does not contain any additions or modifications that do not belong to it
- [ x] The release notes below contain a succinct and comprehensive description of the changes that are sufficiently self-explanatory

BEGINRELEASENOTES

This PR fixes a few small errors in cell positioning (  detectorSegmentations/src/FCCSWEndcapTurbine_k4geo.cpp ) and updates the xml to calculate the geometry in a consistent way as was done in the code.  The detector code was also modified to read the radiusRatio parameter from the xml rather than calculating it, to prevent any divergences between values in the future.  Some variables in the xml were also given more descriptive names to improve readability.

In the process of these updates, I noticed that the calibration and upstream versions of the xml had fallen out of sync with the default xml, so these have been updated as well.

ENDRELEASENOTES

### DIFF
--- a/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/ECalEndcaps_Turbine_o1_v03.xml
+++ b/FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/ECalEndcaps_Turbine_o1_v03.xml
@@ -27,7 +27,7 @@
     <!-- space between the cryostat and calorimeter -->
     <constant name="BathThicknessFront" value="10*mm"/>
     <constant name="BathThicknessBack" value="40*mm"/>
-    <constant name="BathThicknessInner" value="0*mm"/>
+    <constant name="BathThicknessInner" value="0*mm"/> <!-- placeholder in case future designs require space here -->
     <constant name="BathThicknessOuter" value="50*mm"/>
     <!-- electromegnetic calorimeter: EMEC -->
     <constant name="EMEC_z1" value="CryoEndcap_z1 + CryoEMECThicknessFront + BathThicknessFront"/>offset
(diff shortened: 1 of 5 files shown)


FCCee/ALLEGRO/compact/ALLEGRO_o1_v03/ECalEndcaps_Turbine_o1_v03.xml
```diff
@@ -27,11 +27,12 @@
     <!-- space between the cryostat and calorimeter -->
     <constant name="BathThicknessFront" value="10*mm"/>
     <constant name="BathThicknessBack" value="40*mm"/>
+    <constant name="BathThicknessInner" value="0*mm"/>
```
giovannimarchiori: is 0 correct or a typo?

varnes: Hi @giovannimarchiori  It's correct.  The current design does not have space there, but I added the variable in case future designs require it. I put in a comment to that effect so it's clear to others looking at the xml.

giovannimarchiori (approved):

PR #328, opened 2024-03-05
Make LCIO an optional dependency

zaborowska:
BEGINRELEASENOTES
- Make LCIO an optional dependency. If LCIO is not found, some detectors (trackers) will not be built.

ENDRELEASENOTES

The list of the detectors in CMakeLists that depend on LCIO is hard-coded and needs to be modified for each new LCIO-dependent detector.

PR #631, opened 2026-08-10
Properly propagate usage of LCIO to downstream consumers

madbaron:

BEGINRELEASENOTES
- Fixed propagation of cmake arguments to turn off lcio dependencies.

ENDRELEASENOTES

This PR fixes an issue where k4geo would look for lcio unconditionally when included as dependency by another package in the stack, even if it was explicitly built without LCIO.


tmadlener (approved):
Just for documentation purposes for future archeology. This was most likely missed in #328 and not discovered until now because we never tried to build without LCIO in the stacks.

PR #632, opened 2026-08-12
fix: side leak in reflect loop

madbaron:

BEGINRELEASENOTES
- Fixed a leak of `encoder[lcio::LCTrackerCellID::side()]` in the computation of neighbourSurfacesData for TrackerEndcap_o2_v07

ENDRELEASENOTES

When studying how to remove the LCIO dependencies from k4geo (more on that later), I noticed what looks like a bug in the computation of neighbourSurfacesData.
The `encoder[lcio::LCTrackerCellID::side()]` would leak out of module/sensor loop when reflect was used.

I am not sure what's the impact of this bug on physics, but it looks like it's been there for a while.

This PR fixes the bug for TrackerEndcap_o2_v07_geo.cpp (used by the MuColl detector concepts), but also TrackerEndcap_o2_v06, TrackerEndcap_o2_v05 and VertexEndcap_o1_v05 used by CLIC and CLD have the same issue.
I'm happy to change those too, but maybe there's a preference to keep the older detector models as they were for reproducibility of results?

### DIFF
--- a/detector/tracker/TrackerEndcap_o2_v07_geo.cpp
+++ b/detector/tracker/TrackerEndcap_o2_v07_geo.cpp
@@ -279,6 +279,8 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
               continue; // out of disk
 
             // encoding
+            // NB: side has to be set explicitly here, the reflect block below leaves it at bwd
+            encoder[lcio::LCTrackerCellID::side()] = lcio::ILDDetID::fwd;
             encoder[lcio::LCTrackerCellID::module()] = newmodule;
             encoder[lcio::LCTrackerCellID::sensor()] = newsensor;
 


tmadlener:
`TrackerEndcap_o2_v07_geo`, is this in use exclusively by mucoll geometries?

tmadlener:
@Zehvogel did you ever notices something weird in CLD related to this? It's not clear to me what the effects of this would be for tracking or whether this specific bit of information is actually used somewhere downstream.

Zehvogel:
> did you ever notices something weird in CLD related to this?

Not really. From a quick look, I also only see explicit usage of `lcio::ILDDetID::fwd` in the preparation of tracks in Pandora while creating additional track states at the endcap calorimeter face. But as it is used only to identify the calorimeter, it has nothing to do with the changes in this PR. https://github.com/iLCSoft/DDMarlinPandora/blob/e4c53a5bb99a1995376c8d64ec98e6bb749d6074/src/DDTrackCreatorBase.cc#L419-L424

I don't know if there are any other places where this information is used implicitly. Maybe @andresailer remembers :)

I could imagine that it gets used in the tracking somewhere, but I would hope that this change does not break anything. I would expect that this bug can only cause the tracking to try to add hits from the wrong side of the detector to a track, which are then immediately discarded for fitting too badly... :shrug:

andresailer:
`neighbourSurfacesData` was not used in production. This was a development to potentially speed up reconstruction.

tmadlener:
OK. In that case this should be safe to apply also to the other constructors. The bigger question is whether we want to keep the `neighbourSurfacesData` around in the first place if they are unused?

edit: pulling that question into an issue.

tmadlener:
This could also become obsolete if we follow through with #635

tmadlener:
superseded by #636

PR #633, opened 2026-08-12
Remove dead code to facilitate dropping LCIO as a dependency

madbaron:

BEGINRELEASENOTES
- Removed unused dependencies on LCIO classes from TrackerBarrel_o1_v06 and ZSegmentedPlanarTracker
- Removed drivers not depending on LCIO from K4GEO_USE_LCIO exclusion list in CMakeLists
 
ENDRELEASENOTES

A bit of housekeeping before getting rid of the LCIO dependencies (planned to do by swapping the LCIO encoders with the drop-in dd4hep classes already used in some parts of the repo).
I kept this separate because it's technically unrelated to the port.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,15 +88,11 @@ if(NOT K4GEO_USE_LCIO)
     TrackerBarrel_o1_v03_geo.cpp
     TrackerBarrel_o1_v04_geo.cpp
     TrackerBarrel_o1_v05_geo.cpp
-    TrackerBarrel_o1_v06_geo.cpp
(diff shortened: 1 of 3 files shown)


tmadlener:
These constructors are in use by quite a few geometries so I would appreciate a second set of eyes, but this looks like it does indeed only remove dead code. The CMake changes are for shuffling these files into the build without LCIO, right?

madbaron:
Yes. These 4 files never deserved to be in the exclusion list because they didn't use LCIO, so this is just fixing that unnecessary exclusion.

tmadlener (approved):

PR #634, opened 2026-08-12
LCIO dependency removal (step #2)

madbaron:

BEGINRELEASENOTES
- Removed dependency on LCIO by using dd4hep::DDSegmentation::BitFieldCoder instead of UTIL::BitField64. This allows to build and run the software stack without any of the LCIO-related packages.

ENDRELEASENOTES

This PR removes the LCIO dependency from k4geo entirely.

The whole LCIO dependency is `UTIL::BitField64` plus eight compile-time constants, used by ten tracker drivers to pack a 32-bit cellID for the `dd4hep::rec::NeighbourSurfacesData` extension. 

These have a DD4hep equivalent in `UTIL::BitField64` → `dd4hep::DDSegmentation::BitFieldCoder`, taken straight from `sens.readout().idSpec().decoder()` instead of re-parsing `fieldDescription()`.

DD4hep ships no equivalent (or at least I could not find one) for `lcio::ILDDetID::{bwd,barrel,fwd}` so the cost is a new `detector/include/DetSide_k4geo.h`. **Note that paying this costs enables to build/run a complete software stack based on gaudi/edm4hep without any of the legacy LCIO/Marlin packages**

Since we don't need LCIO anymore, the build system drops `find_package(LCIO)`, the `LCIO::lcio` shim, the `K4GEO_USE_LCIO` option and its
source-exclusion list, and both `target_link_libraries(... LCIO::lcio)` calls.

### Additional Notes

* **The cellID fields are now addressed by name, where LCIO addressed them by position.** `LCTrackerCellID::subdet()` and friends returned `static unsigned` field *indices* (0–4), not strings, so the old code indexed into the readout's field list. Every readout reachable from the ten drivers using this was audited and all are `system, side, layer, module, sensor` in that order, so the two are equivalent for everything k4geo ships — confirmed by the byte-identical geometry check below.
Names are preferable going forward: a compact file with a reordered readout now throws `unknown name` at construction instead of silently encoding into the wrong fields. Note LCIO called field 0 `subdet` while every k4geo readout calls it `system`.
* **`BitFieldCoder` is stateless**, so the mutable `BitField64` value becomes a single `dd4hep::CellID encoderValue` carried through the same code paths. This is deliberate: several endcap drivers depend on fields set before a loop persisting into it.
* **This reverts #631.** That PR made `find_dependency(LCIO)` conditional and exported `k4geo_USE_LCIO` so consumers could tell whether the LCIO-dependent constructors were present.
* Note that `k4geoG4` was linking `LCIO::lcio` despite containing no LCIO user at all.

### Verification

Geometry output is unchanged. A dump of every `NeighbourSurfacesData::sameLayer` entry was compared before/after across MAIA_v0, MuSIC_v2, CLIC_o3_v15, CLIC_o3_v07, CLIC_o2_v04_p1 and CLD_o2_v08:
**0 differing lines out of ~2.6M neighbour entries.** 

Also checked, in `ghcr.io/muoncollidersoft/mucoll-sim-ubuntu24`:
* configure + build + install with LCIO stripped from the environment entirely;
* neither `libk4geo.so` nor `libk4geoG4.so` links `liblcio`;
* all ten previously-excluded detectors are present in the resulting library;
* a downstream `find_package(k4geo)` configures against the install tree with no LCIO available;
* no LCIO mention remains in the installed CMake config;
* `ctest`: 43/45. The two failures (`t_name_check_ILD_FCCee_v01`/`_v02`) are missing beampipe STL files and fail identically on unmodified `main` in the same container (the stl files are not installed in the MuColl image).

### DIFF
--- a/detector/tracker/SiTrackerEndcap_o2_v02ext_geo.cpp
+++ b/detector/tracker/SiTrackerEndcap_o2_v02ext_geo.cpp
@@ -20,7 +20,7 @@
 #include "DD4hep/DetFactoryHelper.h"
 #include "DDRec/DetectorData.h"
 #include "DDSegmentation/BitFieldCoder.h"
-#include "TrackerCellID_k4geo.h"
+#include "DetSide_k4geo.h"
 #include "XML/Utilities.h"
 #include <map>
 
@@ -59,7 +59,7 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
 
   const BitFieldCoder& encoder = *sens.readout().idSpec().decoder();
   dd4hep::CellID encoderValue = 0;
-  encoder.set(encoderValue, k4geo::TrackerCellID::subdet, det_id);
+  encoder.set(encoderValue, "system", det_id);
 
   // --- create an envelope volume and position it into the world ---------------------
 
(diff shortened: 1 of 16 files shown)


detector/tracker/SiTrackerEndcap_o2_v02ext_geo.cpp
```diff
@@ -194,20 +193,20 @@ static Ref_t create_detector(Detector& theDetector, xml_h e, SensitiveDetector s
...
         dd4hep::CellID cellID_reflect;
         if (reflect) {
-          encoder[lcio::LCTrackerCellID::side()] = lcio::ILDDetID::bwd;
-          encoder[lcio::LCTrackerCellID::layer()] = l_id;
-          encoder[lcio::LCTrackerCellID::module()] = mod_num;
-          encoder[lcio::LCTrackerCellID::sensor()] = k;
+          encoder.set(encoderValue, "side", k4geo::DetSide::bwd);
+          encoder.set(encoderValue, "layer", l_id);
+          encoder.set(encoderValue, "module", mod_num);
+          encoder.set(encoderValue, "sensor", k);
```
tmadlener, line 199: A possible "optimization" in this case would be to look up the index for each of these things once and then assign through that (similar to what was done previously), i.e.

```cpp
const auto sideIdx = encoder.index("side"); // outside the loop
encoder.set(encoderValue, sideIdx, k4geo::DetSide::bwd);
```

I don't know if the performance that we gain from only doing the internal map lookup once instead of repeating it for every cell id setting.

tmadlener (commented):
Good stuff :) The biggest question about this is probably validating all the affected geometries. Can that be automated for a one-off thing just to be sure things are truly unchanged?

There is a possible performance improvement for the encoder bit.

tmadlener:
Tentatively holding this for due to: https://github.com/key4hep/k4geo/issues/635#issuecomment-5278165525

tmadlener:
superseded by #636

PR #637, opened 2026-08-14
Backport some multithreading guards from upstream to the EDM4hep DRC output action

tmadlener:
The DD4hep EDM4hep output action has gained multithreading support in https://github.com/AIDASoft/DD4hep/pull/1417 which was never ported to here. With the merge of https://github.com/AIDASoft/DD4hep/pull/1240 that starts to lead to runtime crashes when using this output action.

BEGINRELEASENOTES
- Backport multithreading support to the DRC EDM4hep output action from the upstream EDM4hep output action that have been missing so far.

ENDRELEASENOTES

The DD4hep EDM4hep output action has gained multithreading support in
https://github.com/AIDASoft/DD4hep/pull/1417 which was never ported to
here. With the merge of https://github.com/AIDASoft/DD4hep/pull/1240
that starts to lead to runtime crashes when using this output action.

This should fix the CI failures that we currently see for the Key4hep nightlies.

I suppose this could also be used as a trigger for a discussion of whether it's possible to upstream this version of the action so that we don't run into similar issues / divergences in the future.

### DIFF
--- a/plugins/Geant4Output2EDM4hep_DRC.cpp
+++ b/plugins/Geant4Output2EDM4hep_DRC.cpp
@@ -27,6 +27,8 @@ using ROOTWriter = podio::ROOTFrameWriter;
 }
 #endif
 
+#include <atomic>
+
 #include "FiberDRCaloSDAction.h"
 
 /// Namespace for the AIDA detector description toolkit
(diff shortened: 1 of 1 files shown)


jbeirer:
Thanks @tmadlener, looks good to me. I will merge so that the tests will hopefully be fixed for tomorrows nightly. I would suggest to also add an MT variant of the existing tests at some point so the multi-threaded path is covered as well.

PR #640, opened 2026-08-18
Add MAIA_v1 with realistic solenoid field

madbaron:

BEGINRELEASENOTES
- Add initial realistic description of magnetic field for MAIA's central solenoid (MAIA_fieldMap_Solenoid5T_v1.root). A second xml using the field map is introduced alongside the analytic one with a test to keep them from drifting.

ENDRELEASENOTES


MuColl/MAIA/compact/MAIA_v1/fieldmap_description.md
```diff
@@ -0,0 +1,309 @@
...
+* **No coil-end detail.** The solenoid endcap side plates are not in the
+  geometry (`Solenoid_o1_v01_01.xml` says so explicitly) and so are not modelled.
+* **Nothing beyond the solenoid.** No anti-DID, no compensating or shielding
+  solenoids, no final-focus quadrupoles.
+
+## 6. Regenerating
+
+```bash
+scripts/run_in_container.sh python3 utils/make_maia_fieldmap.py \
+    -o fieldmaps/MAIA_fieldMap_Solenoid5T_v1.root --plots /tmp/plots
```
tmadlener, line 276: I think it would be nice if this script would also be here. It doesn't have to be generic, but it would allow to reproduce this map in the future. (and can be made generic at some point).

MuColl/MAIA/compact/README.md
```diff
@@ -13,3 +14,19 @@ The following MAIA detector models are available in k4geo ( current production m
...
+- Same geometry as MAIA_v0, included directly from `../MAIA_v0/`, so the two
+  cannot diverge in geometry; only the field description differs
+- Compact file: [./MAIA_v1/MAIA_v1.xml](./MAIA_v1/MAIA_v1.xml)
+- Solenoid field: [Field_Solenoid_Map_5T.xml](./MAIA_v1/Field_Solenoid_Map_5T.xml) —
+  a computed (Br, Bz) map with a realistic fringe field and a non-zero radial
+  component, so track states in the calorimeter endcaps no longer sit on a
+  5 T to 0 T cliff
+- The map is `fieldmaps/MAIA_fieldMap_Solenoid5T_v1.root`, regenerated with
+  [utils/make_maia_fieldmap.py](../../../utils/make_maia_fieldmap.py). Its
+  methodology, ROOT file format and validation are documented in
```
tmadlener, line 31: as of now this script is not here. So either docs or that fact have to be adapted.

MuColl/MAIA/compact/MAIA_v1/fieldmap_description.md
```diff
@@ -0,0 +1,309 @@
...
+
+### 3.2 Wiring
+
+`MAIA_v1.xml` is MAIA_v0 with the field map in place of the idealised solenoid.
+It includes the whole geometry from `../MAIA_v0/` rather than copying it, so the
+two versions cannot diverge in geometry; only the `<fields>` include differs.
+
+The map is located through `${k4geo_DIR}/fieldmaps/...`, matching the ILD field
+map XMLs. `k4geo_DIR` must point at this checkout rather than an installed
+k4geo; `test/CMakeLists.txt` sets it for ctest, and
```
tmadlener: I think this is not true it should also work with an installed version of the maps. Otherwise we should make sure it does, because that is likely what people will be using mostly, when not working on the geometry directly.

In general this subsection can go entirely, IMO. It's contained in the README describing the detector models (which links back here in any case) and it contains env setup things that are more or less independent of how the field map has been obtained.

tmadlener (commented):
I can't judge whether all of the magnetic field calculations are correct, but I am going to believe the validation that is contained therein has been done correctly.

One could consider moving the constants (and maybe the plugins) into another shared xml file(s) that is only included in both `MAIA_v0` and `MAIA_v1`, to cut down on the code duplication even further, but I have no strong opinion there and we can also leave that for a future PR.

MuColl/MAIA/compact/MAIA_v1/fieldmap_description.md
madbaron: Gone!

MuColl/MAIA/compact/README.md
madbaron: Added the script

MuColl/MAIA/compact/MAIA_v1/fieldmap_description.md
madbaron: Added in its non-generic form

tmadlener (approved):
OK. That script is definitely non-trivial 😅. There are a few magic (hardcoded) numbers here and there, but at least for me it's good enough for now and if every needed it should be adaptable.

PR #665, opened 2026-09-17
Set shared flag to false if single thread

andread3vita:
BEGINRELEASENOTES

- Set the flag controlling how workers share events to `false` when running in single-threaded mode. This change applies to the steering file in the examples.

ENDRELEASENOTES

Dear All,
I open this PR to address [this](https://github.com/AIDASoft/DD4hep/issues/1677) issue. This PR is a follow-up of [this other PR](https://github.com/AIDASoft/DD4hep/pull/1678).

Cheers,
Andrea


example/SteeringFile_IDEA_o1_v03.py
```diff
@@ -451,8 +451,9 @@
 def Geant4Output2EDM4hep_DRC_plugin(dd4hepSimulation):
     from DDG4 import EventAction, Kernel
 
+    shared = dd4hepSimulation.numberOfThreads > 1
```
andresailer, line 454: When I run multithreaded (-j 2) with the latest dd4hep version, I don't get any output for this plugin.
Can you confirm this observation?

Maybe you need to disable multithreading with this output for the moment

andread3vita: I’ve never used the multithreaded version of DD4hep. Do you have a minimal reproducer or a list of commands I could use to learn how it works? Thank you!

jmcarcell: `ddsim -j N`

andread3vita: When I run `ddsim` with `-j 2`, no output is produced, regardless of whether the steering-file modification is applied

andresailer: Thanks, so it isn't just my setup. I am digging to see why there is no output being produced.

BrieucF:
Hi Andrea, thanks! Can you propagate this to FCC-config as well?

andread3vita:
> Hi Andrea, thanks! Can you propagate this to FCC-config as well?

Already done :)

BrieucF:
> > Hi Andrea, thanks! Can you propagate this to FCC-config as well?
> 
> Already done :)

Indeed, sorry I missed it. Thanks!

example/SteeringFile_IDEA_o2_v01.py
```diff
@@ -467,6 +467,23 @@
 ##
 
 
+# Only use shared=True in MT mode to avoid double-save in ST mode
+def Geant4Output2ROOT_plugin(dd4hepSimulation):
```
jbeirer, line 471: Do we need a custom output plugin for o2? DDSim already sets shared based on the number of threads for its normal output, so this seems unnecessary and also replaces the normal EDM4hep output path with Geant4Output2ROOT

andread3vita:
Now this PR depends on https://github.com/AIDASoft/DD4hep/pull/1678 so it can be approved only once the other one is approved and merged. This is due to the last commit, where I have changed the kernel used in the plugin

example/SteeringFile_IDEA_o2_v01.py
andread3vita: yes, you are right. In this version, the PR affects only IDEA v3 o1

